PR #29, opened 2024-01-16
Replace Tachiyomi with Mihon in i18n/zh-rCN resources

nshcr:
Some changes are missing in Simplified Chinese localization.


AntsyLich:
Merged as part of 653940613d661eb371aab3b3c3a8181e4e308c43

nshcr:
Okay, thanks!
By the way, will the translation project on Weblate be revived as well?

PR #30, opened 2024-01-16
Fix local source cover updating

FooIbar:
Ref: tachiyomiorg/tachiyomi#10186

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreenModel.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreenModel.kt
@@ -18,6 +18,7 @@ import eu.kanade.domain.manga.interactor.GetExcludedScanlators
 import eu.kanade.domain.manga.interactor.SetExcludedScanlators
 import eu.kanade.domain.manga.interactor.UpdateManga
 import eu.kanade.domain.manga.model.chaptersFiltered
+import eu.kanade.domain.manga.model.copyFrom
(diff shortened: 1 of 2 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreenModel.kt
```diff
@@ -537,7 +538,13 @@ class MangaScreenModel(
...
         try {
             withIOContext {
-                val chapters = state.source.getChapterList(state.manga.toSManga())
+                val manga = state.manga.toSManga()
+                val chapters = state.source.getChapterList(manga)
+
+                // `LocalSource.getChapterList` has side effects since 82bdf63
+                if (state.manga != state.manga.copyFrom(manga)) {
+                    updateManga.awaitUpdateFromSource(state.manga, manga, manualFetch)
+                }
```
AntsyLich, line 547: Since this is specific to LocalSource move the code inside it?

GameMaster007: Yes

FooIbar: It's not feasible as LocalSource doesn't (and shouldn't) have access to update manga states, have to do it in MangaScreenModel.

FooIbar:
This patch attempts to fix https://git.mihon.dev/tachiyomi/tachiyomi/issues/8034#issuecomment-28275

PR #46, opened 2024-01-16
Add Support for 7zip archives

abdallahmehiz:
basically users can now use 7z and cb7 files to read their manga.
intended to close issue [#10121](https://github.com/tachiyomiorg/tachiyomi/issues/10121) from original tachi repo
Thanks to @claudemirovsky for the help getting this to work

Notes: 
 - The load times for archives with compression methods other than LZMA2 or Deflate are pretty bad
 - PPMd, the compression method, isn't supported by Apache's common compress for 7zip so PPMd archives can't be used

so in the end, supported compression methods are: Deflate, BZip2, LZMA and LZMA2.


AntsyLich:
> PPMd, the compression method, isn't supported by Apache's common compress for 7zip so PPMd archives can't be used
Would it be possible to throw a custom error similar to what we do for Rar 5?

AntsyLich:
And maybe a toast when opening the slow ones

i18n/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -776,6 +776,7 @@
     <string name="page_list_empty_error">No pages found</string>
     <string name="loader_not_implemented_error">Source not found</string>
     <string name="loader_rar5_error">RARv5 format is not supported</string>
+    <string name="loader_ppmd_error">PPMd compression is not supported</string>
```
AntsyLich, line 779:
```suggestion
    <string name="loader_7z_ppmd_error">7z's PPMd compression is not supported</string>
```

gradle/libs.versions.toml
```diff
@@ -31,6 +31,8 @@ jsoup = "org.jsoup:jsoup:1.17.2"
 disklrucache = "com.jakewharton:disklrucache:2.0.2"
 unifile = "com.github.tachiyomiorg:unifile:7c257e1c64"
 junrar = "com.github.junrar:junrar:7.5.5"
+common-compress = "org.apache.commons:commons-compress:1.25.0"
+xz = "org.tukaani:xz:1.9"
```
AntsyLich, line 35: Where is this being used 🤔

core/src/main/java/eu/kanade/tachiyomi/util/storage/SevenZUtil.kt
```diff
@@ -0,0 +1,23 @@
...
+import java.io.IOException
+import java.io.InputStream
+
+object SevenZUtil {
+    fun SevenZFile.getImages(): Sequence<ByteArray> {
+        return generateSequence {
+            try {
+                getNextEntry()
+            } catch (e: IOException) {
+                throw UnsupportedCompressionAlgorithmException()
```
AntsyLich, line 16: Can you check if [this](https://commons.apache.org/proper/commons-compress/apidocs/org/apache/commons/compress/archivers/sevenz/SevenZFile.html#matches(byte%5B%5D,int)) helps?

AntsyLich (changes requested):

core/src/main/java/eu/kanade/tachiyomi/util/storage/SevenZUtil.kt
abdallahmehiz: didn't notice that, thanks.

gradle/libs.versions.toml
abdallahmehiz: It's in the bundles section of the same file

AntsyLich: Where in the codebase

core/src/main/java/eu/kanade/tachiyomi/util/storage/SevenZUtil.kt
abdallahmehiz: i don't think it does what i need :thinking:

app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/ChapterLoader.kt
```diff
@@ -98,6 +104,15 @@ class ChapterLoader(
                 when (format) {
                     is Format.Directory -> DirectoryPageLoader(format.file)
                     is Format.Zip -> ZipPageLoader(tempFileManager.createTempFile(format.file))
+                    is Format.SevenZip -> try {
+                        SevenZipPageLoader(tempFileManager.createTempFile(format.file)) {
```
FooIbar, line 108: Creating temp files is not a good idea. https://github.com/mihonapp/mihon/issues/134
I would suggest `SevenZipPageLoader` takes a `UniFile` and use the `SeekableByteChannel` overload of `SevenZFile`.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/SevenZipPageLoader.kt
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Loader used to load a chapter from a .7z or .cb7 file.
+ */
+internal class SevenZipPageLoader(
+    private val file: UniFile,
+    private val notifySlowArchive: (method: String) -> Unit,
+) : PageLoader() {
+
+    private val zip by lazy { SevenZFile(SeekableInMemoryByteChannel(file.openInputStream().readBytes())) }
```
FooIbar, line 18: Read the file completely isn't ideal.
Instead, you can construct a `FileInputStream` and use `FileInputStream.getChannel`.
You can take https://github.com/FooIbar/EhViewer/blob/aec34c81aa586d2edc3d2a5cf494d238ddae3001/app/src/main/kotlin/com/hippo/unifile/UniFileExtensions.kt#L14 for reference.

abdallahmehiz: Like so? I couldn't get it to work since the URI isn't of "file" scheme.
```kotlin
private val zip by lazy { SevenZFile(File(URI(file.uri.toString())).inputStream().channel) }
```

FooIbar: Use https://developer.android.com/reference/android/content/ContentResolver#openFileDescriptor(android.net.Uri,%20java.lang.String) then wrap it into a `FileInputStream`.

abdallahmehiz: Doesn't the ContentResolver require a context?

FooIbar: Yes, can just use application context.

abdallahmehiz: I can't access the application context inside the SevenZipPageLoader, should i just pass the one from the ChapterLoader?

abdallahmehiz: doing this causes an IO exception of "Bad file descriptor".
image

FooIbar: Didn't I say wrap it into AutoCloseInputStream (which extends FileInputStream)?

abdallahmehiz: sorry, didn't see that. thank you for your patience

source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt
```diff
@@ -340,6 +342,13 @@ actual class LocalSource(
                         entry?.let { coverManager.update(manga, zip.getInputStream(it)) }
                     }
                 }
+                is Format.SevenZip -> {
+                    SevenZFile(tempFileManager.createTempFile(format.file)).use { archive ->
```
FooIbar, line 346: Ditto.

source-local/src/commonMain/kotlin/tachiyomi/source/local/io/Format.kt
```diff
@@ -17,6 +18,7 @@ sealed interface Format {
             when {
                 isDirectory -> Directory(this)
                 extension.equals("zip", true) || extension.equals("cbz", true) -> Zip(this)
+                extension.equals("7z", true) || extension.equals("cb7", true) -> SevenZip(this)
```
stevenyomi, line 21: At this rate, should just do `when (extension.lowercase())`.

```kotlin
fun valueOf(file: UniFile): Format {
    if (file.isDirectory) return Directory(file)
    return when (file.extension.lowercase()) {
        "zip", "cbz" -> Zip(file)
        "7z", "cb7" -> SevenZip(file)
        "rar", "cbr" -> Rar(file)
        "epub" -> Epub(file)
        else -> throw UnknownFormatException()
    }
}
```

gradle/libs.versions.toml
```diff
@@ -110,4 +111,5 @@ shizuku = ["shizuku-api", "shizuku-provider"]
 sqldelight = ["sqldelight-android-driver", "sqldelight-coroutines", "sqldelight-android-paging"]
 voyager = ["voyager-navigator", "voyager-screenmodel", "voyager-tab-navigator", "voyager-transitions"]
 richtext = ["richtext-commonmark", "richtext-m3"]
-test = ["junit", "kotest-assertions", "mockk"]
\ No newline at end of file
+test = ["junit", "kotest-assertions", "mockk"]
+sevenzip = ["common-compress", "xz"]
```
FooIbar, line 115: Can move xz to archive bundle since #326 is merged.

core/common/src/main/kotlin/eu/kanade/tachiyomi/util/storage/SevenZUtil.kt
```diff
@@ -0,0 +1,35 @@
...
+                getNextEntry()
+            } catch (e: IOException) {
+                throw UnsupportedCompressionAlgorithmException()
+            }
+        }.filter { !it.isDirectory && ImageUtil.isImage(it.name) { getInputStream(it) } }
+            .onEachIndexed { i, it ->
+                if (i > 0) return@onEachIndexed
+                val method = it.contentMethods.first().method
+                if (method !in GoodMethods) notifySlowArchives(method.name)
+            }
```
AntsyLich, line 30: Won't this just keep constantly notifying the user?

i18n/src/commonMain/resources/MR/base/strings.xml
```diff
@@ -776,6 +776,8 @@
     <string name="page_list_empty_error">No pages found</string>
     <string name="loader_not_implemented_error">Source not found</string>
     <string name="loader_rar5_error">RARv5 format is not supported</string>
+    <string name="loader_7zip_ppmd_error">PPMd compression is not supported</string>
+    <string name="loader_7zip_slow_archives">%s Archive, expect slow loading times.</string>
```
AntsyLich, line 780:
```suggestion
    <string name="loader_7zip_unsupported">Unsupported 7zip compression method</string>
    <string name="loader_7zip_slow_method">Archive with slow compression method, expect high loading time</string>
```

core/common/src/main/kotlin/eu/kanade/tachiyomi/util/storage/SevenZUtil.kt
```diff
@@ -0,0 +1,35 @@
...
+        SevenZMethod.DEFLATE,
+        SevenZMethod.COPY,
+    )
+
+    fun SevenZFile.getImages(notifySlowArchives: (method: String) -> Unit): Sequence<ByteArray> {
+        return generateSequence {
+            try {
+                getNextEntry()
+            } catch (e: IOException) {
+                throw UnsupportedCompressionAlgorithmException()
```
AntsyLich, line 23: Exception seems to come from `dump` archive. Make a new exception for unsupported 7zip archive one I suppose?

AntsyLich (changes requested):

core/common/src/main/kotlin/eu/kanade/tachiyomi/util/storage/SevenZUtil.kt
abdallahmehiz: Nope, it only notifies him once.

Efreak:
> throw a custom error similar to what we do for Rar 5

Suggest providing an option to open the new file with another app, for example perfect viewer or a file manager.

Oddly, there is a feature request closed by the author for [ppmd support in commons compress](https://issues.apache.org/jira/browse/COMPRESS-288) from 2014

abdallahmehiz:
> Suggest providing an option to open the new file with another app, for example perfect viewer or a file manager.

that's out of the scope of this pr

abdallahmehiz:
did a little oopsie

gradle/libs.versions.toml
AntsyLich: this didn't age well 😂

hollowshiroyuki: mihon spyware confirmed ✅✅✅

AntsyLich:
Supersedes by #949

PR #105, opened 2024-01-17
Updated all user-agent

kitsumed:
Updated forgotten user-agent who still had "Tachiyomi".


AntsyLich:
Can you use the same format as MangaUpdates?

kitsumed:
The user-agent now use the same format as manga-updates. With the only exception of Kavita where I kept previous "Kavita v${BuildConfig.VERSION_NAME}" at the end of the string.

`Mihon (${BuildConfig.APPLICATION_ID}) Kavita v${BuildConfig.VERSION_NAME}`

Issue #108, opened 2024-01-17
Enabling “Flash on page change” gives a black screen on page change.


Luiz-C-Lima:
### Steps to reproduce

1. Open any Manga
2. Enable "Flash on page change"
3. Change page

https://github.com/mihonapp/mihon/assets/54924297/3856447f-07da-46ec-87a5-35c8f97c244c

### Expected behavior

The page should be changed without issue.

### Actual behavior

You get a black screen that only goes away when you press the screen.

### Mihon version

0.16.1

### Android version

Android 10

### Device

Xiaomi Mi A2

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an official extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/extensions/issues/new/choose).
- [X] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [X] I have updated the app to version **[0.16.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

MajorTanya:
Isn't this meant for e-ink displays? 🤔 And the black screen doesn't go away by itself you say?

Luiz-C-Lima:
> Isn't this meant for e-ink displays? 🤔 And the black screen doesn't go away by itself you say?

No, the black screen does not go away by itself, just did a test right now, I waited for over 30 seconds and the screen was kept black.
Yes, it really appears to be for e-ink displays, but the image should still come back by itself, I think.

MajorTanya:
It definitely should come back, yeah
good catch, maybe someone will pick this up and fix it

OmarShezan:
I am working on this

PR #110, opened 2024-01-18
bug fix for #108

OmarShezan:
Bug fix: Enabling “Flash on page change” gives a black screen on page change. #108. fixed by not drawing anything on canvas after display refresh which essentially clears the screen


AntsyLich:
Fixed the issue in a slightly different way in 38d6ab80ce868707829dbc81de4170afe3c2f2a5

PR #112, opened 2024-01-18
Update Filipino Plurals

infyProductions:
Updated some Filipino Plurals to make sense grammatically.

Note:
- These are grammatical changes since some of them doesn't make any sense with the current translation.


MajorTanya:
You could have done this in #111 as well by the way

infyProductions:
I'm actually a beginner when doing pull requests like this. I could checkout both of the language files myself (include from #111) but I'm using the inbuilt editor in Github web. I should have learned to use git checkout instead to get both the standard strings and the plural strings instead of using a web browser to edit both files and submit different PRs.

PR #157, opened 2024-01-20
feat: add per-extension incognito mode

sdaqo:
This PR adds a per-extension incognito mode toggle. With this you can more or less "automate" using the incognito mode as oftentimes there are only specific extensions (heh) where you need it. 

### Image
Screenshot_20240120_132036


AntsyLich:
You need to consider the fact that source can change id and extensions can also have new source

sdaqo:
Do you think something like this will do?
```diff
     private fun registerUpdatedExtension(extension: Extension.Installed) {
         val mutInstalledExtensions = _installedExtensionsFlow.value.toMutableList()
         val oldExtension = mutInstalledExtensions.find { it.pkgName == extension.pkgName }
         if (oldExtension != null) {
+            // If a extension has incognito mode enabled we need to consider that sources change and update them
+            if (oldExtension.sources.first().id.toString() in sourcePreferences.incognitoSources().get()) {
+                oldExtension.sources
+                    .map { it.id }
+                    .let { toggleSource.awaitIncognito(it, false) }
+
+                extension.sources
+                    .map { it.id }
+                    .let { toggleSource.awaitIncognito(it, true) }
+            }
+
             mutInstalledExtensions -= oldExtension
         }
         mutInstalledExtensions += extension
         _installedExtensionsFlow.value = mutInstalledExtensions
     }

```

sdaqo:
I guess removing the sources from the list when the extension is uninstalled is also a good idea, else we would have a bunch of orphan source ids that do not belong to a extension.

sdaqo:
Bumping, because I wanted to know if there is anything you want me to do else, or if this is only stalled because you want to implement other stuff before this.

federicogith:
Since I find this feature really useful and timesaving, @MajorTanya is there any reason why this is stalled?

MajorTanya:
> Since I find this feature really useful and timesaving, @MajorTanya is there any reason why this is stalled?

First: please don't ping random people into PRs, especially when they have not even been there before.

Second: I have no control over what gets accepted or not, and there may be any number of reasons for stuff to be stalled. You coming in to necropost is neither helping nor welcome. No, this is not an invitation to ping AntsyLich instead or anything else.

Either it will be accepted at some point or it won't. Until such time, you'll have to wait.

AntsyLich (changes requested):
Additionally instead of saving each source ids we could potentially resolve the IDs whenever we initialize an extension? Something like a new class that listens for changes in available extensions filters them by incognito disabled extension and maps to the IDs of sources inside them.

sdaqo:
> Additionally instead of saving each source ids we could potentially resolve the IDs whenever we initialize an extension? Something like a new class that listens for changes in available extensions filters them by incognito disabled extension and maps to the IDs of sources inside them.

Honestly I do not quite know what exactly you mean by this... 
I think the current solution is fine, your suggestions seems rather complex, but maybe I am just not comfortable enough with the codebase yet.

Syer10 (approved):

AntsyLich:
Thanks for the initial work and sorry for taking almost a year to merge

PR #194, opened 2024-01-22
Adding Type-safe project accessors

theolm:
# Description
Since the app has several modules, we can take advantage of the [gradle new feature](https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:type-safe-project-accessors). The project accessors are mapped from the project path. For example, if a project path is `:commons:utils:some:lib` then the project accessor will be `projects.commons.utils.some.lib`.

before:
```kt
dependencies {
    implementation(project(":i18n"))
    implementation(project(":core"))
}
```

after:
```kt
dependencies {
    implementation(projects.i18n)
    implementation(projects.core)
}
```

This way we don't need to relay on strings that are prone to error.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -47,3 +47,5 @@ include(":presentation-widget")
 include(":presentation-core")
 include(":source-local")
 include(":core-metadata")
+
+enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -49,5 +49,3 @@ include(":presentation-widget")
 include(":presentation-core")
 include(":source-local")
 include(":core-metadata")
-
-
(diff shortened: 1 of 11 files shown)


settings.gradle.kts
```diff
@@ -47,3 +47,5 @@ include(":presentation-widget")
 include(":presentation-core")
 include(":source-local")
 include(":core-metadata")
+
+enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
```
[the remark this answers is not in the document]
theolm: Done

settings.gradle.kts
```diff
@@ -47,3 +49,5 @@ include(":presentation-widget")
 include(":presentation-core")
 include(":source-local")
 include(":core-metadata")
+
+
```
AntsyLich, line 53: Empty space

theolm: Fixed

PR #218, opened 2024-01-24
update backup file extension

manuel-g-ferreira:
update backup extension file to `mihonbk`


jobobby04:
Why though? Its not like we are changing the format

manuel-g-ferreira:
was just a suggestion. In my point of view was to keep it consistent with the app name.

jobobby04:
I don't think this is a good idea, because then all the forks have to support this new file type and that means a whole 3 file types that we need to support for people moving around.

manuel-g-ferreira:
makes sense! great argument.

Issue #207, opened 2024-01-23
Optionally, increase total and unread count font size on individual library issues for us old guys with cataracts. :-)


cbussa:
### Describe your suggested feature

In the standard library view, there's a count of unread issues and optionally a number of available issues.  They work fine. However, for me even when on a large tablet, they're still hard to read, being too small. I wish there were a "small / normal / larger / largest" slider within the app for those items. 

Believe it or not, the numbers on the attach screenshot below are **barely** readable for me, while the title itself is easier simply because I think they're displayed in bold white, while the contrast is not as great for the previous numbers.

![Screenshot_20240122-223726~2](https://github.com/mihonapp/mihon/assets/3461529/d6992757-411e-410b-bfd2-d62d91acd3a5)

### Other details

The setting is: Settings /  Library  /  Show unread count on Updates icon. The actual count and total items are all correct, but I have a difficult time reading them he went on a 12-in tablet.

THIS IS ACTUALLY FIXABLE LOCALLY by going to android / setting / display / font size and moving the scroller, but now every _other_ app on my tablet has that larger font size as well.  Here's a picture with it set two sizes larger.

![Screenshot_20240122-225607~2](https://github.com/mihonapp/mihon/assets/3461529/b860c26a-3e1f-493d-89cf-75d6a928b033)

It's livable -- but if someone gets bored sometime, I'd really appreciate a application font size boost someday. Thanks!

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an official extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/extensions/issues/new/choose).
- [X] I have updated the app to version **[0.16.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

AntsyLich:
> I wish there were a "small / normal / larger / largest" slider within the app for those items.

Yeah not going to add that. From the look of it seems like you are on some tablet. We could potentially increase the size of those texts on tablet layout

AntsyLich:
The sizes were incresed iirc. Is this still an issue?

PR #273, opened 2024-01-29
Add a font size parameter (#207)

mathis-chapuis:
This PR adds a preferred font size parameter in Settings > Appearance


AntsyLich:
If people need to change the font size they can do so from system settings. For text size issue on specific places we can tweak it there.

PR #274, opened 2024-01-29
Gradle refactor - Kotlin Pre-compiled scripts

theolm:
## Description

This PR isolates all the shared logic of build.gradle files into 3 different plugins:
- `android-application-setup.gradle` used for the app (this one is not really shared but it helps keeping the gradle clean)
- `android-module-setup.gradle` used for all the android modules
- `kmp-module-setup.gradle` used for KMP modules

Besides this changes, I also organized the `buildSrc` in subfolders:
- `config` contain object files with values shared between the gradle files
- `extensions` contains extensions related to the build process
- `plugins` contains all the plugins

This PR is meant to be the first step of a gradle cleanup. I appreciate feedback and help here... I'm sure there is a lot more stuff that we can remove and optimizations to be made.

@AntsyLich Feel free to push changes directly to this PR. Let's just try to keep simple, there is nothing worse than a project with dozens of plugins that no one knows which one to use. 

ps. I tried to move the build logic from `buildSrc` to `build-logic` (I do that in every project that I work on) but for some reason in this project, I could not make it work.... anyway, in gradle 8.5 we should not lose anything by using the buildSrc.


AntsyLich:
Superseded by e448e40406e8d9916120a278e42829a6f1b25a7a

PR #325, opened 2024-02-02
Fix extracting ComicInfo.xml files in local source

FooIbar:
Related: #30


AntsyLich:
Looks fine but can i get more context on the issue it's fixing?

FooIbar:
Follow-up to 4fcdde4, can't use File API to access URI provided by SAF.

PR #415, opened 2024-02-18
Fix some issues from 7ff95e2

sirlag:
Tweaked header code to fix bug introduced in earlier timestamp fix

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryScreenModel.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryScreenModel.kt
@@ -62,7 +62,7 @@ class HistoryScreenModel(
                 val beforeDate = before?.item?.readAt?.time?.toLocalDate()
                 val afterDate = after?.item?.readAt?.time?.toLocalDate()
                 when {
-                    (beforeDate != afterDate) and (afterDate != null) -> afterDate?.let(HistoryUiModel::Header)
+                    beforeDate != afterDate && afterDate != null -> HistoryUiModel.Header(afterDate)
                     // Return null to avoid adding a separator between two items.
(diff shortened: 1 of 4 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryScreenModel.kt
```diff
@@ -63,9 +63,9 @@ class HistoryScreenModel(
                 val beforeDate = before?.item?.readAt?.time?.toLocalDate() ?: LocalDate.MIN
                 val afterDate = after?.item?.readAt?.time?.toLocalDate() ?: LocalDate.MIN
                 when {
-                    beforeDate.isAfter(afterDate)
-                        or afterDate.equals(LocalDate.MIN)
-                        or beforeDate.equals(LocalDate.MIN) -> HistoryUiModel.Header(afterDate)
+                    beforeDate.equals(LocalDate.MIN)
+                        or beforeDate.isAfter(afterDate)
+                        and !afterDate.equals(LocalDate.MIN) -> HistoryUiModel.Header(afterDate)
```
AntsyLich, line 68: Maybe something like this? Alongside the other comment. Same for Updates screen
```suggestion
                    beforeDate != afterDate and  afterDate != null -> HistoryUiModel.Header(afterDate)
```

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryScreenModel.kt
sirlag: Pushed the fix, it looks a lot cleaner. I think I got too hung up on wanting to use the date comparison functions.

PR #437, opened 2024-02-21
Incremental loading and filter system for local source

### LINKED ISSUE #365: Make local sources load in chunks like other extensions/scources

Large local sources with 300+ entries takes forever to load, if it loaded similar to other extensions/sources that would be great.

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #397: Cannot search using tags on local source
### Steps to reproduce

- Go to local source,
- Search for a tag or click in a book, click on the tag and then click search
- No results are shown

### Expected behavior

It should shown content that has that given tag

### Actual behavior

There is not a single result shown

### Crash logs

There is no crash log

### Mihon version

0.16.3

### Android version

android 13

### Device

Galaxy S10e

### Other details

No more details

### Acknowledgements

Shamicen:
Closes  #365
Fixes #397

I implemented my old Tachiyomi PR in Mihon.

**New features:**
- Local Source manga now load in chunks of 10 instead of having to wait for all Manga to load.
- Manga now get loaded with Metadata applied.
- Manga can now be filtered.

**Caveats:**
- Filters can only be applied to manga that have already loaded. 
- Order By is only available once all manga have loaded.
-  Adding new Manga to Local Source requires an app restart.
- To update the filters the user has to press the reset button.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreen.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreen.kt
@@ -37,6 +37,7 @@ import eu.kanade.presentation.manga.components.SetIntervalDialog
 import eu.kanade.presentation.util.AssistContentScreen
 import eu.kanade.presentation.util.Screen
 import eu.kanade.presentation.util.isTabletUi
+import eu.kanade.tachiyomi.source.CatalogueSource
 import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.source.isLocalOrStub
 import eu.kanade.tachiyomi.source.online.HttpSource
@@ -346,7 +347,7 @@ class MangaScreen(
         }
 
         val previousController = navigator.items[navigator.size - 2]
-        if (previousController is BrowseSourceScreen && source is HttpSource) {
+        if (previousController is BrowseSourceScreen && source is CatalogueSource) {
             navigator.pop()
             previousController.searchGenre(genreName)
         } else {
--- a/i18n/src/commonMain/resources/MR/base/strings.xml
+++ b/i18n/src/commonMain/resources/MR/base/strings.xml
@@ -644,6 +644,9 @@
     <string name="chapter_not_found">Chapter not found</string>
     <string name="local_invalid_format">Invalid chapter format</string>
     <string name="local_filter_order_by">Order by</string>
+    <string name="local_filter_random_page">Load random manga page (ignores all filters)</string>
+    <string name="local_filter_text_search_header">Values separated by ,\nuse - before a value to filter it out</string>
+    <string name="local_filter_info_header">⚠️ Please use "Reset" to update your filters\n\nⓘ Filtering only works for manga that have already loaded\n\nⓘ "Order by" will only become available once all your manga have loaded\n\n⚠️ Adding manga to Local Source requires an app restart</string>
     <string name="date">Date</string>
 
     <!-- Manga info -->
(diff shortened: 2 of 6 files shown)


AntsyLich:
Can we have the local source items indexed separately similar to download index and use filesystem observer to detect changes ([FileObserver](https://developer.android.com/reference/android/os/FileObserver) for android).

Shamicen:
I don't really understand what you mean by indexed separately. If this about adding new files to `LocalSource` while the app is running we could achieve that with a FileObserver and adding the new manga to the end of the list, but as far as i can see FileObserver needs to have file access and i don't think this is possible with Scoped Storage.

AntsyLich:
Right forgot about SAF. Lemme think about this for a bit.

Shamicen:
I now implemented logic to only re-scan `ComicInfo` files if the manga directory was modified. It would be good if you could take a look at the migration I wrote and check if it makes sense. I tested it a few time and it seems to work, but I would rather not be responsible for corrupting users databases.

Shamicen:
I added support for adding new manga while the app is running

AntsyLich:
Cache local source entry metadata in the database. Reindex them on `MainActivity.onResume()`. No need to make append new manga to the end users can just search again. You may want to tweak the caching process to make it as efficient as possible. Like if a folder was already cached before compare their modification time and stuff

Shamicen:
> Cache local source entry metadata in the database. 

I already do cache the metadata in the database

> Reindex them on `MainActivity.onResume()`.

Why do you want to index them at that point? currently loading the metadata for an entire page (10 Manga) from the db takes around 0.017s. I do not think we gain anything from indexing them at app start. This would also lead to chunk sizes becoming irregular if a user removed or added manga to their local source directory. 

> No need to make append new manga to the end users can just search again. 

Simply searching again won't help here. When the Local Source class is initialized, it loads all the directories (manga chunks) at that time. After all manga have been loaded once, I append any new directories that have been added to the Local Source directory since app launch to the manga chunks. If we remove this functionality, a complete app restart would be necessary to add new manga to Local Source.

> You may want to tweak the caching process to make it as efficient as possible. Like if a folder was already cached before compare their modification time and stuff

I already do this it only reads the ComicInfo file from disk if the manga has not been cached yet or if it was modified. 

How about testing the current version once to compare the initial uncached load with subsequent cached loads? If it still loads to slow I can try to optimize it further. However, if I recall correctly, most of the remaining page load time was spent outside the functions I implemented.

I also discovered a bug where opening Local Source throws an error if it is empty. I will investigate this issue over the coming weekend, and for now, I will convert this PR to a draft.

AntsyLich:
> > Cache local source entry metadata in the database.
> 
> I already do cache the metadata in the database
> 

Yeah my bad what I meant was having them in a separate local source only table instead of indexing in the manga database. We can also directly do filtering in database then. All my other comments is a follow up to this.

xMohnad:
Check https://github.com/mihonapp/mihon/issues/2071

Smol-Ame:
@xMohnad don't mention/comment unrelated feature request on an existing PR. It's unnecessary & rude to ask/make a contributor work on extra things if that's not the original intention of their PR/not something they want to do in the first place.

If someone wants to work/contribute this feature, they can do it in a separate PR.

xMohnad:
> @xMohnad don't mention/comment unrelated feature request on an existing PR. It's unnecessary & rude to ask/make a contributor work on extra things if that's not the original intention of their PR/not something they want to do in the first place.
> 
> If someone wants to work/contribute this feature, they can do it in a separate PR.

Ok, I'm sorry

PR #450, opened 2024-02-25
feat: db changes to accommodate new cross device syncing logic.

kaiserbh:
Using timestamp to sync is a bit skewed due to system clock etc and therefore there was a lot of issues with it such as removing a manga that shouldn't have been removed. Marking chapters as unread even though it was marked as a read. By using versioning system it should eliminate those issues.


data/src/main/sqldelight/tachiyomi/migrations/2.sqm
```diff
@@ -0,0 +1,40 @@
...
+END;
+
+-- manga_categories table
+DROP TRIGGER IF EXISTS insert_manga_category_update_version;
+CREATE TRIGGER insert_manga_category_update_version AFTER INSERT ON mangas_categories
+BEGIN
+    UPDATE mangas
+    SET version = version + 1
+    WHERE _id = new.manga_id AND (SELECT is_syncing FROM mangas WHERE _id = new.manga_id) = 0;
+END;
```
AntsyLich, line 40: Add newline

app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
```diff
@@ -98,4 +98,6 @@ private fun Manga.toBackupManga() =
         updateStrategy = this.updateStrategy,
         lastModifiedAt = this.lastModifiedAt,
         favoriteModifiedAt = this.favoriteModifiedAt,
+        version = this.version,
+        isSyncing = this.isSyncing,
```
AntsyLich, line 102: I've been out of the loop on the syncing logic for a while. What is `isSyncing` used for?

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
kaiserbh: It's to ensure consistency across devices during sync operations. This flag plays a crucial role in preventing automatic version increments by our triggers during the sync process. Without this flag, the act of syncing itself would inadvertently increment the manga's version number, leading to inconsistencies across devices.

Here's a practical scenario to illustrate why the is_syncing flag is essential: Imagine two devices, A and B, where Device A is on version 6 and Device B on version 7 of a manga. Ideally, syncing Device A should update its manga version to match Device B's version 7. However, without the is_syncing flag acting as a safeguard, the sync process would unintentionally increment the version to 8. This increment during syncing could lead to a situation where, if Device B holds the latest version and subsequent changes are made on Device A, those changes might not be recognized appropriately due to the version discrepancy.

AntsyLich: In that case it's not necessary to make it a part of the Manga/Chapter model is it? You can just make it a parameter of the insert function.

kaiserbh: I don't think that will work, since the issue is not really during insert but when updating manga/chapter details from the database, the db trigger will always run when there is an update to a specific fields in this case, for manga it would be when adding and removing (favorite) value changes, during insert it defaults to 0.

It's basically used during restoring chapter/manga:

```kotlin
private suspend fun updateExistingChapters(chapters: List<Chapter>) {
        handler.await(true) {
            chapters.forEach { chapter ->
                chaptersQueries.update(
                    mangaId = null,
                    url = null,
                    name = null,
                    scanlator = null,
                    read = chapter.read,
                    bookmark = chapter.bookmark,
                    lastPageRead = chapter.lastPageRead,
                    chapterNumber = null,
                    sourceOrder = null,
                    dateFetch = null,
                    dateUpload = null,
                    chapterId = chapter.id,
                    version = chapter.version,
                    isSyncing = chapter.isSyncing,
                )
            }
        }
    }
``` 


```kotlin
        return if (isSync) {
            chapter.copy(
                id = dbChapter.id,
                bookmark = chapter.bookmark || dbChapter.bookmark,
                read = chapter.read,
                lastPageRead = chapter.lastPageRead,
                isSyncing = 1,
            )
        } else {
            chapter.copyFrom(dbChapter).let {
                when {
                    dbChapter.read && !it.read -> it.copy(read = true, lastPageRead = dbChapter.lastPageRead)
                    it.lastPageRead == 0L && dbChapter.lastPageRead != 0L -> it.copy(
                        lastPageRead = dbChapter.lastPageRead,
                    )
                    else -> it
                }
            }
        }
```


I need to set the isSyncing "flag" to 1 that way it disables the trigger for that particular manga/chapter temporarily. If there is a better way to disable a trigger in sqlite do let me know.

AntsyLich: What i'm proposing is something like the code below instead of making it part of the Manga or Chapter class. It's fine to keep it in db. Will this not work?
```kt
private suspend fun updateExistingChapters(chapters: List<Chapter>, isSyncing: Boolean) {
    handler.await(true) {
        chapters.forEach { chapter ->
            chaptersQueries.update(
                mangaId = null,
                url = null,
                name = null,
                scanlator = null,
                read = chapter.read,
                bookmark = chapter.bookmark,
                lastPageRead = chapter.lastPageRead,
                chapterNumber = null,
                sourceOrder = null,
                dateFetch = null,
                dateUpload = null,
                chapterId = chapter.id,
                version = chapter.version,
                isSyncing = isSyncing.toLong(),
            )
        }
    }
}
    ```

Comment (kaiserbh) [reply app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt]: Oh yea, that should work, apologies misunderstood you.

Comment (kaiserbh) [reply app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt]: > Some class still has that field

Oh which one do you mean? I know it's in the Manga and Chapter.kt

> Some class still has that field

Okay I think it's done now, let me know if that's all?

File: domain/src/main/java/tachiyomi/domain/chapter/model/Chapter.kt [hunk 3]
```diff
@@ -14,6 +14,8 @@ data class Chapter(
     val chapterNumber: Double,
     val scanlator: String?,
     val lastModifiedAt: Long,
+    val version: Long,
+    val isSyncing: Long,
```
AntsyLich, line 18: Here

domain/src/main/java/tachiyomi/domain/manga/model/Manga.kt
```diff
@@ -29,6 +29,9 @@ data class Manga(
     val initialized: Boolean,
     val lastModifiedAt: Long,
     val favoriteModifiedAt: Long?,
+    val version: Long,
+    val isSyncing: Long,
+
```
AntsyLich, line 34: Here

AntsyLich (changes requested):
`isSyncing` are still here

domain/src/main/java/tachiyomi/domain/manga/model/Manga.kt
kaiserbh: I can't remove those two, I will get error when mapping the manga or chapter? in MangaRepositoryImpl and Chapter, since the model is not 1:1? 

I mean I could fix it by adding dummy or unused param of isSyncing with a long type but that's counter intuitive and the model should be 1:1 to what the db have. 

```prolog
Type mismatch.
Required:
(Long, Long, String, String?, String?, String?, List<String>?, String, Long, String?, Boolean, Long?, Long?, Boolean, Long, Long, Long, Long, UpdateStrategy, Long, Long, Long?, Long, Long) → TypeVariable(T)
Found:
KFunction23<Long, Long, String, String?, String?, String?, List<String>?, String, Long, String?, Boolean, Long?, Long?, Boolean, Long, Long, Long, Long, UpdateStrategy, Long, Long, Long?, Long, Manga>
```

AntsyLich: In that case might as well have a separate table for this stuff since it's needed but not necessarily as part of manga 🤔

kaiserbh: > In that case might as well have a separate table for this stuff since it's needed but not necessarily as part of manga/chapter 🤔

Hmm that will make it more complex, and it's part of it technically. Since we want to disable the trigger for that particular manga or chapter, like we still want it to work for the other ones while it's syncing if the user like reads other manga or chapter that's not being synced the trigger should still work. 

It's the same thing with the `LastFavoritedAt` field which we only use to check if a manga has ever been Favorited even once that way we don't sync junk stuff like users searches etc, keeps the data small etc. Besides that we are not using it anywhere else.

AntsyLich: `LastFavoritedAt` is used for sorting too. Also pretty sure you can do a join inside a trigger. If that doesn't work let's just go with a unused parameter route

kaiserbh: Okay done.

PR #459, opened 2024-02-28
Made some changes to ComicInfo metadata

Shamicen:
I made some changes to `ComicInfo` metadata generation:

The `web field` had a schema change and can now contain multiple space separated reference website links. Therefore ComicInfo metadata will from now on include the `source URL` and all available `tracker URLs`.
For more information look at this issue https://github.com/anansi-project/comicinfo/issues/56 and this PR https://github.com/anansi-project/comicinfo/pull/57

I changed the `translator field` to use the source name if no `scanlator` name is available.

### DIFF
--- a/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
+++ b/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
@@ -98,7 +98,7 @@ fun Manga.hasCustomCover(coverCache: CoverCache = Injekt.get()): Boolean {
 fun getComicInfo(
     manga: Manga,
     chapter: Chapter,
-    urls: String,
+    urls: List<String>,
     categories: List<String>?,
     sourceName: String,
 ) = ComicInfo(
--- a/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
+++ b/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
@@ -111,7 +111,7 @@ fun getComicInfo(
             ComicInfo.Number(it.toString())
         }
     },
-    web = ComicInfo.Web(urls),
+    web = ComicInfo.Web(urls.joinToString(" ")),
     summary = manga.description?.let { ComicInfo.Summary(it) },
     writer = manga.author?.let { ComicInfo.Writer(it) },
     penciller = manga.artist?.let { ComicInfo.Penciller(it) },
--- a/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
+++ b/app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
@@ -95,7 +95,13 @@ fun Manga.hasCustomCover(coverCache: CoverCache = Injekt.get()): Boolean {
 /**
  * Creates a ComicInfo instance based on the manga and chapter metadata.
  */
-fun getComicInfo(manga: Manga, chapter: Chapter, chapterUrl: String, categories: List<String>?) = ComicInfo(
+fun getComicInfo(
+    manga: Manga,
+    chapter: Chapter,
+    urls: String,
+    categories: List<String>?,
+    translator: String?,
+) = ComicInfo(
     title = ComicInfo.Title(chapter.name),
     series = ComicInfo.Series(manga.title),
     number = chapter.chapterNumber.takeIf { it >= 0 }?.let {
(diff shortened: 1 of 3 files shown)


AntsyLich:
> I changed the `translator field` to use the source name if no `scanlator` name is available.

Not sure i like that change

Shamicen:
I was planning to make it so that `LocalSource` chapters will appear with their metadata in the future. 

As of now we would be unable to display a scanlator for most sources, because a lot of them don't populate the scanlator field with their name if they are also the scanlators.

Their are also a few aggregator sources that already "incorrectly" populate the scanlator field with their source name.

For use inside `Mihon` the field will only show up in `LocalSource` and even if the source wasn't the original scanlator seeing which source the manga is from looks better / gives the user more information then just leaving it blank. 

We could make it standard policy for sources that are also scanlators to populate the scanlator field with their source name and for sources that aren't to not populate it, if we don't want to risk "wrong" translator names. Whether or not the 3rd party extension will actually follow it is a different question but it can't hurt to try.

We could alternatively create another non schema field and use this for our `LocalSource` when no scanlator is available.

Tell me what you think and I'll make the necessary changes.

AntsyLich:
Could do a custom field then for source and in local source do `scanlator ?: mihon_source`

Shamicen:
> Could do a custom field then for source

I implemented it

core-metadata/src/main/java/tachiyomi/core/metadata/comicinfo/ComicInfo.kt
```diff
@@ -154,6 +156,10 @@ data class ComicInfo(
     @Serializable
     @XmlSerialName("Categories", "http://www.w3.org/2001/XMLSchema", "ty")
     data class CategoriesTachiyomi(@XmlValue(true) val value: String = "")
+
+    @Serializable
+    @XmlSerialName("SourceMihon", "http://www.w3.org/2001/XMLSchema", "ty")
```
AntsyLich, line 161:
```suggestion
    @XmlSerialName("SourceMihon", "http://www.w3.org/2001/XMLSchema", "mh")
```

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
```diff
@@ -95,7 +95,13 @@ fun Manga.hasCustomCover(coverCache: CoverCache = Injekt.get()): Boolean {
 /**
  * Creates a ComicInfo instance based on the manga and chapter metadata.
  */
-fun getComicInfo(manga: Manga, chapter: Chapter, chapterUrl: String, categories: List<String>?) = ComicInfo(
+fun getComicInfo(
+    manga: Manga,
+    chapter: Chapter,
+    urls: String,
```
AntsyLich, line 101:
```suggestion
    urls: List<String>,
```

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
```diff
@@ -105,7 +111,7 @@ fun getComicInfo(manga: Manga, chapter: Chapter, chapterUrl: String, categories:
             ComicInfo.Number(it.toString())
         }
     },
-    web = ComicInfo.Web(chapterUrl),
+    web = ComicInfo.Web(urls),
```
AntsyLich, line 114:
```suggestion
    web = ComicInfo.Web(urls.joinToString(" ")),
```

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -626,9 +628,22 @@ class Downloader(
...
     ) {
-        val chapterUrl = source.getChapterUrl(chapter.toSChapter())
         val categories = getCategories.await(manga.id).map { it.name.trim() }.takeUnless { it.isEmpty() }
-        val comicInfo = getComicInfo(manga, chapter, chapterUrl, categories)
+        val urls = listOf(source.getChapterUrl(chapter.toSChapter()))
+            .plus(getTracks.await(manga.id).mapNotNull { it.remoteUrl.takeUnless { url -> url.isEmpty() } })
+            .map { it.trim() }
+            .distinct()
+            .joinToString(" ")
+            .trim()
```
AntsyLich, line 637:
```suggestion
        val urls = getTracks.await(manga.id)
            .mapNotNull { track ->
                track.remoteUrl.takeUnless { url -> url.isBlank() }?.trim()
             }
            .plus(source.getChapterUrl(chapter.toSChapter()).trim())
            .distinct()
```

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -626,9 +628,22 @@ class Downloader(
         chapter: Chapter,
         source: HttpSource,
     ) {
-        val chapterUrl = source.getChapterUrl(chapter.toSChapter())
         val categories = getCategories.await(manga.id).map { it.name.trim() }.takeUnless { it.isEmpty() }
-        val comicInfo = getComicInfo(manga, chapter, chapterUrl, categories)
+        val urls = getTracks.await(manga.id)
+            .mapNotNull { track ->
+                track.remoteUrl.takeUnless { url -> url.isBlank() }?.trim()
+             }
```
AntsyLich, line 635:
```suggestion
            }
```

PR #466, opened 2024-02-29
Update image-decoder

az4521:
fixes crashing when trying to load corrupt images below 12 bytes in size


az4521:
give it a minute though you gotta let jitpack cook (20 minutes to unzip two files, then 10 seconds of compiling)

az4521:
ok should be good now, jitpack is done building

PR #483, opened 2024-03-05
Fix: Split wide pages

### LINKED ISSUE #479: New `ZipPageLoader` breaks "Split wide pages"
### Steps to reproduce

Try opening a `CBZ` or `Zip` file, containing a wide image, with `Split wide pages` enabled (should also be broken for rar and epub but untested).

### Expected behavior

Should load.

### Actual behavior

Doesn't load.

### Crash logs

```
java.io.IOException: Resetting to invalid mark
at java.io.BufferedInputStream.reset(BufferedInputStream.java:450)
at tachiyomi.core.common.util.system.ImageUtil.extractImageOptions(ImageUtil.kt:559)
at tachiyomi.core.common.util.system.ImageUtil.extractImageOptions$default(ImageUtil.kt:550)
at tachiyomi.core.common.util.system.ImageUtil.isWideImage(ImageUtil.kt:128)
at eu.kanade.tachiyomi.ui.reader.viewer.pager.PagerPageHolder.process(PagerPageHolder.kt:198)
at eu.kanade.tachiyomi.ui.reader.viewer.pager.PagerPageHolder.access$process(PagerPageHolder.kt:33)
at eu.kanade.tachiyomi.ui.reader.viewer.pager.PagerPageHolder$setImage$2.invokeSuspend(PagerPageHolder.kt:144)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
```

### Mihon version

newest debug version

### Android version

Android 12

### Device

Pixel 5 emulator

### Other details

The cause of this bug is, that ZipFiles created from channels do not support `mark` and `reset` (their streams do return true when using `markSupported()` but they actually don't). 

Split wide images uses ImageUtils `extractImageOptions()`, which needs the stream to support reset to work.

Possible solutions: 

1. pass the stream function to `extractImageOptions()`. (I tested this and it works. Performance did seem to be terrible, but that could just be my emulator.)
2. read the stream into a ByteArray in `setImage()` and use a buffered `byteArrayInputStream` from there. (I tested this as well and performance seemed to be OK but I can't really say anything definitve, because my emulator performance is not the best.) 
3. Pass the ZipFile and its entry to `extractImageOptions()`. We would probably need to do that for `Rar` and `Epub` files as well so this is probably not a good solution. (this should in theory have the exact same effect as 1 but the performance was a lot better in the limited tests I ran, but like i said it could just have been my emulator being weird.)

I could write a fix myself, but I think it would be better if someone that can test the performance impact would do it instead / give feedback about the best performing solution.

### Acknowledgements

Shamicen:
Fixes  #479

`Apache Commons Compress` archives do not support mark and reset, this causes problems when functions like "Split wide pages" are used, that rely on these functions working. More info in  #479.

I fixed it by loading the images into a `ByteArray` instead of reading them directly from file.

Someone should test the performance in `Long strip` mode before merging this PR, because my emulator is too laggy too tell whether or not this approach degrades reader performance.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/EpubPageLoader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/loader/EpubPageLoader.kt
@@ -3,6 +3,11 @@ package eu.kanade.tachiyomi.ui.reader.loader
 import eu.kanade.tachiyomi.source.model.Page
 import eu.kanade.tachiyomi.ui.reader.model.ReaderPage
 import eu.kanade.tachiyomi.util.storage.EpubFile
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.runBlocking
 import java.nio.channels.SeekableByteChannel
 
 /**
@@ -17,9 +22,14 @@ internal class EpubPageLoader(channel: SeekableByteChannel) : PageLoader() {
     override suspend fun getPages(): List<ReaderPage> {
         return epub.getImagesFromPages()
             .mapIndexed { i, path ->
-                val streamFn = { epub.getInputStream(epub.getEntry(path)!!) }
+                val imageBytesDeferred: Deferred<ByteArray> = CoroutineScope(Dispatchers.IO).async {
+                    epub.getInputStream(epub.getEntry(path)!!).buffered().use { stream ->
+                        stream.readBytes()
+                    }
+                }
+                val imageBytes by lazy { runBlocking { imageBytesDeferred.await() } }
                 ReaderPage(i).apply {
-                    stream = streamFn
+                    stream = { imageBytes.copyOf().inputStream() }
                     status = Page.State.READY
                 }
             }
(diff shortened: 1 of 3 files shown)


AntsyLich:
what if we turn it into a BufferedInputStream

Shamicen:
> what if we turn it into a `BufferedInputStream`

If you mean using something like `BufferedInputStream()` or `.buffered()`, that does not work, is there another way to turn an `InputStream` into a `BufferedInputStream`?

AntsyLich:
> A `BufferedInputStream` adds functionality to another input stream-namely, the ability to buffer the input and to support the `mark` and `reset` methods.

Excert from https://developer.android.com/reference/java/io/BufferedInputStream

You can do `BufferedInputStream(existingInputStream)` to make a new one

Shamicen:
Y

> > A `BufferedInputStream` adds functionality to another input stream-namely, the ability to buffer the input and to support the `mark` and `reset` methods.
> 
> Excert from https://developer.android.com/reference/java/io/BufferedInputStream
> 
> You can do `BufferedInputStream(existingInputStream)` to make a new one

Yes but that will still consume the backing InputStream:
So the Buffer size needs to be set to a value that is always larger then the file you are reading.
Tested it with:
`BufferedInputStream(zip.getInputStream(entry), 8192)` -> does not load
`BufferedInputStream(zip.getInputStream(entry), 9000000)` -> does load

Shamicen:
Closed in favor of #485

PR #494, opened 2024-03-08
Fix shizuku being buggy for multi user setups

### LINKED ISSUE #493: Extension install or update via shizuku applied to main user when Mihon installed in secondary user
### Steps to reproduce

- Install Mihon in secondary user (I used island to create userID 10)
- Set Install method to Shizuku in settings and give Mihon permissions
- install an extension app
- update an extension app

### Expected behavior

installing an extension installs it for user 10 but not user 0
updating an extension only updates the extension

### Actual behavior

installing an extension installs it for user 0 but not user 10 (where Mihon lives), Mihon doesn't see the extension until it is manually mirrored
updating an extension installs it for user 0 if not installed there. As app versions are global the extension is still updated for user 10

### Crash logs

_No response_

### Mihon version

0.16.4

### Android version

Android 13 (lineageos 20)

### Device

Sony Xperia 1 III (with lineageos custom rom)

### Other details

[This file](https://github.com/mihonapp/mihon/blob/d8b9a9f593911569ff2bceb49b4f020978d0d2e1/app/src/main/java/eu/kanade/tachiyomi/ext%65nsion/installer/ShizukuInstaller.kt#L52) seems to run  
`pm install-create --user current -r -i ${service.packageName} -S $size`
using a root thread by shizuku. Is it possible that the `--user current` may be getting the user of the shizuku session, which is the main user 0?

### Acknowledgements

Redjard:
Fixes bug #493.
I have successfully tested this fix by building and installing the app on my phone.
Fetch the current userid separately because shizuku always runs as the main user and would otherwise install and update for the main user

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/extension/installer/ShizukuInstaller.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/extension/installer/ShizukuInstaller.kt
@@ -50,7 +50,6 @@ class ShizukuInstaller(private val service: Service) : Installer(service) {
             try {
                 val size = service.getUriSize(entry.uri) ?: throw IllegalStateException()
                 service.contentResolver.openInputStream(entry.uri)!!.use {
-                    
                     val userId = Process.myUserHandle().hashCode()
                     val createCommand = "pm install-create --user $userId -r -i ${service.packageName} -S $size"
                     val createResult = exec(createCommand)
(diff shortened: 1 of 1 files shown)


sirlag:
Closes #493

app/src/main/java/eu/kanade/tachiyomi/extension/installer/ShizukuInstaller.kt
```diff
@@ -49,7 +50,9 @@ class ShizukuInstaller(private val service: Service) : Installer(service) {
             try {
                 val size = service.getUriSize(entry.uri) ?: throw IllegalStateException()
                 service.contentResolver.openInputStream(entry.uri)!!.use {
-                    val createCommand = "pm install-create --user current -r -i ${service.packageName} -S $size"
+
```
AntsyLich, line 53:
```suggestion
```

PR #519, opened 2024-03-15
Proxy support

### LINKED ISSUE #366: Support for proxies (Socks5 if possible) per extension based or app based.

I think if possible adding proxies support per extension will be great or even with the app only. Some sources have very slow download speeds and using VPN does not always v due to cloudflare on some sources that I have encountered. I may be missing something, but just wanted to confirm if this will be a possibility.

From my limited knowledge of Android dev and googling, OkHttpClient does have proxy support.

### Other details

_No response_

### Acknowledgements

Shamicen:
Closes #366

Implemented proxy support. 

This is what the dialog looks like:


app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsAdvancedScreen.kt
```diff
@@ -254,6 +307,19 @@ object SettingsAdvancedScreen : SearchableSettings {
...
+                ),
+                Preference.PreferenceItem.SwitchPreference(
+                    pref = enableProxyGloballyPref,
+                    title = stringResource(MR.strings.pref_enable_proxy),
+                    onValueChanged = {
+                        context.toast(MR.strings.requires_app_restart)
+                        true
+                    },
+                    enabled = proxyString.isNotBlank(),
+                ),
```
AntsyLich, line 322: Make this part of the dialog

AntsyLich:
I did some thinking and how likely is it that average user will use a proxy? This feature will probably be used by <0.01% of users. Most will probably install some free vpn and use that. Though the feature is simple to implement do we really want another option that will not be used by most of the user base?

AntsyLich:
Closing based on my rationale

PR #633, opened 2024-04-07
chore: optimize query

Cologler:
optimize the number of SQL queries when creating backups.

Before: 

For each manga, execute one SQL query

Now: 

For all manga, execute one SQL query

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
@@ -9,6 +9,7 @@ import eu.kanade.tachiyomi.data.backup.models.backupChapterMapper
 import eu.kanade.tachiyomi.data.backup.models.backupTrackMapper
 import eu.kanade.tachiyomi.ui.reader.setting.ReadingMode
 import tachiyomi.data.DatabaseHandler
+import tachiyomi.domain.category.interactor.GetCategories
 import tachiyomi.domain.category.model.Category
 import tachiyomi.domain.history.interactor.GetHistory
 import tachiyomi.domain.manga.model.Manga
@@ -17,15 +18,12 @@ import uy.kohesive.injekt.api.get
 
 class MangaBackupCreator(
     private val handler: DatabaseHandler = Injekt.get(),
+    private val getCategories: GetCategories = Injekt.get(),
     private val getHistory: GetHistory = Injekt.get(),
 ) {
     suspend fun backupMangas(mangas: List<Manga>, options: BackupOptions): List<BackupManga> {
         val categoriesMap = if (options.categories) {
-            handler.awaitList {
-                categoriesQueries.getCategoriesWithMangaId { mangaId, categoryId, name, order, flags ->
-                    Pair(mangaId, Category(categoryId, name, order, flags))
-                }
-            }.groupBy({ it.first }, { it.second })
+            getCategories.awaitWithMangaId()
         } else {
             emptyMap()
         }
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
@@ -9,6 +9,7 @@ import eu.kanade.tachiyomi.data.backup.models.backupTrackMapper
 import eu.kanade.tachiyomi.ui.reader.setting.ReadingMode
 import tachiyomi.data.DatabaseHandler
 import tachiyomi.domain.category.interactor.GetCategories
+import tachiyomi.domain.category.model.Category
 import tachiyomi.domain.history.interactor.GetHistory
 import tachiyomi.domain.manga.model.Manga
 import uy.kohesive.injekt.Injekt
@@ -21,12 +22,15 @@ class MangaBackupCreator(
 ) {
 
     suspend fun backupMangas(mangas: List<Manga>, options: BackupOptions): List<BackupManga> {
+        val categoriesMap = if (options.categories) getCategories.awaitWithMangaId() else null
+
         return mangas.map {
-            backupManga(it, options)
+            backupManga(it, options, categoriesMap)
         }
     }
 
-    private suspend fun backupManga(manga: Manga, options: BackupOptions): BackupManga {
+    private suspend fun backupManga(manga: Manga, options: BackupOptions,
+                                    categoriesMap: Map<Long, List<Category>>?): BackupManga {
         // Entry for this manga
         val mangaObject = manga.toBackupManga()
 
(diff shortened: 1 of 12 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/MangaRestorer.kt
```diff
@@ -295,8 +297,11 @@ class MangaRestorer(
         categories: List<Long>,
         backupCategories: List<BackupCategory>,
     ) {
-        val dbCategories = getCategories.await()
-        val dbCategoriesByName = dbCategories.associateBy { it.name }
+        if (this.dbCategoriesMapByNameCache == null) {
```
sirlag, line 300: Having this inline with a force non-null is really messy. Put it behind a function which fills the cache on first call instead. then this reduces to something like
```kotlin
val dbCategoriesByName = dbCategoriesByName()
```
Even though it is still effectively a side effect, it makes it more clear that this is just a cache fill. I would have suggested turning it into a delegated property, but suspend makes it difficult.

app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
```diff
@@ -4,29 +4,51 @@ import eu.kanade.tachiyomi.data.backup.create.BackupOptions
...
-
     suspend fun backupMangas(mangas: List<Manga>, options: BackupOptions): List<BackupManga> {
+        val categoriesMap: Map<Long, List<Category>>? = if (options.categories) {
+            handler.awaitList {
+                categoriesQueries.getCategoriesWithMangaId { mangaId, categoryId, name, order, flags ->
+                    Pair(mangaId, Category(categoryId, name, order, flags))
+                }
+            }.groupBy({ x -> x.first }, { x -> x.second })
+        } else {
+            null
```
sirlag, line 30: Replace this with an empty map

app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
```diff
@@ -4,29 +4,51 @@ import eu.kanade.tachiyomi.data.backup.create.BackupOptions
...
+        }
+
+        val trackingMap = if (options.tracking) {
+            handler.awaitList {
+                manga_syncQueries.getTracks(backupTrackMapper)
+            }.groupBy({ x -> x.first }, { x -> x.second })
+        } else {
+            null
```
sirlag, line 38: Here as well, replace this with an empty map

app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
```diff
@@ -4,29 +4,51 @@ import eu.kanade.tachiyomi.data.backup.create.BackupOptions
...
-            backupManga(it, options)
+            backupManga(it, options, categoriesMap, trackingMap)
         }
     }
 
-    private suspend fun backupManga(manga: Manga, options: BackupOptions): BackupManga {
+    private suspend fun backupManga(
+        manga: Manga,
+        options: BackupOptions,
+        categoriesMap: Map<Long, List<Category>>?,
```
sirlag, line 49: Make these non-null

app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
```diff
@@ -48,16 +70,18 @@ class MangaBackupCreator(
         }
 
         if (options.categories) {
+            assert(categoriesMap != null)
```
sirlag, line 73: Remove the assert and null checks to the category map

app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
```diff
@@ -48,16 +70,18 @@ class MangaBackupCreator(
...
+            val categoriesForManga = categoriesMap!![manga.id]
+            if (categoriesForManga?.isNotEmpty() == true) {
                 mangaObject.categories = categoriesForManga.map { it.order }
             }
         }
 
         if (options.tracking) {
-            val tracks = handler.awaitList { manga_syncQueries.getTracksByMangaId(manga.id, backupTrackMapper) }
-            if (tracks.isNotEmpty()) {
+            assert(trackingMap != null)
```
sirlag, line 82: Same thing, remove the null checks

app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
```diff
@@ -4,29 +4,51 @@ import eu.kanade.tachiyomi.data.backup.create.BackupOptions
...
 class MangaBackupCreator(
     private val handler: DatabaseHandler = Injekt.get(),
-    private val getCategories: GetCategories = Injekt.get(),
     private val getHistory: GetHistory = Injekt.get(),
 ) {
-
     suspend fun backupMangas(mangas: List<Manga>, options: BackupOptions): List<BackupManga> {
+        val categoriesMap: Map<Long, List<Category>>? = if (options.categories) {
+            handler.awaitList {
+                categoriesQueries.getCategoriesWithMangaId { mangaId, categoryId, name, order, flags ->
```
sirlag, line 25: Move this up to  an interactor

sirlag (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/MangaRestorer.kt
Cologler: I have moved this up to `BackupRestorer`

Cologler:
Even after refactoring, `MangaBackupCreator.backupMangas()` still takes 1 minute to process my 506 mangas in DEBUG mode.

app/src/main/java/eu/kanade/tachiyomi/data/backup/create/creators/MangaBackupCreator.kt
Cologler: This looks unnecessary and overly designed, as it will only be used during backup (load all into memory).

Cologler (commented):
I have no idea why this takes 17 seconds

image

sirlag:
> I have no idea why this takes 17 seconds
> 
> image

If you profile your backups, where is it hanging?

Cologler:
I guess it's a debugger issue. After disconnecting the debugger, creating a backup is fast, almost close to 1 second.

Cologler:
Close now, as this may cause bugs. See #647.

PR #692, opened 2024-04-19
Use Coil pipeline instead of SSIV for image decode

FooIbar:
Follow-up of #691

This PR switchs to Coil pipeline for image decode in webtoon mode, images are now always loaded as a whole instead of tiles.
This significantly improves performance with extremely long (> 16k) images from testing.

Here is a benchmark build if anyone wants to test.
https://github.com/FooIbar/mihon/actions/runs/8765855174

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/coil/TachiyomiImageDecoder.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/coil/TachiyomiImageDecoder.kt
@@ -1,12 +1,16 @@
 package eu.kanade.tachiyomi.data.coil
 
+import android.graphics.Bitmap
 import coil3.ImageLoader
 import coil3.asCoilImage
 import coil3.decode.DecodeResult
+import coil3.decode.DecodeUtils
 import coil3.decode.Decoder
 import coil3.decode.ImageSource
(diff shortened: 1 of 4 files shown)


FooIbar:
Note the maximum width to not scale is set to `MAX_ZOOM_SCALE * screen width`, this might need to be adjusted.

wwww-wwww:
This is missing the subsampling which is to prevent heavy moire on screentones and halftones.

FooIbar:
> This is missing the subsampling which is to prevent heavy moire on screentones and halftones. Is there something that can be done to the imageview that can have it scale in linear light or do custom scaling?

Yeah that's a problem.
And it becomes worse with crop broders enabled, which wasn't handled well previously.

FooIbar:
> This is missing the subsampling which is to prevent heavy moire on screentones and halftones. Is there something that can be done to the imageview that can have it scale in linear light or do custom scaling?

Now the requested width is set to the width of the view, that's roughly what SSIV did with minimum tile dpi at 180 in webtoon mode.
The moire issue still exists, but at least it's not a regression.

wwww-wwww:
I will daily drive this and report back in a week

wwww-wwww:
works for me

PR #737, opened 2024-05-02
add flip image functionality

nopsqi:
sometimes there are chapters that are flipped
Screenshot_20240502_165917_Mihon
Screenshot_20240502_164846_Mihon


MajorTanya:
This would mirror the text on the images as well. I don't think there is any use for this setting.

PR #738, opened 2024-05-02
add flip image functionality

nopsqi:
sometimes there are chapters that are flipped
Screenshot_20240502_165917_Mihon
Screenshot_20240502_164846_Mihon

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/reader/settings/ColorFilterPage.kt
+++ b/app/src/main/java/eu/kanade/presentation/reader/settings/ColorFilterPage.kt
@@ -117,6 +117,10 @@ internal fun ColumnScope.ColorFilterPage(screenModel: ReaderSettingsScreenModel)
         label = stringResource(MR.strings.pref_inverted_colors),
         pref = screenModel.preferences.invertedColors(),
     )
+    CheckboxItem(
+        label = stringResource(MR.strings.pref_flip_horizontally),
+        pref = screenModel.preferences.flipHorizontally(),
+    )
 }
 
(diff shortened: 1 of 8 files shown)


nopsqi:
@MajorTanya yeah, but sometimes there are chapters that are flipped, so instead read from another source this will fix that

MajorTanya:
I've genuinely never encountered or heard a user mention that their chapter images are flipped. Also, please don't make new PRs of already existing ones.

I stand by what I said in #737:
> This would mirror the text on the images as well. I don't think there is any use for this setting.

I'll keep this open for now but don't hold your breath.

AntsyLich:
Only happens when sites are actively trying to mess with scrapers. And in the last 3 years I've only seen 3 sites do that and on all 3 cases the extension was updated to fix the issue. So no I'm not going to add this when this is fault of users poor choice extension

nopsqi:
> I've genuinely never encountered or heard a user mention that their chapter images are flipped
ngl, it's pretty rare. "The First Ancestor in HIstory" chapter 139 from xCaliBR Scans is one of the example

> Also, please don't make new PRs of already existing ones.
mb, accidentally deleted the repo

> I'll keep this open for now but don't hold your breath.
sure, thanks

nopsqi:
> Only happens when sites are actively trying to mess with scrapers. And in the last 3 years I've only seen 3 sites do that and on all 3 cases the extension was updated to fix the issue. So no I'm not going to add this when this is fault of users poor choice extension

ah okay, didn't know that

PR #742, opened 2024-05-02
Log exceptions in dumped crash logs

FooIbar:
So the user won't need to screenshot the crash screen.

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/crash/CrashScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/crash/CrashScreen.kt
@@ -37,7 +37,7 @@ fun CrashScreen(
         acceptText = stringResource(MR.strings.pref_dump_crash_logs),
         onAcceptClick = {
             scope.launch {
-                CrashLogUtil(context).dumpLogs()
+                CrashLogUtil(context).dumpLogs(exception)
             }
         },
         rejectText = stringResource(MR.strings.crash_screen_restart_application),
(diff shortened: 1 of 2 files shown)


MajorTanya:
The thing is, sometimes the exception does appear in the log file. When I first looked at this a while ago, I couldn't even reproduce the missing exception from the logs, even with a release build compile.

FooIbar:
> The thing is, sometimes the exception does appear in the log file. When I first looked at this a while ago, I couldn't even reproduce the missing exception from the logs, even with a release build compile.

It's just a few bytes more even if it's duplicate, I think we can afford that.

MajorTanya:
> > The thing is, sometimes the exception does appear in the log file. When I first looked at this a while ago, I couldn't even reproduce the missing exception from the logs, even with a release build compile.
> 
> It's just a few bytes more even if it's duplicate, I think we can afford that.

Oh, absolutely, I'd love to not have to ask people to give me the screenshot, only for them to have already restarted the app (or never respond in the first place).
Looking forward to this.

Would it be too much to put a "Latest Exception" header before it or something like that? To differentiate it from later ones

FooIbar:
> > > The thing is, sometimes the exception does appear in the log file. When I first looked at this a while ago, I couldn't even reproduce the missing exception from the logs, even with a release build compile.
> > 
> > 
> > It's just a few bytes more even if it's duplicate, I think we can afford that.
> 
> Oh, absolutely, I'd love to not have to ask people to give me the screenshot, only for them to have already restarted the app (or never respond in the first place). Looking forward to this.
> 
> Would it be too much to put a "Latest Exception" header before it or something like that? To differentiate it from later ones

The following log will have something like
```
--------- beginning of main
```

Issue #686, opened 2024-04-19
Add indication for failed updates on library page


destroyedData:
### Describe your suggested feature

**Issue:** When a manga fails to update for some reason (missing/broken source, missing chapters etc) it is difficult to see which entries in the library failed. 

**Current situation:** At this time, the problematic entries are just listed in a txt file which makes it a pain, making users swap between the txt and mihon to find the right library entry, then migrating that.

**My suggested solution:** It would be nice to have some indication on the library page so users can easily see what to migrate. I suggest to add something like a warning sign in the corner similar to the unread chapter count so users can tell at a glance if their favorite manga failed to update. Even better if filtering by these failed entries is possible.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.16.5](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

yatounoneko:
After a recent update, I found that version 0.16.5 could not update the missing komga. Even if the smart update is turned off, the status bar prompts the card to be stuck in a certain percentage and disappears. 

If you can add this function, I think it is very useful.

AntsyLich:
I would rather have #132

Issue #132, opened 2024-01-18
In app Error list


DDRook:
### Describe your suggested feature

when you update and get errors on certain chapters you get basically a text document of all the errors that occured such as

- Https 400
- No chapters found
- source URL has changed
- ETC

it would be lovely to be able to see every errored out title in a list in the app so we dont have to bounce back and forth between the document and the app to one by one fix them, instead a list in the app where we can tap and either migrate to a different source or find the new source URL or even just remove the title as a whole would be pretty handy.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an official extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/extensions/issues/new/choose).
- [X] I have updated the app to version **[0.16.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

Noriller:
Instead of a list... why not "error badges"? It could have a "❌" or something at the side of other badges.

On opening the title, the error will show with a description and maybe a how-to-solve helper.

A "has error" filter could also be available.

This way we could easily filter and manage any errors that appear.

DDRook:
This would also work for me

QuentinVdr:
Can we also have a list of the source with cloudflare issues to access all of them quickly without having to search in the library ?

PR #757, opened 2024-05-03
Add error badges and filter

komarnickolas:
Add error badges and filter to library.
2 issues asking for this kind of feature:
https://github.com/mihonapp/mihon/issues/132
https://github.com/mihonapp/mihon/issues/686

image
image
image
image
image


AntsyLich:
I would rather have a separate screen listing all failed updates. Nothing else nowhere else

AntsyLich:
Closing since I heard no response back or saw any activities.

PR #941, opened 2024-06-21
Upload build artifacts

FooIbar:
To decode obfuscated stack traces and help debugging R8 issues.


AntsyLich:
Shouldn't we upload the universal apk?

FooIbar:
> Shouldn't we upload the universal apk?

Unless you want to make it signed and available to users.
For debugging I don't think we need native libraries for other ABIs.

PR #1031, opened 2024-07-21
Hide smart update icon when disabled

perokhe:
It is only hidden when smart update is disabled AND there is no predicted release date.

When smart update is disabled that icon doesn't have much purpose other than maybe checking the prediction, but when there is no prediction it should be hidden.

| Before | After |
|:---:|:---:|
| Smart update visible | Smart update hidden |


AntsyLich:
Only hide when disabled

perokhe:
> Only hide when disabled

What do you mean? Hide if smart update is disabled regardless of if there's a predicted release date?

AntsyLich:
> > Only hide when disabled
> 
> What do you mean? Hide if smart update is disabled regardless of if there's a predicted release date?

Yes. Otherwise people might be confused why it isn't present for a handful of entries.

AntsyLich:
Actually I misunderstood the purpose of the PR. The PR is fine as-id but I want some more opinion on this change.

Xori71:
Okay, so what you want to do is have the hourglass icon disappear only when there’s no available prediction, right? (Assuming that the 4th smart update parameter is not checked)

Otherwise, even with the 4th parameter unchecked, the hourglass icon and predicted time window will still appear like normal.

Asking ’cause your wording has me a little confused.

perokhe:
That's currently how the PR is.

I was thinking that the prediction isn't even that reliable so maybe hide it altogether if the 4th parameter is unchecked would be a better idea.

But that's up to the mihon people to decide

Xori71:
> I was thinking that the prediction isn't even that reliable so maybe hide it altogether if the 4th parameter is unchecked would be a better idea.

Yes, I was thinking about this as well. It would also help users with configuring smart update, as the 4th parameter gives close to no feedback as to what it does.

“Predict next release time? Don’t I already have a time window prediction even with it off?”

kyoyacchi:
Any updates on this?

AntsyLich:
It's been a while but what's the issue with having it?

PR #1035, opened 2024-07-23
Fix disappearance items when fast scrolling

cuong-tran:
Due to update compose-bom and migration from `animateItemPlacement` to `animateItem` which by default includes fade-in/fade-out spec, while doing a fast scrolling, items look like disappeared.
=> Don't use fade-in/fade-out with `FastScrollLazyColumn`.


AntsyLich:
Since the same thing is being repeated make it an extension function. Can use the compose name `animateItemPlacement`

FooIbar:
Add related bug to the workaround function's comment.
https://issuetracker.google.com/352584409

cuong-tran:
> animateItemPlacement

since `animateItem` must be used in `LazyItemScope`, I don't know how to make an extension function of both LazyItemScope & Modifier at same time. Could you help with that?

AntsyLich:
Use context receivers. Example: https://github.com/mihonapp/mihon/blob/04aa5b36a5893ef9991312d61f84f830b83535f7/core/common/src/main/kotlin/eu/kanade/tachiyomi/network/OkHttpExtensions.kt#L134-L137

Issue #1023, opened 2024-07-17
App Crashes When Deleting an Untracked Entry After Removing it from Tracker


CasperKristiansson:
### Steps to reproduce

1. Track a specific Entry (Anilist for example)
2. Via the website (anilist) remove it
3. Go back to the app and check the status on the tracking (this will show a popup that "Could not find manga"
4. Try to delete the entry
5. Crash...

### Expected behavior

Shouldn't crash, just handle the exception that it isn't anymore tracked if the provider cant find itcan't

### Actual behavior

Crashes because it tries to remove a entry which is not being trackedan

### Crash logs

App version: 0.16.5 (standard, 2a01a2a, 6, 2024-04-09T09:58:14Z)
Android version: 14 (SDK 34; build UP1A.231005.007.G998BXXSBGXED)
Device brand: samsung
Device manufacturer: samsung
Device name: p3s (p3sxeea)
Device model: SM-G998B
WebView: Android System WebView 126.0.6478.134

--------- beginning of main
07-17 23:16:16.790 21016 21016 E app.mihon: Not starting debugger since process cannot load the jdwp agent.
07-17 23:16:25.793 21727 21727 E app.mihon: Not starting debugger since process cannot load the jdwp agent.
07-17 23:16:25.907 21727 21770 E SQLiteLog: (28) double-quoted string literal: ""
07-17 23:16:25.927 21727 21782 E app.mihon: No package ID 79 found for resource ID 0x790b0007.
07-17 23:16:25.927 21727 21782 E app.mihon: No package ID 79 found for resource ID 0x790b0007.
07-17 23:16:26.192 21727 21829 E SQLiteLog: (284) automatic index on C(manga_id)
07-17 23:16:26.192 21727 21829 E SQLiteLog: (284) automatic index on mangas_categories(manga_id)
07-17 23:21:41.505 24764 24764 E n:error_handler: Not starting debugger since process cannot load the jdwp agent.
07-17 23:21:41.586 24764 24789 E SQLiteLog: (283) recovered 939 frames from WAL file /data/data/app.mihon/databases/tachiyomi.db-wal
07-17 23:21:42.967 24940 24940 E app.mihon: Not starting debugger since process cannot load the jdwp agent.
07-17 23:21:43.034 24940 24973 E SQLiteLog: (5) statement aborts at 2: [PRAGMA journal_mode=TRUNCATE] database is locked
[...]

PR #1038, opened 2024-07-23
Solve Issue : App Crashes When Deleting an Untracked Entry After Removing it from Tracker #1023

TanishMoral11:
Solve Issue : App Crashes When Deleting an Untracked Entry After Removing it from Tracker #1023
Summary of MyCode
**1. ** TrackImpl Class:

Implements the Track interface with various properties (e.g., id, manga_id, tracker_id, etc.).
Initializes default values for these properties.
**2. ** deleteTrackedEntry Function:

Deletes a tracked entry, provided it exists.
Uses a try-catch block to handle any exceptions that may occur during the deletion process.
Calls checkEntryExistsInProvider to verify the entry's existence.
Deletes the entry from the database with deleteEntryFromDatabase.
Shows a toast message to inform the user of the result.
**3. ** checkEntryExistsInProvider Function:

Placeholder function to check if a tracked entry exists in the provider.
Currently returns false as a placeholder.
**4. ** deleteEntryFromDatabase Function:

Placeholder function to delete the entry from the local database.
Currently does not perform any actual deletion.
**5. ** showToast Function:

Displays a toast message to the user with a provided message.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/TrackImpl.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/TrackImpl.kt
@@ -1,30 +1,64 @@
 package eu.kanade.tachiyomi.data.database.models
 
+import android.content.Context
+import android.util.Log
+import android.widget.Toast
+
 class TrackImpl : Track {
 
     override var id: Long? = null
-
     override var manga_id: Long = 0
-
     override var tracker_id: Long = 0
-
     override var remote_id: Long = 0
-
     override var library_id: Long? = null
-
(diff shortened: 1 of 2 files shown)


Animeboynz:
I know very very little about kotlin, so forgive me if I'm wrong but should you be writing that inside the database models file? Instead of someplace like [here](https://github.com/mihonapp/mihon/blob/main/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt)

Soitora:
If this is a draft then please mark it as such, there's no way this is getting merged anywhere near its current form

PR #1044, opened 2024-07-24
Solve Issue : App Crashes When Deleting an Untracked Entry After Removing it from Tracker #1023

TanishMoral11:
Summary of Code Changes

File Modified: eu.kanade.tachiyomi.data.track.anilist.Anilist

Problem: The app crashes when trying to delete an untracked entry that has been removed from the tracker.

Solution: Added checks to ensure that the app does not attempt to delete an entry that is no longer tracked.

Changes Made
Added Check for Null or Zero libraryId:

Before attempting to delete an entry, the code checks if the libraryId is null or zero. This indicates that the entry might not be tracked.
Find Tracked Entry:

If libraryId is null or zero, the code uses api.findLibManga to search for the entry in the user's library. If the entry is not found, the deletion is skipped, preventing the crash.
Delete Tracked Entry:

If the entry is found, the code proceeds to delete it using api.deleteLibManga.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/TrackImpl.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/TrackImpl.kt
@@ -1,30 +1,64 @@
 package eu.kanade.tachiyomi.data.database.models
 
+import android.content.Context
+import android.util.Log
+import android.widget.Toast
+
 class TrackImpl : Track {
 
     override var id: Long? = null
-
     override var manga_id: Long = 0
-
     override var tracker_id: Long = 0
-
     override var remote_id: Long = 0
-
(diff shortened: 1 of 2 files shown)


AntsyLich:
All this code change just for a try-catch?

TanishMoral11:
> All this code change just for adding a try-catch?

I mistakenly changed another file as well, so I corrected it and then pushed it. That's why you see so many changes."

AntsyLich:
There's a bunch of change on the existing file. Also you have only "fixed" it for Anilist. And this is not the solution. If it fails to remove remote manga due to network errors we shouldn't just silently ignore it.

TanishMoral11:
So @AntsyLich , Can You Please Guide me the more requirements .. 
I wants to Contribute

PR #1056, opened 2024-07-26
Improve error message if restoring from JSON file

vetleledaal:
### Current state
Older versions of Tachiyomi support backing up and restoring from JSON files. If a user tries to restore from such a backup they should be nudged in correct direction. Currently the error message is very cryptic and leads to a support burden:
> kotlinx.serialization.protobuf.internal.ProtobufDecodingException: Unsupported start group or end group wire type: 3

### Proposed change
Users are likely to expect old JSON backups to still be usable in Mihon, unless told directly. With the proposed change there is a de minimis maintenance/performance burden as it hooks into the existing gzip signature check. The new error message:
> java.lang.Exception: JSON backup not supported

### Limitations
If the user picks a JSON file that is not a backup file from Tachiyomi the error message will falsely imply that the file is a valid Tachiyomi JSON backup file. Accounting for this would be non-feasible, and an unlikely scenario.

**[!]** I do not have an actual JSON backup file from Tachiyomi at hand, I assume it looks like a regular JSON file.

### Images
| Image 1 | Image 2 |
| ------- | ------- |
| before | after |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupDecoder.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/BackupDecoder.kt
@@ -11,6 +11,7 @@ import tachiyomi.core.common.i18n.stringResource
 import tachiyomi.i18n.MR
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
+import java.io.IOException
 
 class BackupDecoder(
     private val context: Context,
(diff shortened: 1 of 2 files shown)


MajorTanya:
They hadn't been usable in Tachiyomi for quite a while either afaik

BrutuZ:
How do people even manage to select these, aren't JSON files filtered out of the file picker already? 🤔

vetleledaal:
> They hadn't been usable in Tachiyomi for quite a while either afaik

Still, people do manage to run into these issues, and not everyone bothers asking for support.


> aren't JSON files filtered out of the file picker already?

It shows every file type for me ¯\\_(ツ)\_/¯

AntsyLich:
> How do people even manage to select these, aren't JSON files filtered out of the file picker already? 🤔

No file is filtered out

AntsyLich (commented):
Think you can also do a try catch on the decode function for`ProtobufDecodingEception` and say the backup file is corrupted?

BrutuZ:
> No file is filtered out

Could / should they be?

vetleledaal:
I have changed it so it will throw a more generic error in all cases. Should the `Full error:` string and the `java.io.IOException` also be changed/removed?

| Image 1, old JSON backup | Image 2, invalid file |
| ------- | ------- |
|test_json|test_other|

AntsyLich:
I dedicate this PR merge to @ThaDaVos

PR #1067, opened 2024-07-31
Add Crayola Blue Theme

1-minute-to-midnight:
Add the Crayola Blue Theme. Personally I think it's easier on the eyes and the white bold unread chapter number brings contrast with the chapter covers.

### Dark Mode
signal-2024-07-31-140321

### Light Mode
signal-2024-07-31-140313


AntsyLich:
Don't we have enough theme already

AntsyLich:
It's a no. Sorry.

PR #1082, opened 2024-08-06
Improve tag search from MangaScreen

weebsamachan:
If you click on a tag while on a MangaScreen you would normally be redirected back to the BrowseSourceScreen with the filter being adjusted correctly. But on some extensions the filter will be left untouched and the search query set to the tag instead.

This happens because those extensions use a text filter for their tags like "tag1,tag2,-negativetag". And that is not picked up by BrowseSourceScreenModel:searchGenre due to its messy way of searching for the tag filter.

The fix I have used is to search for text filters named "Tags" to give such extensions the possibility to have a correctly working tag search.

This is really is just a stopgap solution until you potentially add some way to uniquely identify filters based on their purpose, which would require the extensions to be changed as well.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreenModel.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreenModel.kt
@@ -201,6 +201,18 @@ class BrowseSourceScreenModel(
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


AntsyLich:
The implementation is stupid and very dependent on the filter to contain the text "Tags" which really isn't the case for the majority of them.

weebsamachan:
I suppose there are no plans to fix the existing stupid implementation then.

PR #1103, opened 2024-08-10
Use DTOs to parse tracking API responses

MajorTanya:
Changes the handling of tracker API responses to be parsed to DTOs instead of doing so "manually" by use of `jsonPrimitive`s and/or `Json.decodeFromString` invocations.

This greatly simplifies the API response handling.

Renamed constants to SCREAMING_SNAKE_CASE.

Largely tried to name the DTOs in a uniform pattern, with the tracker's (short) name at the beginning of file and data class names (ALOAuth instead of OAuth, etc).

With these changes, no area of the code base should be using `jsonPrimitive` and/or `Json.decodeFromString` anymore.


MajorTanya:
Checked MAL, AL, Kitsu, MU, and Shikimori, they all seem to work just fine.

I don't have a Bangumi account nor do I have Komga/Kavita/Suwayomi set up (although I didn't touch those last three, they should be the same as before).

Bangumi needs checking whether it works correctly (searching, adding/deleting, updating progress, status, scores, dates). Should also check if Mihon can correctly fetch data from Bangumi when the user opens the tracking sheet by manually changing data on the website first and seeing if Mihon correctly updates when the tracking sheet is opened.

MajorTanya:
Seems like Kitsu is going up in flames right after I checked this approach works. Presumably their API will be the same even if they switch domains but for the time being, that's another blocker for the PR I'd wager.

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistUtils.kt
```diff
@@ -0,0 +1,45 @@
+package eu.kanade.tachiyomi.data.track.anilist
+
+import eu.kanade.domain.track.service.TrackPreferences
+import eu.kanade.tachiyomi.data.database.models.Track
+import uy.kohesive.injekt.injectLazy
+import tachiyomi.domain.track.model.Track as DomainTrack
+
+fun Track.toAnilistStatus() = when (status) {
```
AntsyLich, line 8:
```suggestion
fun Track.toApiStatus() = when (status) {
```

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistUtils.kt
```diff
@@ -0,0 +1,45 @@
...
+    Anilist.DROPPED -> "DROPPED"
+    Anilist.PLAN_TO_READ -> "PLANNING"
+    Anilist.REREADING -> "REPEATING"
+    else -> throw NotImplementedError("Unknown status: $status")
+}
+
+private val preferences: TrackPreferences by injectLazy()
+
+@Suppress("MagicNumber", "CyclomaticComplexMethod")
+fun DomainTrack.toAnilistScore(): String = when (preferences.anilistScoreType().get()) {
```
AntsyLich, line 21:
```suggestion
fun DomainTrack.toApiScore(): String = when (preferences.anilistScoreType().get()) {
```

app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/BangumiUtils.kt
```diff
@@ -0,0 +1,12 @@
+package eu.kanade.tachiyomi.data.track.bangumi
+
+import eu.kanade.tachiyomi.data.database.models.Track
+
+fun Track.toBangumiStatus() = when (status) {
```
AntsyLich, line 5:
```suggestion
fun Track.toApiStatus() = when (status) {
```

app/src/main/java/eu/kanade/tachiyomi/data/track/kitsu/KitsuUtils.kt
```diff
@@ -0,0 +1,17 @@
+package eu.kanade.tachiyomi.data.track.kitsu
+
+import eu.kanade.tachiyomi.data.database.models.Track
+
+@Suppress("TooGenericExceptionThrown")
+fun Track.toKitsuStatus() = when (status) {
```
AntsyLich, line 6:
```suggestion
fun Track.toApiStatus() = when (status) {
```

app/src/main/java/eu/kanade/tachiyomi/data/track/kitsu/KitsuUtils.kt
```diff
@@ -0,0 +1,17 @@
...
+    Kitsu.READING -> "current"
+    Kitsu.COMPLETED -> "completed"
+    Kitsu.ON_HOLD -> "on_hold"
+    Kitsu.DROPPED -> "dropped"
+    Kitsu.PLAN_TO_READ -> "planned"
+    else -> throw Exception("Unknown status")
+}
+
+fun Track.toKitsuScore(): String? {
```
AntsyLich, line 15:
```suggestion
fun Track.toApiScore(): String? {
```

AntsyLich (changes requested):
Bangumi can be shortened to BGM

MajorTanya:
might be slightly blocked by #1130, depending on whether or not it happens

NeKoOuO:
@MajorTanya if you need a BGM account i can help you register one

MajorTanya:
Shikimori still crashes the app if a user tries to delete an entry from it when the remote entry doesn't exist anymore (user deleted it through other means like the website) due to an uncaught HttpException (404). This is unchanged to previous behaviour but should be addressed somehow.

AntsyLich (changes requested):

AntsyLich (approved):

PR #1113, opened 2024-08-11
Fetch release repository from mihonapp/kujaku instead of hardcoding

AntsyLich:

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/updater/AppUpdateChecker.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/updater/AppUpdateChecker.kt
@@ -1,10 +1,14 @@
 package eu.kanade.tachiyomi.data.updater
 
 import android.content.Context
+import eu.kanade.domain.base.BasePreferences
 import eu.kanade.tachiyomi.BuildConfig
 import eu.kanade.tachiyomi.util.system.isInstalledFromFDroid
+import tachiyomi.core.common.preference.Preference
 import tachiyomi.core.common.util.lang.withIOContext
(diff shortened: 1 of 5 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/updater/AppUpdateChecker.kt
```diff
@@ -42,14 +48,18 @@ class AppUpdateChecker {
...
     if (BuildConfig.PREVIEW) {
-        "mihonapp/mihon-preview"
+        "mihon-preview-0.x"
     } else {
-        "mihonapp/mihon"
+        "mihon-0.x"
     }
 }
 
+val GITHUB_REPO: Preference<String> by lazy {
```
Syer10, line 59: Its not a constant anymore so it should be renamed to something that doesn't give it the idea of one.

domain/src/main/java/tachiyomi/domain/release/interactor/GetApplicationRelease.kt
```diff
@@ -81,6 +83,8 @@ class GetApplicationRelease(
         val commitCount: Int,
         val versionName: String,
         val repository: String,
+        val changeRepository: (String) -> Unit = {},
+        val kujakuKey: String = "",
```
Syer10, line 87: Do these need a default, is GetApplicationRelease used anywhere else?

AntsyLich: in test

Syer10 (approved):

PR #1140, opened 2024-08-20
Adding bottom padding to Page Indicator to avoid phone's frame hide it

### LINKED ISSUE #1105: Page number not completely visible on some devices
### Steps to reproduce

1. Install the device on devices that have some type of small edge (maybe only from Samsung?)
2. Have the Show page number option enabled in general reading settings
3. Open some chapter

### Expected behavior

Being able to fully see the page number at the bottom of screen as the image below
<details><summary>Screenshot from my device with open chapter</summary>
<p>

image

</p>
</details> 

### Actual behavior

On the real device the page number is cut off, and it is not possible to read the numbers. It is hidden by the small edges that the device has. Both in the screenshot and in the emulator the number will appear correctly, so I will attach a photo of my real device with this problem.
<details><summary>Full Picture</summary>
<p>

image

</p>
</details> 

<details><summary>With zoom on page numbers</summary>
<p>

image

</p>
</details> 

### Crash logs

_No response_

### Mihon version

0.16.5

### Android version

Android 14

### Device

Samsung A54

### Other details

This issue seems to be related to the edges of the devices. I noticed that when taking a screenshot or mirroring my phone, it appears correctly at the bottom of the screen, but this is not the visible behavior on the actual device. For the same reason, emulating a device will display it correctly.

It won't be possible to fully demonstrate this behavior with screenshots from the devices, so I have attached a photo of my actual device showing this problem in the Actual Behavior section. Additionally, I will attach some related complaints that I found from other users both on Discord and in an Issue here on GitHub.

Issue on Github #526

<details><summary>Message on Support channel (Mihon Discord) </summary>
<p>

[https://discord.com/channels/1195734228319617024/1196111629482463422/1198382299763445871](https://discord.com/channels/1195734228319617024/1196111629482463422/1198382299763445871)

image

</p>
</details> 

### Acknowledgements

OdisBy:
This fixes #1105 

The Page Indicator at the bottom of the page was being cut in half because of Samsung's Phone Frame, and it is not possible to read the page number.

The padding I'm using is based on other places in the app that use the same padding at the bottom to avoid this.

note: This is my first PR to Mihon repository

### Images
| Image 1 | Image 2 |
| ------- | ------- |
mihon-no-padding |  |

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/reader/PageIndicatorText.kt
+++ b/app/src/main/java/eu/kanade/presentation/reader/PageIndicatorText.kt
@@ -1,16 +1,19 @@
 package eu.kanade.presentation.reader
 
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.drawscope.Stroke
 import androidx.compose.ui.text.TextStyle
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.tooling.preview.PreviewLightDark
+import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import eu.kanade.presentation.theme.TachiyomiPreviewTheme
 
@@ -35,6 +38,7 @@ fun PageIndicatorText(
     )
 
     Box(
+        modifier = Modifier.padding(bottom = 12.dp),
         contentAlignment = Alignment.Center,
     ) {
         Text(


app/src/main/java/eu/kanade/presentation/reader/PageIndicatorText.kt
```diff
@@ -35,6 +38,7 @@ fun PageIndicatorText(
     )
 
     Box(
+        modifier = Modifier.padding(bottom = 12.dp),
```
AntsyLich, line 41: This is not the way. You need to apply insets as padding instead of some arbitrary value. THo I've tried doing it but was unsuccessful.

OdisBy: I'm also having issues with WindowInsets. I guess it's because we're hiding the bottom buttons of the phone, and this causes it not to behave as usual.

But I'll try a few more ways to work around this without using a fixed value for the padding.

OdisBy: I couldn't come to a conclusion about which WindowInsets to use. We have `safeGestures` and `safeContent` that work, but I'm not sure if they are the most appropriate for this case. 

We also have `navigationBarsIgnoringVisibility`, which seems the most correct in this situation in my view, and it works very well when the phone is using the Navigation Bar with Swipe Gestures. However, when using the Navigation Bar with buttons, the padding becomes really large.

 Can you help me decide which direction to take?

AntsyLich:
Superseded by #1908 (probably)

PR #1184, opened 2024-09-02
Fix bgm tracker showing empty due to no Chinese name

NeKoOuO:
### Images

| Before | After |
| ------- | ------- |
|  |


AntsyLich:
I'll keep this in hold until #1103 is merged

NeKoOuO:
> I'll keep this in hold until #1103 is merged

I know it's okay 😌

AntsyLich:
is name not summary?

AntsyLich:
Found some more stuff that needs fixing. i'll takeover this

NeKoOuO:
> is name not summary?

What do you mean?

NeKoOuO:
> Found some more stuff that needs fixing. i'll takeover this

Yep, there's also a problem since the trackers have been create, the original site have changed by adding an NSFW option for entries, so the [API](https://bangumi.github.io/api/#/) has also changed. Now, if you want to track NSFW, the website will return empty results.

PR #1193, opened 2024-09-05
Ignore "intent://" urls on webview

bapeey:
Webview can't handle `intent://` urls, returning `true` cancels the request.
Issue example: https://github.com/keiyoushi/extensions-source/issues/3810


AntsyLich:
Shouldn't the `blob:http` case also return `true`?

bapeey:
> Shouldn't the `blob:http` case also return `true`?

Seems like webview can handle that

PR #1206, opened 2024-09-07
Fix: wrong calculation of nextUpdate when setting custom fetchInterval

cuong-tran:
it was using the negative value of custom fetchInterval to calculate nextUpdate and causing a nextUpdate date in the past

### DIFF
--- a/domain/src/main/java/tachiyomi/domain/manga/interactor/FetchInterval.kt
+++ b/domain/src/main/java/tachiyomi/domain/manga/interactor/FetchInterval.kt
@@ -107,7 +107,7 @@ class FetchInterval(
             interval.absoluteValue.takeIf { interval < 0 }
                 ?: increaseInterval(interval, timeSinceLatest, increaseWhenOver = 10),
         )
-        return latestDate.plusDays((cycle + 1) * interval.toLong()).toEpochSecond(dateTime.offset) * 1000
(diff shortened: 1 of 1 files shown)


domain/src/main/java/tachiyomi/domain/manga/interactor/FetchInterval.kt
```diff
@@ -107,7 +107,7 @@ class FetchInterval(
             interval.absoluteValue.takeIf { interval < 0 }
                 ?: increaseInterval(interval, timeSinceLatest, increaseWhenOver = 10),
         )
-        return latestDate.plusDays((cycle + 1) * interval.toLong()).toEpochSecond(dateTime.offset) * 1000
+        return latestDate.plusDays((cycle + 1) * interval.absoluteValue.toLong()).toEpochSecond(dateTime.offset) * 1000
```
AntsyLich, line 110: Should just fix this inside `calculateInterval`

cuong-tran: `calculateInterval` is used to automatically calculate the interval based on chapter's release date, when SmartUpdate is set to 'default'.

The case here is when user set a custom SmartUpdate, interval is set to a negative value and used as is, without calling to calculateInterval.

PR #1209, opened 2024-09-08
Fix WheelPicker Manual Input

### LINKED ISSUE #1147: Chapter count in tracker dialog not saved when writing instead of scrolling
### Steps to reproduce

1. Open the track info dialog of any manga
2. Try changing the chapter count by writing instead of scrolling
3. Click ok

### Expected behavior

The chapter count should be what I wrote

### Actual behavior

The chapter count doesn't change

### Crash logs

_No response_

### Mihon version

0.16.5-6825

### Android version

Android 10

### Device

Samsung Galaxy Note 9

### Other details

For some reason it only works when I hide the status bar.

https://github.com/user-attachments/assets/44d614b4-a8a8-4dba-a8d4-b0d6458ddb40

https://github.com/user-attachments/assets/7ac1e6ff-48be-42a2-b722-0657cec63cd3

### Acknowledgements

Animeboynz:
Resolves #1147 
Previously - Manual Input just doesn't work.

Now
- Manual Input works if the user press `Done` on the keyboard.
- Cant get it to work if the user presses back and closes the keyboard without pressing done

Either way better than it was before

### DIFF
--- a/presentation-core/src/main/java/tachiyomi/presentation/core/components/WheelPicker.kt
+++ b/presentation-core/src/main/java/tachiyomi/presentation/core/components/WheelPicker.kt
@@ -166,12 +166,8 @@ private fun <T> WheelPicker(
                 modifier = Modifier
                     .align(Alignment.Center)
                     .showSoftKeyboard(true)
-                    .clearFocusOnSoftKeyboardHide {
-                        processManualInput()
-                    },
-                onKeyboardAction = {
-                    processManualInput()
-                },
+                    .clearFocusOnSoftKeyboardHide { processManualInput() },
+                onKeyboardAction = { processManualInput() },
                 state = value,
                 lineLimits = TextFieldLineLimits.SingleLine,
                 keyboardOptions = KeyboardOptions(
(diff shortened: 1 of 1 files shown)


presentation-core/src/main/java/tachiyomi/presentation/core/components/WheelPicker.kt
```diff
@@ -166,6 +166,18 @@ private fun <T> WheelPicker(
...
+                    scope.launch {
+                        items
+                            .indexOfFirst { it.toString() == value.text }
+                            .takeIf { it >= 0 }
+                            ?.apply {
+                                internalOnSelectionChanged(this)
+                                lazyListState.scrollToItem(this)
+                            }
+                        showManualInput = false
+                    }
```
AntsyLich, line 179: since this is just a copy of the stuff above make this a lambda function and apply to both

presentation-core/src/main/java/tachiyomi/presentation/core/components/WheelPicker.kt
```diff
@@ -149,23 +149,29 @@ private fun <T> WheelPicker(
...
-                                    lazyListState.scrollToItem(this)
-                                }
-
-                            showManualInput = false
-                        }
+                        processManualInput()
                     },
+                onKeyboardAction = {
+                    processManualInput()
+                },
```
AntsyLich, line 174: how about this? same thing above
```suggestion
                onKeyboardAction = processManualInput,
```

PR #1297, opened 2024-10-05
Fix AniList `ALSearchItem.status` nullibility

Secozzi:
Fix anilist tracking search not working when the status is unavailable for a manga


AntsyLich:
maybe a better title?

MajorTanya:
Do you have an example title on AniList for this? SFW if at all possible

Secozzi:
> Do you have an example title on AniList for this? SFW if at all possible

https://anilist.co/manga/166123/Kidou-Senshi-Cross-Bone-Gundam-LOVEPIECE/

PR #1298, opened 2024-10-05
Fix EnhancedTracker not auto binding when adding manga to library

### LINKED ISSUE #306: Komga tracking not enabled by default when adding an entry to the library
### Steps to reproduce

1. Install de Komga extension (1.4.50) and setup
2. Explore the newly configured Komga server
3. Long tap on an entry to add it to the Mihon library
4. Open the newly added entry, tracking is not enabled by default

### Expected behavior

Step 3 should add the Komga's entry to Mihon library and enable tracking.

### Actual behavior

Step 3 adds the Komga's entry to Mihon library but does not enable tracking.

EDIT: I had the same issue with Tachiyomi.

### Crash logs

_No response_

### Mihon version

0.16.3

### Android version

Android 13

### Device

Pixel 4a

### Other details

_No response_

### Acknowledgements

brewkunz:
This PR is an attempt to fix the issue #306.

Tracking used to be enabled by default when adding an item to the library, which is not the case anymore.
This fix tries to bind with all logged In enhanced trackers upon adding an item to the library.

~~I think there still is an issue with the track being saved with a `last_read_chapter` value of 0.0, but the problem seems to lie in the `SyncChapterProgressWithTrack` and is not introduced by this fix.~~

### DIFF
--- a/app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
+++ b/app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
@@ -5,6 +5,7 @@ import eu.kanade.domain.track.model.toDomainTrack
 import eu.kanade.tachiyomi.data.database.models.Track
 import eu.kanade.tachiyomi.data.track.EnhancedTracker
 import eu.kanade.tachiyomi.data.track.Tracker
+import eu.kanade.tachiyomi.data.track.TrackerManager
 import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.util.lang.convertEpochMillisZone
 import logcat.LogPriority
@@ -14,17 +15,16 @@ import tachiyomi.core.common.util.system.logcat
 import tachiyomi.domain.chapter.interactor.GetChaptersByMangaId
 import tachiyomi.domain.history.interactor.GetHistory
 import tachiyomi.domain.manga.model.Manga
-import tachiyomi.domain.track.interactor.GetTracks
 import tachiyomi.domain.track.interactor.InsertTrack
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
 import java.time.ZoneOffset
 
 class AddTracks(
-    private val getTracks: GetTracks,
     private val insertTrack: InsertTrack,
     private val syncChapterProgressWithTrack: SyncChapterProgressWithTrack,
     private val getChaptersByMangaId: GetChaptersByMangaId,
+    private val trackerManager: TrackerManager,
 ) {
 
     // TODO: update all trackers based on common data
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
```diff
@@ -79,19 +79,19 @@ class AddTracks(
 
     suspend fun bindEnhancedTrackers(manga: Manga, source: Source) = withNonCancellableContext {
         withIOContext {
-            getTracks.await(manga.id)
```
brewkunz, line 82: This only returned already active tracks, which I don't think is intended.

app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
```diff
@@ -79,19 +79,19 @@ class AddTracks(
...
+            trackerManager.loggedInTrackers()
                 .filterIsInstance<EnhancedTracker>()
                 .filter { it.accept(source) }
                 .forEach { service ->
                     try {
                         service.match(manga)?.let { track ->
                             track.manga_id = manga.id
                             (service as Tracker).bind(track)
-                            insertTrack.await(track.toDomainTrack()!!)
+                            insertTrack.await(track.toDomainTrack(idRequired = false)!!)
```
brewkunz, line 90: This is now needed since the track might not have an id yet.

AntsyLich:
Good job trying to fix your own issue

app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
AntsyLich: Seems like the previous maintainer fucked up during a cleanup

app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
AntsyLich: the old implementation didn't need this (which actually did auto bind) either so should be fine? unless it was a issue during testing

brewkunz: It was indeed an issue during my tests. If the item you add to the library has no exisiting track with the tracker, the track has no id yet and `toDomainTrack` throws an exception.
It seems like this method allows for inserting and updating, but when you are inserting you have to set this parameter to false.

AntsyLich:
> I think there still is an issue with the track being saved with a last_read_chapter value of 0.0, but the problem seems to lie in the SyncChapterProgressWithTrack and is not introduced by this fix.

Should be a separate PR

brewkunz:
> > I think there still is an issue with the track being saved with a last_read_chapter value of 0.0, but the problem seems to lie in the SyncChapterProgressWithTrack and is not introduced by this fix.
> 
> Should be a separate PR

I moved this part to its new PR.

PR #1315, opened 2024-10-11
Added a random sort option to the library screen

jackhamilton:
Title. I've been suffering without this for a while so I decided to develop it myself. This is actually my first android PR ever, I'm an iOS dev in my day job, so go easy on me please <3

Tested manually, no unit tests because I've never done that for android and it intimidates me. Please double check any edge cases you can think of.

### Images
| Selected | Unselected |
| ------- | ------- |
|  |  |

### DIFF
--- a/gradle/kotlinx.versions.toml
+++ b/gradle/kotlinx.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-kotlin_version = "2.0.20"
+kotlin_version = "2.0.21"
 serialization_version = "1.7.3"
 xml_serialization_version = "0.86.3"
 


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
```diff
@@ -281,7 +281,7 @@ class LibraryScreenModel(
         val sortFn: (LibraryItem, LibraryItem) -> Int = { i1, i2 ->
             val sort = keys.find { it.id == i1.libraryManga.category }!!.sort
             when (sort.type) {
-                LibrarySort.Type.Alphabetical -> {
+                LibrarySort.Type.Alphabetical, LibrarySort.Type.Random -> {
```
jackhamilton, line 284: I'm using alphabetical sort as a fallback so this function doesn't need to be nullable, this code path should never be accessed because in the random case the value map doesn't fetch a comparator.

presentation-core/src/main/java/tachiyomi/presentation/core/components/SettingsItems.kt
```diff
@@ -115,6 +116,30 @@ fun SortItem(label: String, sortDescending: Boolean?, onClick: () -> Unit) {
     )
 }
 
+@Composable
+fun NondirectionalSortItem(label: String, enabled: Boolean, enabledIcon: ImageVector, onClick: () -> Unit) {
```
jackhamilton, line 120: IconItem has no provision for no icon but would otherwise work for this. I decided this was the least destructive way to do it, but I also don't have a lot of context for that decision so I welcome feedback.

jackhamilton:
Update: after some testing, the shuffle needs to maintain a consistent seed every time the button is pressed, as reloading should not cause the order to change. Will update and reopen when I get the chance.

jackhamilton:
> Update: after some testing, the shuffle needs to maintain a consistent seed every time the button is pressed, as reloading should not cause the order to change. Will update and reopen when I get the chance.

Had time to update, I'm using a single random seed for all categories. Behavior will be that the seed for all categories using random sort will change any time a section switches to random or taps on the randomization again. I believe this should line up with expected behavior and that users don't expect a distinctly saved random ordering for each tab, but I'm open to discussion there.

Using libraryPreferences for storage because it was the easiest, if it doesn't belong there let me know.

AntsyLich:
why?

jackhamilton:
> why?

If you have a large bulk from a source that uses something that interferes with alphabetical tagging - e.g. a comic source that leads its releases with the year ([2019], ex), the sort options will group every title like that together when, personally, I've found I sometimes want to view my library in a way that scatters them more evenly. Sometimes I don't want to view my stuff chronologically.

AntsyLich:
i'll review after a rebase

AntsyLich:
next time please don't open a pr from your main branch

jackhamilton:
replaced https://github.com/mihonapp/mihon/pull/1317

PR #1317, opened 2024-10-12
Added random library sort

jackhamilton:
Replacing https://github.com/mihonapp/mihon/pull/1315 because I was using my main branch like a goon

Problem:
If you have a large bulk from a source that uses something that interferes with alphabetical tagging - e.g. a comic source that leads its releases with the year ([2019], ex), the sort options will group every title like that together when, personally, I've found I sometimes want to view my library in a way that scatters them more evenly. Sometimes I don't want to view my stuff chronologically.
Also I just like random sort.

Behavior:
Using a single random seed for all categories. The seed for all categories using random sort will change any time a section switches to random or taps on the randomization again. I believe this should line up with expected behavior and that users don't expect a distinctly saved random ordering for each tab, but I'm open to discussion there.

### Images
| Selected | Unselected |
| ------- | ------- |
|  |  |

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/library/LibrarySettingsDialog.kt
+++ b/app/src/main/java/eu/kanade/presentation/library/LibrarySettingsDialog.kt
@@ -6,6 +6,8 @@ import androidx.compose.foundation.layout.ColumnScope
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.rememberScrollState
 import androidx.compose.foundation.verticalScroll
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Refresh
 import androidx.compose.material3.FilterChip
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
@@ -29,6 +31,7 @@ import tachiyomi.domain.library.service.LibraryPreferences
 import tachiyomi.i18n.MR
 import tachiyomi.presentation.core.components.CheckboxItem
 import tachiyomi.presentation.core.components.HeadingItem
+import tachiyomi.presentation.core.components.NondirectionalSortItem
 import tachiyomi.presentation.core.components.SettingsChipRow
 import tachiyomi.presentation.core.components.SliderItem
 import tachiyomi.presentation.core.components.SortItem
@@ -178,27 +181,42 @@ private fun ColumnScope.SortPage(
         MR.strings.action_sort_latest_chapter to LibrarySort.Type.LatestChapter,
         MR.strings.action_sort_chapter_fetch_date to LibrarySort.Type.ChapterFetchDate,
         MR.strings.action_sort_date_added to LibrarySort.Type.DateAdded,
+        MR.strings.action_sort_random to LibrarySort.Type.Random,
     ).plus(trackerSortOption).map { (titleRes, mode) ->
-        SortItem(
-            label = stringResource(titleRes),
-            sortDescending = sortDescending.takeIf { sortingMode == mode },
-            onClick = {
-                val isTogglingDirection = sortingMode == mode
-                val direction = when {
-                    isTogglingDirection -> if (sortDescending) {
-                        LibrarySort.Direction.Ascending
-                    } else {
-                        LibrarySort.Direction.Descending
-                    }
-                    else -> if (sortDescending) {
-                        LibrarySort.Direction.Descending
-                    } else {
-                        LibrarySort.Direction.Ascending
-                    }
-                }
-                screenModel.setSort(category, mode, direction)
-            },
-        )
+        when(mode) {
+            LibrarySort.Type.Random -> {
+                NondirectionalSortItem(
+                    label = stringResource(titleRes),
+                    enabled = sortingMode == LibrarySort.Type.Random,
+                    enabledIcon = Icons.Default.Refresh,
+                    onClick = {
+                        screenModel.setSort(category, mode, LibrarySort.Direction.Ascending)
+                    },
+                )
+            }
+            else -> {
+                SortItem(
+                    label = stringResource(titleRes),
+                    sortDescending = sortDescending.takeIf { sortingMode == mode },
+                    onClick = {
+                        val isTogglingDirection = sortingMode == mode
+                        val direction = when {
+                            isTogglingDirection -> if (sortDescending) {
+                                LibrarySort.Direction.Ascending
+                            } else {
+                                LibrarySort.Direction.Descending
+                            }
+                            else -> if (sortDescending) {
+                                LibrarySort.Direction.Descending
+                            } else {
+                                LibrarySort.Direction.Ascending
+                            }
+                        }
+                        screenModel.setSort(category, mode, direction)
+                    },
+                )
+            }
+        }
     }
 }
 
--- a/app/src/main/java/eu/kanade/presentation/library/LibrarySettingsDialog.kt
+++ b/app/src/main/java/eu/kanade/presentation/library/LibrarySettingsDialog.kt
@@ -29,9 +29,9 @@ import tachiyomi.domain.library.model.LibrarySort
 import tachiyomi.domain.library.model.sort
 import tachiyomi.domain.library.service.LibraryPreferences
 import tachiyomi.i18n.MR
+import tachiyomi.presentation.core.components.BaseSortItem
 import tachiyomi.presentation.core.components.CheckboxItem
 import tachiyomi.presentation.core.components.HeadingItem
-import tachiyomi.presentation.core.components.NondirectionalSortItem
 import tachiyomi.presentation.core.components.SettingsChipRow
 import tachiyomi.presentation.core.components.SliderItem
 import tachiyomi.presentation.core.components.SortItem
@@ -184,10 +184,14 @@ private fun ColumnScope.SortPage(
         MR.strings.action_sort_random to LibrarySort.Type.Random,
     ).plus(trackerSortOption).map { (titleRes, mode) ->
         if (mode == LibrarySort.Type.Random) {
-            NondirectionalSortItem(
+            val enabledIcon = if (sortingMode == LibrarySort.Type.Random) {
+                Icons.Default.Refresh
+            } else {
+                null
+            }
+            BaseSortItem(
                 label = stringResource(titleRes),
-                enabled = sortingMode == LibrarySort.Type.Random,
-                enabledIcon = Icons.Default.Refresh,
+                icon = enabledIcon,
                 onClick = {
                     screenModel.setSort(category, mode, LibrarySort.Direction.Ascending)
                 },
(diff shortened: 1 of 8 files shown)


domain/src/main/java/tachiyomi/domain/library/service/LibraryPreferences.kt
```diff
@@ -26,6 +26,8 @@ class LibraryPreferences(
         LibrarySort.Serializer::deserialize,
     )
 
+    fun currentRandomSortSeed() = preferenceStore.getInt("library_random_sort_seed", 0)
```
jackhamilton, line 29: Default of zero, but will be changed when the option is selected, so the seed should never actually be read as zero (though it's not a problem if it is)

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
```diff
@@ -267,7 +268,7 @@ class LibraryScreenModel(
 
         fun LibrarySort.comparator(): Comparator<LibraryItem> = Comparator { i1, i2 ->
             when (this.type) {
-                LibrarySort.Type.Alphabetical -> {
+                LibrarySort.Type.Alphabetical, LibrarySort.Type.Random -> {
```
jackhamilton, line 271: As fallback so this function doesn't have to be nullable. This is bypassed by the map function in the random case.

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
```diff
@@ -304,11 +305,16 @@ class LibraryScreenModel(
...
+            when (key.sort.type) {
+                LibrarySort.Type.Random -> value.shuffled(Random(libraryPreferences.currentRandomSortSeed().get()))
+                else -> {
+                    val comparator = key.sort.comparator()
+                        .let { if (key.sort.isAscending) it else it.reversed() }
+                        .thenComparator(sortAlphabetically)
+
+                    value.sortedWith(comparator)
+                }
+            }
```
AntsyLich, line 317:
```suggestion
           if (key.sort.type == LibrarySort.Type.Random) {
               return@mapValues value.shuffled(Random(libraryPreferences.currentRandomSortSeed().get()))
           }

            val comparator = key.sort.comparator()
                .let { if (key.sort.isAscending) it else it.reversed() }
                .thenComparator(sortAlphabetically)

            value.sortedWith(comparator)
```

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
AntsyLich: do something like 
```kt
LibrarySort.Type.Random -> {
    error("Should never throw")
}
```

app/src/main/java/eu/kanade/presentation/library/LibrarySettingsDialog.kt
```diff
@@ -178,27 +181,42 @@ private fun ColumnScope.SortPage(
...
+                                LibrarySort.Direction.Descending
+                            } else {
+                                LibrarySort.Direction.Ascending
+                            }
+                        }
+                        screenModel.setSort(category, mode, direction)
+                    },
+                )
+            }
+        }
```
AntsyLich, line 219:
```suggestion
        if (mode == LibrarySort.Type.Random) {
            NondirectionalSortItem(
                label = stringResource(titleRes),
                enabled = sortingMode == LibrarySort.Type.Random,
                enabledIcon = Icons.Default.Refresh,
                onClick = {
                    screenModel.setSort(category, mode, LibrarySort.Direction.Ascending)
                },
            )
            return@map
        }
        SortItem(
            label = stringResource(titleRes),
            sortDescending = sortDescending.takeIf { sortingMode == mode },
            onClick = {
                val isTogglingDirection = sortingMode == mode
                val direction = when {
                    isTogglingDirection -> if (sortDescending) {
                        LibrarySort.Direction.Ascending
                    } else {
                        LibrarySort.Direction.Descending
                    }
                    else -> if (sortDescending) {
                        LibrarySort.Direction.Descending
                    } else {
                        LibrarySort.Direction.Ascending
                    }
                }
                screenModel.setSort(category, mode, direction)
            },
        )
```

domain/src/main/java/tachiyomi/domain/library/model/LibrarySortMode.kt
```diff
@@ -31,6 +31,7 @@ data class LibrarySort(
         data object ChapterFetchDate : Type(0b00011000)
         data object DateAdded : Type(0b00011100)
         data object TrackerMean : Type(0b000100000)
+        data object Random : Type(0b000100100)
```
AntsyLich, line 34: I'm not well versed in bit masking but the mask probably needs to be changed?

presentation-core/src/main/java/tachiyomi/presentation/core/components/SettingsItems.kt
```diff
@@ -115,6 +116,30 @@ fun SortItem(label: String, sortDescending: Boolean?, onClick: () -> Unit) {
     )
 }
 
+@Composable
+fun NondirectionalSortItem(label: String, enabled: Boolean, enabledIcon: ImageVector, onClick: () -> Unit) {
```
AntsyLich, line 120: This should be made into `BaseSortItem` that takes in `label: String`, `icon: ImageVector?` and `onClick: () -> Unit`. SortItem should use `BaseSortItem`. Then in `LibrarySettingsDialog` create the random sort item using `BaseSortItem`

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
jackhamilton: Much better, thanks

domain/src/main/java/tachiyomi/domain/library/model/LibrarySortMode.kt
jackhamilton: Up to you! All that's going on here is that the bits in the mask that are one, e.g.

`override val mask: Long = 0b00111100L`

are relevant, everything else is ignored. What this means practically is that we only care about the four numbers in the middle and not the outer two on either side. These seem to just ascending integers in binary - 0, 1, 10, 11, 100, etc. The last one we have currently is TrackerMean at 1000 (8, as it's the 8th indexed element), so I put Random at 9 (1001). Factoring in the mask that's 0b00100100 - though I'm just now noticing that some of these have an extra leading zero, so I'll add a fix for that. Random is definitely arguably set apart from the rest, though, so if we wanted to make it 0b00111100 that wouldn't be a bad idea either.

jackhamilton: There are extra sort items in some of the forks and SY for example went with 00100100, same as here

jackhamilton:
Thanks for the review, learned return@ and that you can throw an error from a switch instead of making it nullable. Updates made.

domain/src/main/java/tachiyomi/domain/library/model/LibrarySortMode.kt
jackhamilton: Oh, also, unless I'm very wrong, removing the leading zero on the previous sort shouldn't change the value since it's leading, it'll just step it down to a eight bit value (which it is almost certainly being squashed to anyway when the flag is composed)

app/src/main/java/eu/kanade/presentation/library/LibrarySettingsDialog.kt
```diff
@@ -178,7 +181,23 @@ private fun ColumnScope.SortPage(
...
     ).plus(trackerSortOption).map { (titleRes, mode) ->
+        if (mode == LibrarySort.Type.Random) {
+            val enabledIcon = if (sortingMode == LibrarySort.Type.Random) {
+                Icons.Default.Refresh
+            } else {
+                null
+            }
+            BaseSortItem(
+                label = stringResource(titleRes),
+                icon = enabledIcon,
```
AntsyLich, line 194:
```suggestion
            BaseSortItem(
                label = stringResource(titleRes),
                icon = Icons.Default.Refresh
                    .takeIf { sortingMode == LibrarySort.Type.Random },
```

domain/src/main/java/tachiyomi/domain/library/model/LibrarySortMode.kt
AntsyLich: > Random is definitely arguably set apart from the rest, though, so if we wanted to make it 0b00111100 that wouldn't be a bad idea either.

Fair point. Bump it to max

AntsyLich (changes requested):

app/src/main/java/eu/kanade/presentation/library/LibrarySettingsDialog.kt
jackhamilton: good to know!

domain/src/main/java/tachiyomi/domain/library/service/LibraryPreferences.kt
AntsyLich, line 29:
```suggestion
    fun randomSortSeed() = preferenceStore.getInt("library_random_sort_seed", 0)
```

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
```diff
@@ -300,10 +301,18 @@ class LibraryScreenModel(
                     val item2Score = trackerScores[i2.libraryManga.id] ?: defaultTrackerScoreSortValue
                     item1Score.compareTo(item2Score)
                 }
+                LibrarySort.Type.Random -> {
+                    error("A comparator should not be requested for the random sort style. Instead, intercept this " +
+                        "case and call .shuffle()")
```
AntsyLich, line 306: As this is a local function it's fine to just have
```suggestion
                    error("Why Are We Still Here? Just To Suffer?")
```

domain/src/main/java/tachiyomi/domain/library/service/LibraryPreferences.kt
AntsyLich, line 29:
```suggestion
    fun randomSortSeed() = preferenceStore.getInt("library_random_sort_seed", 0)
```

AntsyLich (changes requested):
Besides this looks fine

AntsyLich:
ALso make sure to run `gradlew spotlessApply`

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
jackhamilton: lmao sounds good

jackhamilton:
> ALso make sure to run `gradlew spotlessApply`

Has this been considered as a precommit hook? Would save having to remember to do it

AntsyLich:
Just to confirm. Clicking "Random" while it's active changes the sort?

jackhamilton:
> Just to confirm. Clicking "Random" while it's active changes the sort?

That's right, yeah

jackhamilton:
swear I made that change already but that should fix (the?) failing unit test

AntsyLich (approved):

PR #1345, opened 2024-10-18
make global search pinned/all sticky

cuong-tran:
Some people don't have pinned sources at all, they don't want to use pinned sources feature. But Global Search or Migration search always start with Pinned search first.

This PR attempts to save the last used choice, whether it's Pinned or All search.

Users which although have pinned sources but often want to do All search can also get the benefit.

### DIFF
--- a/app/src/main/java/eu/kanade/domain/source/service/SourcePreferences.kt
+++ b/app/src/main/java/eu/kanade/domain/source/service/SourcePreferences.kt
@@ -1,6 +1,7 @@
 package eu.kanade.domain.source.service
 
 import eu.kanade.domain.source.interactor.SetMigrateSorting
+import eu.kanade.tachiyomi.ui.browse.source.globalsearch.SourceFilter
 import eu.kanade.tachiyomi.util.system.LocaleHelper
 import tachiyomi.core.common.preference.Preference
 import tachiyomi.core.common.preference.PreferenceStore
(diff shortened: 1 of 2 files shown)


AntsyLich:
- Untested PR
- They can just pin sources it's not that hard

NotMyAltAcc:
A PR being untested (and IDK if it is true in this case or not) is not something I have ever seen brought up in this project before, let alone minutes after opening.

It also doesn't seem to be that difficult to sync the toggle to a global preference.

BrutuZ:
> Some people don't have pinned sources at all

Some people are using the app wrong

AntsyLich:
@NotMyAltAcc your point being?

cuong-tran:
> * Untested PR
> * They can just pin sources it's not that hard

sorry, was coping but missed one change.

AntsyLich:
> > * Untested PR
> > * They can just pin sources it's not that hard
> 
> sorry, was coping but missed one change.

It's fine. Bit as I was saying I don't see the justification on adding this as people can just oij sources compared to has result option.

cuong-tran:
I added some more explanations, for people always want to search all sources or for people who don't want to have pinned sources. As for currently, they always have to tap twice to do a search.

Unless we are encouraging people to use pinned sources as a must to avoid spam multiple sources.

AntsyLich:
> Unless we are encouraging people to use pinned sources as a must to avoid spam multiple sources.

That's the intention. Most people (from what I've seen) have 10 to 20 sources installed.

cuong-tran:
> > Unless we are encouraging people to use pinned sources as a must to avoid spam multiple sources.
> 
> That's the intention. Most people (from what I've seen) have 10 to 20 sources installed.
> 
> 

Oh OK then, if this is not necessary then. Thanks.

AntsyLich:
I'm closing this. I suggest and plead you do the same.

PR #1364, opened 2024-10-24
avoid using runBlocking to load categories in Settings

cuong-tran:
`SettingsLibraryScreen` & `SettingsDownloadScreen` is using runBlocking to load categories list which will cause a bit lag while waiting for data from DB, especially for users with large library.


AntsyLich:
Just set the initial to an empty list

cuong-tran:
> Just set the initial to an empty list

then how to get the categories later?

AntsyLich:
> > Just set the initial to an empty list
> 
> then how to get the categories later?

`getCategories.subscribe()` will return the full category list swiftly. unless i'm mistaken

cuong-tran:
> > > Just set the initial to an empty list
> > 
> > 
> > then how to get the categories later?
> 
> `getCategories.subscribe()` will return the full category list swiftly. unless i'm mistaken

you are right, I forgot that part.

PR #1365, opened 2024-10-24
Rework Auto Track on Mark as Read

Animeboynz:
# Setting
image

## When off
image

## When on
image


BrutuZ:
Just a toggle? Weren't you adding a selection with the option to disable the snackbar too?

Animeboynz:
Was the plan and then antsy said no.

PR #1372, opened 2024-10-25
update all trackers progress when new tracker is added

cuong-tran:
Whenever a new tracker is add, it will try to sync all trackers together with local reading progress

### DIFF
--- a/app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
+++ b/app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
@@ -1,5 +1,6 @@
 package eu.kanade.domain.track.interactor
 
+import android.app.Application
 import eu.kanade.domain.track.model.toDbTrack
 import eu.kanade.domain.track.model.toDomainTrack
 import eu.kanade.tachiyomi.data.database.models.Track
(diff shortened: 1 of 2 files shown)


AntsyLich:
what's the reasoning?

cuong-tran:
for example, if user has 2 trackers but with different progress, when user adds the 2nd tracker with a higher progress, this will sync all trackers to that higher progress instead of waiting for the next sync.

Also, there is the TODO put right there state that.

app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
```diff
@@ -73,7 +77,23 @@ class AddTracks(
                 }
             }
 
-            syncChapterProgressWithTrack.await(mangaId, track, tracker)
+            val context = Injekt.get<Application>()
```
AntsyLich, line 80: Move to constructor

app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
```diff
@@ -73,7 +77,23 @@ class AddTracks(
...
+            refreshTracks.await(mangaId)
+                .filter { it.first != null }
+                .forEach { (track, e) ->
```
AntsyLich, line 83:
```suggestion
                .forEach { (track, e) ->
                    if (track == null) return@forEach
```

app/src/main/java/eu/kanade/domain/track/interactor/AddTracks.kt
```diff
@@ -73,7 +77,23 @@ class AddTracks(
...
+                        "Failed to refresh track data mangaId=$mangaId for service ${track!!.id}"
+                    }
+                    withUIContext {
+                        context.toast(
+                            context.stringResource(
+                                MR.strings.track_error,
+                                track!!.name,
+                                e.message ?: "",
+                            ),
+                        )
```
AntsyLich, line 94:
```suggestion
                        context.toast(context.stringResource(MR.strings.track_error, track.name, e.message ?: ""))
```

AntsyLich (changes requested):
Apply the suggestion to the rest of the changes and add changelog

BrutuZ:
Will it prompt the user? It should, there's the snackbar from "Update on manually mark as read" already, so could call that

cuong-tran:
This is wrong. Closing.

PR #1393, opened 2024-10-28
Trim whitespace from Repo URL and lowercase it

### LINKED ISSUE #1392: Repo URLs aren't lowercased or trimmed of whitespace
### Steps to reproduce

When adding a repo, there can be various issues with whitespace on the sides or a capitalised `Https`. Both of these end up with Mihon rejecting the link entirely.

### Expected behavior

The link should be trimmed and lowercased, then accepted.

### Actual behavior

Dialog closes and the "Invalid repo URL" toast pops up.

### Crash logs

_No response_

### Mihon version

0.17.0, probably before too

### Android version

any

### Device

any

### Other details

_No response_

### Acknowledgements

- [X] I have updated the app to version **[0.17.0](https://github.com/mihonapp/mihon/releases/latest)**.

MajorTanya:
This should close #1392.

Trimming whitespace and lowercasing the URL isn't an invasive procedure on the URL itself, and helps users who may have accidentally copied a space, or somehow capitalised or uppercased the URL because of an overzealous autocorrect, etc.

### DIFF
--- a/domain/src/main/java/mihon/domain/extensionrepo/interactor/CreateExtensionRepo.kt
+++ b/domain/src/main/java/mihon/domain/extensionrepo/interactor/CreateExtensionRepo.kt
@@ -14,11 +14,12 @@ class CreateExtensionRepo(
     private val repoRegex = """^https://.*/index\.min\.json$""".toRegex()
 
     suspend fun await(repoUrl: String): Result {
-        if (!repoUrl.matches(repoRegex)) {
+        val cleanedRepoUrl = repoUrl.trim().lowercase()
+        if (!cleanedRepoUrl.matches(repoRegex)) {
             return Result.InvalidUrl
         }
 
-        val baseUrl = repoUrl.removeSuffix("/index.min.json")
+        val baseUrl = cleanedRepoUrl.removeSuffix("/index.min.json")
         return service.fetchRepoDetails(baseUrl)?.let { insert(it) } ?: Result.InvalidUrl
     }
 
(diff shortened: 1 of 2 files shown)


domain/src/main/java/mihon/domain/extensionrepo/interactor/CreateExtensionRepo.kt
```diff
@@ -14,11 +14,12 @@ class CreateExtensionRepo(
     private val repoRegex = """^https://.*/index\.min\.json$""".toRegex()
 
     suspend fun await(repoUrl: String): Result {
-        if (!repoUrl.matches(repoRegex)) {
+        val cleanedRepoUrl = repoUrl.trim().lowercase()
```
AntsyLich, line 17: Lowercasing a url seems like a bad idea

MajorTanya: > Lowercasing a url seems like a bad idea

How about at least lowercasing the protocol bit?

Anutrix: https://www.rfc-editor.org/rfc/rfc3986

So, the scheme and host are case-insensitive.
The rest of the URL is case-sensitive.

MajorTanya: So basically, lowercasing the scheme should be fine. I'll adjust it in a moment.

MajorTanya: Fixed by case-insensitively replacing the first match for the `^https` regex.

PR #1396, opened 2024-10-28
Various improvements to Bangumi search

MajorTanya:
In short:
- fetch & show actual summary
- fallback to "name" if "name_cn" is empty
- request larger responseGroup to get & display the summary & rating
- add type filter query param to make Bangumi filter, not us

Previously, we only displayed the "name" in the summary area and used "name_cn" as the entry name. However, "name_cn" (Chinese name) can be an empty string at times, resulting in an awkward looking search result list where some, many, or even all the results have no title displayed and only show the "name" (Japanese name) in the summary area. This has been solved by using "name" as a fallback value should "name_cn" be empty.

By using the "responseGroup=large" query parameter, we can request the required data we need to display the actual summary for an entry and the entry's average rating.
The "name" is prepended to the summary contents, if any exist, so it is still accessible for series identification if a "name_cn" exists too and was used for the result title.

Adding the "type=1" filter query parameter means Bangumi will only return entries of type 1 ("book") instead of all types and Mihon needing to filter, resulting in potentially missed entry matches.

Sadly, Bangumi does not return any data on start or end dates in the search, no matter what responseGroup is used. Their data on this is sometimes contradictory at best anyway and often mixes Japanese and Chinese release dates.

| Old | New |
|--------|--------|
| image | image |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/dto/BGMSearch.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/dto/BGMSearch.kt
@@ -28,7 +28,11 @@ data class BGMSearchItem(
         remote_id = this@BGMSearchItem.id
         title = nameCn.ifEmpty { name }
         cover_url = images?.common ?: ""
-        summary = name + this@BGMSearchItem.summary?.let { "\n$it" }.orEmpty()
+        summary = if (nameCn.isNotEmpty()) {
+            "作品原名：$name" + this@BGMSearchItem.summary?.let { "\n$it" }.orEmpty()
+        } else {
+            this@BGMSearchItem.summary.orEmpty()
+        }
         score = rating?.score ?: -1.0
         tracking_url = url
         total_chapters = epsCount ?: 0
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/dto/BGMSearch.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/dto/BGMSearch.kt
@@ -26,9 +26,9 @@ data class BGMSearchItem(
 ) {
     fun toTrackSearch(trackId: Long): TrackSearch = TrackSearch.create(trackId).apply {
         remote_id = this@BGMSearchItem.id
-        title = nameCn.ifEmpty { name }
+        title = nameCn.ifBlank { name }
         cover_url = images?.common.orEmpty()
-        summary = if (nameCn.isNotEmpty()) {
+        summary = if (nameCn.isNotBlank()) {
             "作品原名：$name" + this@BGMSearchItem.summary?.let { "\n$it" }.orEmpty()
         } else {
             this@BGMSearchItem.summary.orEmpty()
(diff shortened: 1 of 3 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/dto/BGMSearch.kt
```diff
@@ -25,9 +26,9 @@ data class BGMSearchItem(
 ) {
     fun toTrackSearch(trackId: Long): TrackSearch = TrackSearch.create(trackId).apply {
         remote_id = this@BGMSearchItem.id
-        title = nameCn
+        title = nameCn.ifEmpty { name }
         cover_url = images?.common ?: ""
-        summary = this@BGMSearchItem.name
+        summary = name + this@BGMSearchItem.summary?.let { "\n$it" }.orEmpty()
```
AntsyLich, line 31: don't show the name?

MajorTanya: > don't show the name?

But the "name" field is the JP name. The search entry title is normally the Chinese name unless that doesn't exist, then it falls back to the JP name.
We've previously shown the JP name in the summary, thought it would be convenient for users to be able to see it still.

AntsyLich: Should probably have it as 
```
Native title: $name

$summary
```
Native title being in summary. Also maybe don't include it in summary if Chinese name is blank?

MajorTanya: Should it say "native title" or is there a was to make it a localised string? Or maybe put it in Chinese? 😅

AntsyLich: It's a Chinese tracker. Hard coding Chinese is fine

MajorTanya:
Rebased as well because why not

AntsyLich:
Where changelog

app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/dto/BGMSearch.kt
```diff
@@ -25,9 +26,13 @@ data class BGMSearchItem(
 ) {
     fun toTrackSearch(trackId: Long): TrackSearch = TrackSearch.create(trackId).apply {
         remote_id = this@BGMSearchItem.id
-        title = nameCn
-        cover_url = images?.common ?: ""
-        summary = this@BGMSearchItem.name
+        title = nameCn.ifEmpty { name }
+        cover_url = images?.common.orEmpty()
+        summary = if (nameCn.isNotEmpty()) {
```
AntsyLich, line 31:
```suggestion
        title = nameCn.ifBlank { name }
        cover_url = images?.common.orEmpty()
        summary = if (nameCn.isNotBlank()) {
```
```suggestion
        title = nameCn.ifEmpty { name }
        cover_url = images?.common.orEmpty()
        summary = if (nameCn.isNotEmpty()) {
```

AntsyLich (approved):

PR #1435, opened 2024-11-02
Always update fresh data from source if manga isn't favorited

### LINKED ISSUE #923: Prefer new data from source instead of db when browsing

when browsing, use new data from source instead of db for manga not in library.

### Other details

Say if a source has temporary urls for thumbnails, if the fetched manga are already in db (from previous browsing), it will try to load that thumbnail url which may have already expired, instead of the new one fetched now.

### Acknowledgements

cuong-tran:
while browsing source, new data (title, cover...) will be updated to db if it isn't favorited yet

Closes #923

### DIFF
--- a/domain/src/main/java/tachiyomi/domain/manga/interactor/NetworkToLocalManga.kt
+++ b/domain/src/main/java/tachiyomi/domain/manga/interactor/NetworkToLocalManga.kt
@@ -17,7 +17,13 @@ class NetworkToLocalManga(
             }
             !localManga.favorite -> {
                 // if the manga isn't a favorite, update new info from source to db
-                manga.updateManga(localManga.id)
+                mangaRepository.update(
+                    manga.toMangaUpdate()
+                        .copy(
+                            id = localManga.id,
+                            thumbnailUrl = manga.thumbnailUrl?.takeIf { it.isNotBlank() },
+                        ),
+                )
                 manga.copy(id = localManga.id)
             }
             else -> {
@@ -32,15 +38,5 @@ class NetworkToLocalManga(
 
     private suspend fun insertManga(manga: Manga): Long? {
         return mangaRepository.insert(manga)
-    }
-
-    private suspend fun Manga.updateManga(id: Long) {
-        mangaRepository.update(
-            toMangaUpdate()
-                .copy(
-                    id = id,
-                    thumbnailUrl = thumbnailUrl?.takeIf { it.isNotBlank() },
-                ),
-        )
     }
 }
--- a/domain/src/main/java/tachiyomi/domain/manga/interactor/NetworkToLocalManga.kt
+++ b/domain/src/main/java/tachiyomi/domain/manga/interactor/NetworkToLocalManga.kt
@@ -17,10 +17,18 @@ class NetworkToLocalManga(
             }
             !localManga.favorite -> {
                 // if the manga isn't a favorite, update new info from source to db
-                val mangaUpdate = manga.toMangaUpdate().copy(
-                    id = localManga.id,
-                    thumbnailUrl = manga.thumbnailUrl?.takeUnless { it.isBlank() },
-                )
+                val newThumbnail = manga.thumbnailUrl?.takeUnless { it.isBlank() }
+                val mangaUpdate = if (manga.initialized) {
+                    manga.toMangaUpdate().copy(
+                        id = localManga.id,
+                        thumbnailUrl = newThumbnail,
+                    )
+                } else {
+                    localManga.toMangaUpdate().copy(
+                        title = manga.title,
+                        thumbnailUrl = newThumbnail,
+                    )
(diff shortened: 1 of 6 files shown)


domain/src/main/java/tachiyomi/domain/manga/interactor/NetworkToLocalManga.kt
```diff
@@ -14,11 +14,6 @@ class NetworkToLocalManga(
                 val id = insertManga(manga)
                 manga.copy(id = id!!)
             }
-            !localManga.favorite -> {
-                // if the manga isn't a favorite, set its display title from source
-                // if it later becomes a favorite, updated title will go to db
-                localManga.copy(title = manga.title)
```
AntsyLich, line 20: Should just upsert the value here which also resolves #923

cuong-tran: Oh nice, my intention was really aim to load new cover until I found this line of code. Will implement that.

domain/src/main/java/tachiyomi/domain/manga/interactor/NetworkToLocalManga.kt
```diff
@@ -32,4 +33,14 @@ class NetworkToLocalManga(
     private suspend fun insertManga(manga: Manga): Long? {
         return mangaRepository.insert(manga)
     }
+
+    private suspend fun Manga.updateManga(id: Long) {
+        mangaRepository.update(
+            toMangaUpdate()
```
AntsyLich, line 39: Won't this override other fields? But i guess that's not an issue for non favorite manga?

cuong-tran: Any SManga field with non-null data will overwrite DB data. For non-favorite entries, it's will be fine, just like fresh browsing.

Other fields like reading mode, updated date... won't be affected

AntsyLich:
remove changelog update

domain/src/main/java/tachiyomi/domain/manga/interactor/NetworkToLocalManga.kt
```diff
@@ -32,4 +33,14 @@ class NetworkToLocalManga(
     private suspend fun insertManga(manga: Manga): Long? {
         return mangaRepository.insert(manga)
     }
+
+    private suspend fun Manga.updateManga(id: Long) {
```
AntsyLich, line 37: inline the code

domain/src/main/java/tachiyomi/domain/manga/interactor/NetworkToLocalManga.kt
```diff
@@ -15,9 +16,15 @@ class NetworkToLocalManga(
...
-                // if it later becomes a favorite, updated title will go to db
-                localManga.copy(title = manga.title)
+                // if the manga isn't a favorite, update new info from source to db
+                mangaRepository.update(
+                    manga.toMangaUpdate()
+                        .copy(
+                            id = localManga.id,
+                            thumbnailUrl = manga.thumbnailUrl?.takeIf { it.isNotBlank() },
+                        ),
+                )
```
AntsyLich, line 26:
```suggestion
                val mangaUpdate = manga.toMangaUpdate().copy(
                    id = localManga.id,
                    thumbnailUrl = manga.thumbnailUrl?.takeUnless { it.isBlank() },
                )
                mangaRepository.update(mangaUpdate)
```

AntsyLich:
What happens if you have a 'initialized' non library entry loaded with full data but while browsing source returns a not 'initialized' data 🤔

cuong-tran:
> What happens if you have a 'initialized' non library entry loaded with full data but while browsing source returns a not 'initialized' data 🤔

'initialized' is being used to make app check for manga's detail. If a 'initialized' non-library entry loaded with full data while source returns a not 'initialized' will make app auto refresh tags/descriptions/author... again when user decides to open the entry. Pretty mush like pull-refresh every time he open that entry (except for not checking chapters).

I think this will be useful too, since a non-library entry can also updates its details information.
Or we can skip overwriting 'initialized' to avoid that.

AntsyLich:
If the browse manga is initialized update everything otherwise only update the title and thumbnail

cuong-tran:
> If the browse manga is initialized update everything otherwise only update the title and thumbnail

although I made the change as per your request but I don't really get it though. When browsing, an non-initialized manga would also only has title & cover available, the rest are null. And null value won't overwrite already existed data of database's entries.

cuong-tran:
This seems cause performance issue while browsing caused by database overwrite. Need to find another way

AntsyLich:
i guess keep this draft until the db rework (#1195)

AntsyLich:
Superseded by #1967

PR #1445, opened 2024-11-05
fix: download job stops all together even if 1 entry is failed to create directory

cuong-tran:
Also, when some entry failed to create download folder, it was showing a notification saying only "invalid location: <tachi>/downloads" which is misleading.

This PR would:
- If Download folder missing or no permission allowed, it will still stop the whole Downloader & show notification: "Invalid download directory or No permission allowed"
- If only failed to create folder for 1 manga, it will skip that one, showing notification: "Invalid location: [source_dir/manga_dir]" and still continue other downloading


app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
```diff
@@ -45,7 +45,8 @@ class DownloadProvider(
             throw Exception(
                 context.stringResource(
                     MR.strings.invalid_location,
-                    downloadsDir?.displayablePath ?: "",
+                    (downloadsDir?.displayablePath ?: "") +
+                        "/${getSourceDirName(source)}/${getMangaDirName(mangaTitle)}",
```
AntsyLich, line 49: Imo we should try to figure if the issue is with the full path or we really can't write in the download dir

AntsyLich (commented):
And changelog

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadProvider.kt
cuong-tran: Isn't full path maximum length exceeded or any other issues are all meaning we can't write in the download dir?
Also, I'm not sure if the maximum length restriction is the same on various devices.

AntsyLich: by full path i meant until the manga dir. did the issue happen for the manga dir specifically or we can't write to the download dir in general

cuong-tran: or maybe can do this?
```
catch (e: IOException) {
        Log.e("DirectoryCreation", "IOException during directory creation: $directoryPath", e)
        if (e.message?.contains("too long") == true) {
            Log.e("DirectoryCreation", "Path length likely exceeded.")
        }
    } catch (e: SecurityException) {
        Log.e("DirectoryCreation", "SecurityException during directory creation: $directoryPath", e)
        Log.e("DirectoryCreation", "Permission issue likely.")
    } catch (e: Exception) {
        Log.e("DirectoryCreation", "Unexpected error during directory creation: $directoryPath", e)
    }
```

AntsyLich: Yeah something like that works

PR #1448, opened 2024-11-05
Add FOSS flavor

papjul:
This PR implements a FOSS flavor (i.e. with no proprietary dependencies) and builds it when a new version is tagged.
It's a pre-requisite for #1 although publishing it on F-Droid is not mandatory (some users may want to use the FOSS flavor with the internal updater, or something else like Obtainium). Note that compared to Tachiyomi, the updater can be left enabled this time, if you choose to make reproducible builds (which is the recommended way to publish on F-Droid).

Notes:
- I didn't bother building/testing the FOSS flavor everytime there is a commit or a new PR. The differences between the two flavors are so minor, that I didn't feel the additional build time was worth it.
- I only publish the universal APK for the FOSS flavor in GitHub releases, but I can add other ABIs if you want


.github/workflows/build_push.yml
```diff
@@ -93,6 +106,10 @@ jobs:
           sha=`sha256sum mihon-x86_64-${{ env.VERSION_TAG }}.apk | awk '{ print $1 }'`
           echo "APK_X86_64_SHA=$sha" >> $GITHUB_ENV
 
+          mv app/build/outputs/apk/foss/release/app-foss-universal-release-unsigned-signed.apk mihon-${{ env.VERSION_TAG }}.apk
```
folkeishii, line 109:
```suggestion
          mv app/build/outputs/apk/foss/release/app-foss-universal-release-unsigned-signed.apk mihon-foss-${{ env.VERSION_TAG }}.apk
```

papjul: Btw, why do you use `mv` for universal build and `cp` for ABIs?

AntsyLich:
FOSS build  setup can exist for F-Droid but I don't necessarily see a point of having them on GitHub.

papjul:
> FOSS build setup can exist for F-Droid but I don't necessarily see a point of having them on GitHub.

The point of having them on GitHub is to allow reproducible builds. Basically, you build and sign the release with your key, and F-Droid compares it with their own build. If they match, they will publish your release with your key. This makes the release process safer by making the developer sign the releases (and not F-Droid), while allowing F-Droid to still check for any tampering attempt.

AntsyLich:
FOSS builds can stay in a separate repo then similar to preview/beta

papjul:
Good idea! Should I revert the changes from the `.github` directory and let you create the other repo?

app/src/main/java/eu/kanade/presentation/more/onboarding/PermissionStep.kt
```diff
@@ -117,28 +118,30 @@ internal class PermissionStep : OnboardingStep {
                 },
             )
 
-            HorizontalDivider(
-                modifier = Modifier.padding(vertical = 8.dp, horizontal = 16.dp),
-                color = MaterialTheme.colorScheme.onPrimaryContainer,
-            )
+            if (!isFossFlavor) {
```
AntsyLich, line 121:
```suggestion
            if (isFossFlavor) return@Column 
```
and revert the other bit

app/build.gradle.kts
```diff
@@ -99,6 +99,10 @@ android {
             buildConfigField("boolean", "INCLUDE_UPDATER", "true")
             dimension = "default"
         }
+        create("foss") {
+            buildConfigField("boolean", "INCLUDE_UPDATER", "true")
+            dimension = "default"
```
AntsyLich, line 104:
```suggestion
            dimension = "default"
            applicationIdSuffix = "t-foss"
```

AntsyLich (changes requested):
Revert the workflow and download changes

app/src/main/java/eu/kanade/tachiyomi/data/updater/AppUpdateChecker.kt
```diff
@@ -45,6 +45,8 @@ class AppUpdateChecker {
 val GITHUB_REPO: String by lazy {
     if (BuildConfig.PREVIEW) {
         "mihonapp/mihon-preview"
+    } else if (BuildConfig.FLAVOR == "foss") {
+        "mihonapp/mihon-foss"
```
papjul, line 49: @AntsyLich Let me know if I need to change the repo name

papjul: @AntsyLich Do you want me to initialize a repo you will then fork? Let me know if I can help ;)

AntsyLich: Kinda busy and this is low priority. I'll get it merged soon

AntsyLich:
Superseded by #1775

Anything else that needs to be done before one can apply to be included in fdroid?

PR #1486, opened 2024-11-15
Add a setting for users to lower the threshold for HW bitmaps

### LINKED ISSUE #1436: Long strip sometime showing blank image
### Steps to reproduce

Reading with longstrip mode

### Expected behavior

Screenshot_20241103-020206

### Actual behavior

Screenshot_20241103-020200

### Crash logs

_No response_

### Mihon version

0.17.0

### Android version

android 10

### Device

Sony xz2 premium

### Other details

Strangely enough, turning on crop borders seem to fix it sometime, did i need to including the log for this issue?

### Acknowledgements

- [X] I have updated the app to version **[0.17.0](https://github.com/mihonapp/mihon/releases/latest)**.

Raikuha:
Allow users to configure a new value that will override maxTextureSize when deciding if a software bitmap should be replaced by a hardware one after being decoded, to prevent textures from failing on older devices. Default is maxTextureSize since that works for most people.

Fix #1436 with a better approach than #1440 since it's all up to the user. It references the way the user-agent setting is set, so if there's anything that should be changed or streamlined in the code I'm all ears.

The issue of blanks and flickering:

https://github.com/user-attachments/assets/a7286d96-2c6b-4281-8a79-56d15e950a22

What the setting looks like and how it works:

https://github.com/user-attachments/assets/70cc1832-54e3-4833-8b21-7fb6124dc77e

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsAdvancedScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsAdvancedScreen.kt
@@ -22,7 +22,6 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalUriHandler
 import androidx.core.net.toUri
-import androidx.core.text.isDigitsOnly
 import cafe.adriel.voyager.navigator.LocalNavigator
 import cafe.adriel.voyager.navigator.currentOrThrow
 import eu.kanade.domain.base.BasePreferences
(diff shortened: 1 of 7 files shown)


app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsAdvancedScreen.kt
```diff
@@ -345,6 +350,28 @@ object SettingsAdvancedScreen : SearchableSettings {
...
+                    },
+                ),
+                Preference.PreferenceItem.TextPreference(
+                    title = stringResource(MR.strings.pref_max_bitmap_size_reset),
+                    enabled = remember(maxBitmap) { maxBitmap != maxBitmapSize.defaultValue() },
+                    onClick = {
+                        maxBitmapSize.delete()
+                        context.toast(MR.strings.pref_max_bitmap_size_success)
+                    },
+                ),
```
AntsyLich, line 374: I'm not sure how i feel about an arbitrary value

Raikuha: @AntsyLich Which arbitrary value are you referring to? The default value is set as GLUtil.maxTextureSize when declaring the function in BasePreferences, so mantain the way it currently works for non-affected users (which are the majority of the userbase).

I took this approach precisely because I don't want to be hardcoding a value that might not cover all cases or that might cover more than needed. This would only allow the user to lower the threshold in the event that maxTextureSize results too permissive in practice, so unless someone changes this value, the decoder will work just as it does currently.

FooIbar: I'd say use a slider with limited values, like 4k, 8k, 16k, etc.

PR #1618, opened 2025-01-03
Fix MAL manga cover nullability

MajorTanya:
If a manga doesn't have a cover, MAL doesn't provide the `main_picture` element in the API response at all.

The long-unwanted sequel to the DTO migration fuckup saga. Please let this be the last one.


AntsyLich:
changelog

PR #1670, opened 2025-01-27
add multiple selection mode on browse source

L-oung:


MajorTanya:
What is this for?

L-oung:
> What is this for?

Batch add manga from sources

MajorTanya:
> > What is this for?
> 
> Batch add manga from sources

But you can already add several of them individually? Just long press on an entry and it'll add it to your default category/show the category selection screen. This new feature doesn't seem worth it in that respect.

L-oung:
> > > What is this for?
> > 
> > 
> > Batch add manga from sources
> 
> But you can already add several of them individually? Just long press on an entry and it'll add it to your default category/show the category selection screen. This new feature doesn't seem worth it in that respect.
> 
> Edit: And even if the long press option didn't exist, you can still open the entry with a tap, then favourite it, then tap back and continue through the source's listings.

Adding manga in large quantities is annoying:(

L-oung:
And the process of adding manga will become slower and slower. I dont know why

cuong-tran:
The last commit should go to a different PR, perhap?

MajorTanya:
It sure doesn't look related to this PR 🤔

AntsyLich:
Why is there entirely different feature changes in this PR?

PR #1681, opened 2025-01-30
Fix Bangumi tracking losing track of login state

MajorTanya:
kotlinx.serialization does NOT serialize default values (like createdAt in BGMOAuth.kt), so every time the Bangumi tracker deserialized the tracker OAuth, createdAt was set to the time of the read, not the time of issuance.

Separately, BangumiInterceptor did correctly fetch new OAuth credentials upon detected expiry of the stored credentials and saved them, but did not use them for the current request (the new credentials were used for all subsequent requests only). This led to 401 errors from Bangumi because the expired access_token was provided.
 A subsequent request using the newly acquired access_token would end
 up being successful.

Some version or another of this bug has been around since before #1103 so for once it's not me breaking things. 😅

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/BangumiInterceptor.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/BangumiInterceptor.kt
@@ -21,12 +21,13 @@ class BangumiInterceptor(private val bangumi: Bangumi) : Interceptor {
     override fun intercept(chain: Interceptor.Chain): Response {
         val originalRequest = chain.request()
 
-        val currAuth = oauth ?: throw Exception("Not authenticated with Bangumi")
+        var currAuth: BGMOAuth = oauth ?: throw Exception("Not authenticated with Bangumi")
 
         if (currAuth.isExpired()) {
             val response = chain.proceed(BangumiApi.refreshTokenRequest(currAuth.refreshToken!!))
             if (response.isSuccessful) {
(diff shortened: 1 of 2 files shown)


AntsyLich:
The doc is talking about serialization and we're doing deserialization the annotation has no effect here.

MajorTanya:
> The doc is talking about serialization and we're doing deserialization the annotation has no effect here.

It very much has. When we first request the data from Bangumi, we get a new instance of BGMOAuth, with the current timestamp added as createdAt. We proceed to serialise it in the trackPreferences (in the saveToken method). Here, we do not get createdAt stored. 

Once the app restarts or the Bangumi tracker initialises itself anew for any other reason, we pull the BGMOAuth Json from the trackPreferences, without createdAt. So, we get the default value provided in the instance. But this will be the current time, not token creation time.

With the annotation, createdAt gets stored when the BGMOAuth object gets serialised to the trackPreferences and therefore will be available on next restart/re-init/etc.

AntsyLich:
Hmm that's interesting. Not how I expected it to work.

MajorTanya:
I only figured this out because I had the same "Huh?" reaction with kotlinx.serialization on a different project before. When I checked the default assignment and looked into the preferences this gets serialised to, I suddenly rem and checked the docs again.

AntsyLich:
I was under the impression it would recheck the defaults but I suppose it's just easier to check if the property had custom value or not

MajorTanya:
It wouldn't know what the original timestamp value was, so it assigns the default again (the result of a assignment-time function call). This would be perfectly workable with constants but alas.

PR #1688, opened 2025-01-31
feat(ui): added a size slider for list display mode

### LINKED ISSUE #1687: Add a size slider for the List Display Mode

I've already done the same PR to the aniyomi fork, [here's how it looks with images.](https://github.com/aniyomiorg/aniyomi/pull/1906) I don't personally use mihon, but I don't mind adding the same feature here 'cause I really lacked it in the fork. Maybe mihon users would also appreciate it.

Ideally I would've liked to wait until it was approved in aniyomi before suggesting here, but there's no telling how long this is gonna take.

Pros of this change:
- being able to scale the list (duh)
- the long full names of mangas can fit, if upscaled enough
- visual change to match the grids (that's optional ofc, but it does look better like this imho).

My main question is: **Is this a good suggestion? Would you guys want it here?**

### Other details

_No response_

### Acknowledgements

MavikBow:
Copies a feature added to the aniyomi fork. [Here's a visual demonstration of what this adds.](https://github.com/aniyomiorg/aniyomi/pull/1906)

Added a size slider for the List Display Mode. It uses the same size variable as Grids do to determine size. However, the size of the List gets interpreted not as \<number\> of columns per row, but rather as \<number\> of entries of the list visible on the screen at once.

The style of the List was changed: the cover is of Book resolution to match the Grids and for better visibility.

Other two lists (browse list and migration list) don't have editable size, but neither do browse grids, so it's fine. Instead they now have a predetermined size of 76.dp (was 56.dp previously) and are visually changed to match the library list style. Same for the default size of the list.

The size slider also has two additional text fields added, which would need translation (though i assume that's automated).

The max line restriction for titles was removed, allowing for longer names to be fully visible, given the List is set to a big enough size.

closes #1687

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/browse/BrowseSourceScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/browse/BrowseSourceScreen.kt
@@ -41,6 +41,8 @@ fun BrowseSourceContent(
     source: Source?,
     mangaList: LazyPagingItems<StateFlow<Manga>>,
     columns: GridCells,
+    entries: Int = 0,
+    topBarHeight: Int = 0,
     displayMode: LibraryDisplayMode,
     snackbarHostState: SnackbarHostState,
     contentPadding: PaddingValues,
@@ -129,6 +131,8 @@ fun BrowseSourceContent(
         LibraryDisplayMode.List -> {
             BrowseSourceList(
                 mangaList = mangaList,
+                entries = entries,
+                topBarHeight = topBarHeight,
                 contentPadding = contentPadding,
                 onMangaClick = onMangaClick,
                 onMangaLongClick = onMangaLongClick,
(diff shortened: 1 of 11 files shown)


MavikBow:
I need to work on this a little more, so I'm closing this temporarily.

AntsyLich:
Just mark it as draft

MavikBow:
Sorry for the delay. I'm waiting when these changes will be merged the aniyomi fork. Because if my code would need corrections, I wouldn't want to bother both repositories to fixup them.

AntsyLich:
> Added a size slider for the List Display Mode. It uses the same size variable as Grids do to determine size. However, the size of the List gets interpreted not as <number> of columns per row, but rather as <number> of entries of the list visible on the screen at once.
> 

I don't really like this approach

MavikBow:
Well... could you go into more detail as to why you don't like it? The way I see it, the grids scale as a number per row, so imho that's the logical way to implement this for the list. 
If I went for a hardcoded size for all slider steps, several issues would emerge:
1\) might look bad on some devices,
2\) may needs further work if slider sees changes.

I guess in an ideal world the list would have to be rewritten as a one column grid to be able to utilize this Adaptive Cell feature (I hope I'm correct, I'm quite new to jetpack compose), but I don't need all lists to be affected, just the library and browse ones; nor does anyone have the time to do that.

What approach _do_ you like? I'll see what I can propose. I still want to able to scale the list, that's why I'm here.

AntsyLich:
I don't have any ideas at the moment but one thing I can say is that this should be a separate setting and you get to decrease the number of visible items not increase it.

PR #1695, opened 2025-02-03
Fix: null (visual) categories in Settings/Downloads/Exclude

cuong-tran:
image

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDownloadScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDownloadScreen.kt
@@ -107,6 +107,14 @@ object SettingsDownloadScreen : SearchableSettings {
         return Preference.PreferenceItem.MultiSelectListPreference(
             pref = downloadPreferences.removeExcludeCategories(),
             title = stringResource(MR.strings.pref_remove_exclude_categories),
+            subtitleProvider = { v, e ->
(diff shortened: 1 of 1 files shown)


MajorTanya:
Would it be possible to use slightly more informative names for v and e?

AntsyLich:
We should be cleaning up the invalid IDs instead of doing this

cuong-tran:
> We should be cleaning up the invalid IDs instead of doing this

they are not invalid, it's just that it's not ready loaded on first screen load so the `subtitleProvider` couldn't see map anything.

Original caused after this PR: #1364

cuong-tran:
> Would it be possible to use slightly more informative names for v and e?

I actually copied the default block from inside that method

MajorTanya:
> > Would it be possible to use slightly more informative names for v and e?
> 
> I actually copied the default block from inside that method

Would still be good if we had specific names in this location as opposed to the general purpose v, e in the block
That way it's actually readable in this state (right now I have no idea what v and e are)

AntsyLich:
Why is this an issue? Doesn't the items get replaced with actual name afterwards?

MajorTanya:
> Why is this an issue? Doesn't the items get replaced with actual name afterwards?

Are you talking about my comment or why the categories get listed as "null"?

AntsyLich:
I'm talking about the PR

AntsyLich:
Superseded by #1769

PR #1718, opened 2025-02-13
new features: migrate downloaded, download cbz in ram

fatotak:
Introduced 2 new features:
1. add migrate downloaded
2. add download cbz in ram

issues:
1. migrate downloaded vs delete downloaded functionality may be unclear to users.  What migrate downloaded does is to move/rename downloaded files from one source to the other via meta-data matching.  Any files that are not matched will remain in the source folder unchanged.  Delete downloaded will clean up these orphan files.
2. download cbz in ram toggle not disabled when cbz is disabled.  This function uses the java standard zip functionality rather than libarchive due to fstat not being available.

This change was done as a response to the recent decommissioning of mangasee/mangalife, where migrating from these sources to weeb central is a real pain factor.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,6 @@
 [*.{kt,kts}]
 max_line_length = 120
+indent_style = space
 indent_size = 4
 insert_final_newline = true
 ij_kotlin_allow_trailing_comma = true
--- a/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDownloadScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDownloadScreen.kt
@@ -46,6 +46,10 @@ object SettingsDownloadScreen : SearchableSettings {
                 pref = downloadPreferences.saveChaptersAsCBZ(),
                 title = stringResource(MR.strings.save_chapter_as_cbz),
             ),
+	        Preference.PreferenceItem.SwitchPreference(
+		        pref = downloadPreferences.saveChaptersCBZInRAM(),
+		        title = stringResource(MR.strings.save_chapter_cbz_in_ram),
+	        ),
             Preference.PreferenceItem.SwitchPreference(
                 pref = downloadPreferences.splitTallImages(),
                 title = stringResource(MR.strings.split_tall_images),
--- a/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDownloadScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDownloadScreen.kt
@@ -46,10 +46,10 @@ object SettingsDownloadScreen : SearchableSettings {
                 pref = downloadPreferences.saveChaptersAsCBZ(),
                 title = stringResource(MR.strings.save_chapter_as_cbz),
             ),
-	        Preference.PreferenceItem.SwitchPreference(
-		        pref = downloadPreferences.saveChaptersCBZInRAM(),
-		        title = stringResource(MR.strings.save_chapter_cbz_in_ram),
-	        ),
+            Preference.PreferenceItem.SwitchPreference(
+                pref = downloadPreferences.saveChaptersCBZInRAM(),
+                title = stringResource(MR.strings.save_chapter_cbz_in_ram),
+            ),
             Preference.PreferenceItem.SwitchPreference(
                 pref = downloadPreferences.splitTallImages(),
                 title = stringResource(MR.strings.split_tall_images),
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
@@ -38,435 +38,421 @@ import uy.kohesive.injekt.api.get
  * downloaded chapters.
  */
 class DownloadManager(
-	private val context: Context,
-	private val provider: DownloadProvider = Injekt.get(),
-	private val cache: DownloadCache = Injekt.get(),
-	private val getCategories: GetCategories = Injekt.get(),
-	private val sourceManager: SourceManager = Injekt.get(),
-	private val downloadPreferences: DownloadPreferences = Injekt.get(),
-	private val getChaptersByMangaId: GetChaptersByMangaId = Injekt.get(),
+    private val context: Context,
+    private val provider: DownloadProvider = Injekt.get(),
+    private val cache: DownloadCache = Injekt.get(),
+    private val getCategories: GetCategories = Injekt.get(),
+    private val sourceManager: SourceManager = Injekt.get(),
+    private val downloadPreferences: DownloadPreferences = Injekt.get(),
+    private val getChaptersByMangaId: GetChaptersByMangaId = Injekt.get(),
 ) {
 
-	/**
-	 * Downloader whose only task is to download chapters.
-	 */
-	private val downloader = Downloader(context, provider, cache)
+    /**
+     * Downloader whose only task is to download chapters.
+     */
+    private val downloader = Downloader(context, provider, cache)
 
-	val isRunning: Boolean
-		get() = downloader.isRunning
+    val isRunning: Boolean
+        get() = downloader.isRunning
 
-	/**
-	 * Queue to delay the deletion of a list of chapters until triggered.
-	 */
-	private val pendingDeleter = DownloadPendingDeleter(context)
+    /**
+     * Queue to delay the deletion of a list of chapters until triggered.
+     */
+    private val pendingDeleter = DownloadPendingDeleter(context)
 
-	val queueState
-		get() = downloader.queueState
+    val queueState
+        get() = downloader.queueState
 
-	// For use by DownloadService only
-	fun downloaderStart() = downloader.start()
-	fun downloaderStop(reason: String? = null) = downloader.stop(reason)
+    // For use by DownloadService only
+    fun downloaderStart() = downloader.start()
+    fun downloaderStop(reason: String? = null) = downloader.stop(reason)
 
-	val isDownloaderRunning
-		get() = DownloadJob.isRunningFlow(context)
+    val isDownloaderRunning
+        get() = DownloadJob.isRunningFlow(context)
 
-	/**
-	 * Tells the downloader to begin downloads.
-	 */
-	fun startDownloads() {
-		if (downloader.isRunning) return
+    /**
+     * Tells the downloader to begin downloads.
+     */
+    fun startDownloads() {
+        if (downloader.isRunning) return
 
-		if (DownloadJob.isRunning(context)) {
-			downloader.start()
-		} else {
-			DownloadJob.start(context)
-		}
-	}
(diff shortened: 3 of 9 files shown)


AntsyLich:
Don't make a single PR for multiple features

AntsyLich:
And do not change tab spacing and what not for no reason at all

fatotak:
1. They edit the same file.  What's the current practice for separate PR in this instance?
2. what do you mean? can you show me what you mean by this?

fatotak:
Added PR for .editorconfig changes.
https://github.com/mihonapp/mihon/pull/1720

Waiting on answer for Q.1. before proceeding with the other 2 branches.

MajorTanya:
> 2 branches from where? Both from main? Or one on top of the other?

For two independent features, yes, both from `main`. If one depends on the other, make both from `main` as well and mark the dependent second one as a draft.

MajorTanya:
Matching on numbers makes sense in a lot more cases. Usually, there is wide agreement on "this is chapter 23". Some may use split chapters and others full chapters, but full-on different numbering is exceedingly rare. Preludes and extras throwing off the numbering is an issue with that, yes, but that's an issue with how sources choose to misrepresent chapter count, not something Mihon would ever have any hopes of detecting, let alone handling.

Migration is not as niche as you make it out to be either. It's a useful tool for when sources break/die, are incomplete, or just plain bad quality. You know, the main reasons anyone would migrate. It moves over all sorts of previous settings like sorting, filters, tracker linkings, schedules, and of course reading progress.

Anyway.

My saying "I don't get it" was mainly about the (more common) cases where the two sources and their chapters differ in some way (translation, picture quality, extras, etc). In that situation, migrating downloaded chapters makes absolutely no sense.

> I spent a few weeks migrating them manually

_But why?_ Just delete the old files, migrate, and download a few chapters again?

> reduce sd card wear from writing all those temp files

See, you could even take this opportunity and enabling "Download ahead" & "Delete after read" so you can save yourself some storage, some write-wear on the SD card,  and also be kind to the source(s)?

> 2. Scanlator field (absent in the old source, present in weeb central)

Are you sure this isn't just down to a recent update to the extension rather than the website(s) changing behaviour?

> Anyway, my testing was done via 2 unrelated sources using a random title that I checked to have matched chapter numbering (by looking at page 2 of 2 random chapters on both sources to see they are the same):

I'm not entirely surprised that when checking big aggregators you ended up with the same content. But that is by far not a guarantee either. Even between aggregator sites, you'll commonly find chapters from different original sources. I strongly believe that your feature, _as nice and convenient as it may be for situations where it makes sense_ (reopenings, moves, aggregators with the same content), I don't think it's anything Mihon needs to include. The number of caveats attached to migrating of downloaded chapters limit the actually useful scope of this down to a feature that'd have to live under some sort of advanced setting to even show it. An average user will not think about any of them before just hitting "Migrate" and then continue to read their already downloaded chapters from Source A instead of the chapters provided by Source B.

Normally we don't want to see source names anywhere near our GitHub, in issues or PRs, so avoid them in the future. Their skill issues aren't our problem.

--- 

All that aside, there are still plenty of unrelated changes (yes autoformat does not work all the time, we know), especially in sections that break down chained functions for ease of reading without being long enough for the IDE to break the line itself, let alone breaking down chained functions (which is already a struggle to make happen in the IntelliJ/Android Studio fleet).
Like [this](https://github.com/mihonapp/mihon/pull/1718/files#diff-b7efc1a3b751dfb8880808c85ff68deffa9da639e180e7e01957d88a43184506L74-R86) which gets barely readable from its broken down form. A lot of changes across this PR are literally only down to this. One setting to avoid this I personally recommend is turning off "Remove user defined line breaks" or whatever that setting's name was in CTRL + ALT + SHIFT + L

Why did you refactor the flag checking [here](https://github.com/mihonapp/mihon/pull/1718/files#diff-b7efc1a3b751dfb8880808c85ff68deffa9da639e180e7e01957d88a43184506R38-L51)?
It's a very simple check that can stay as a repeated line in each checked flag anyway. The new functions does nothing useful, considering it wraps the `value and FLAGCONST` line.

MajorTanya:
You're free to make your own dedicated Mihon fork that includes this feature if you really need/want it but _personally_, I don't think this is something for Mihon. If you do that, make sure to change the relevant stuff like package names, app name, socials links, crash reporting, analytics, etc. since it'd be outside of Mihon's scope.

I don't know if the main maintainers have a different opinion so I'll defer to them on whether this will actually be closed or not.

fatotak:
> Migration is not as niche as you make it out to be either. It's a useful tool for when sources break/die, are incomplete, or just plain bad quality. You know, the main reasons anyone would migrate. It moves over all sorts of previous settings like sorting, filters, tracker linkings, schedules, and of course reading progress.

And when sources break/die, why wouldn't you migrate downloaded chapters.

> My saying "I don't get it" was mainly about the (more common) cases where the two sources and their chapters differ in some way (translation, picture quality, extras, etc). In that situation, migrating downloaded chapters makes absolutely no sense.

The more common case is that:
1. Source dies
2. Source is incomplete

Under both cases, migrate downloaded chapters will work perfectly.

For the picture quality/etc, no it won't work (just as migrate chapters won't work either - it won't magically tell you that you haven't read this particularly better quality release of a given chapter), but really... who's changing sources to get slightly better picture quality?

> _But why?_ Just delete the old files, migrate, and download a few chapters again?

You're assuming a tiny database.
If I delete the old files, migrate, and download again, it's going to:
1. Download for the next 8 weeks.
2. Erode even more write endurance from my storage.

If I migrate downloaded chapters:
1. I can continue reading downloaded chapters, without it spending the next 8 weeks downloading, and without it unnecessarily wearing out my sd card.

> See, you could even take this opportunity and enabling "Download ahead" & "Delete after read" so you can save yourself some storage, some write-wear on the SD card, and also be kind to the source(s)?

Both of which doesn't help with sd card write-wear AT ALL.
Mihon is pretty nasty in terms of sd card write wear.
Look at how it handles CBZ.
1. create temp folder
3. download individual images as temp files under the temp folder
4. rename the temp images
5. zip the temp folder
6. rename the temp zip file
7. delete the temp folder

Compared to downloading into RAM, compressing in RAM, and then writing the compressed to disk, it is using 2+ x the writes on data cells and 2 x page count + 4 more writes into the file system tables.  This change reduces disk wear to the minimum level: 1 entry in the file system table, and the minimum number of storage cells for the compressed data only.

It can lead to using more network (i.e. failed download -> restart app -> resume will start again, rather than finding the temp files from the last session, since temp files don't exist), but failed downloads will not trigger any writes, so sd card wear is preserved.

> Are you sure this isn't just down to a recent update to the extension rather than the website(s) changing behaviour?

It's one of the things I found during the manual renaming... and it doesn't matter if it's the extension update or the website update that's causing it.  It just matters that under 2 different sources, the downloaded chapter names are completely different, making it a massive pain to migrate download chapters manually.

fatotak:
> but _personally_, I don't think this is something for Mihon.

I don't think you are warranted, _at all_.
1. You're not a maintainer.
2. They are opt-in features (controlled by a checkbox/toggle) for those who may find it useful.  If you don't like it, then don't click on the checkbox.
3. 1 saves a lot of time for migrating (time saved for the user).
4. The other saves your sd card from being write worn prematurely ($$ for the user).
5. I repeat - it is OPT IN.  You need to check the checkbox for migration change, and select the toggle for the Download to RAM change.  When these are not set, the behaviour is same as before.

BrutuZ:
Looks opt-out to me
```suggestion
    fun saveChaptersCBZInRAM() = preferenceStore.getBoolean("save_chapter_cbz_in_ram", true)
```

Can't help but wonder about OoM crashes with this too. It's bound to happen when you add low-ish memory devices, users keeping 17 apps open and having a video on a pop-out player while downloading with this being on by default

fatotak:
> Looks opt-out to me

Oh... I'll change that to opt-in.  Good catch.

> Can't help but wonder about OoM crashes with this too. It's bound to happen when you add low-ish memory devices, users keeping 17 apps open and having a video on a pop-out player while downloading with this being on by default

Yep, that's a valid concern, and it will have to be up to the user to not use the feature if they don't have enough RAM.

fatotak:
Raised PRs for the split, so will be closing this one.

Download CBZ chapters in RAM: https://github.com/mihonapp/mihon/pull/1724
Migrate downloaded chapters: https://github.com/mihonapp/mihon/pull/1725

cuong-tran:
> > 1. Create 2 different branches. Each branch only changes what needed for 1 feature you want to make. Then from each branch create a PR for it's own.
> 
> 2 branches from where? Both from `main`? Or one on top of the other?
> 
> > 2. Don't reformat the file and change lines you don't really add changes beside from white space, tab.
> 
> The issue was that `.editorconfig` in `main` doesn't specify `indent_style` and I used `ctrl`+`alt`+`L`, which changed `spaces` to `tabs` (I use tabs for my own stuff). I'll separate the changes and push 3 updates soon (2 above + 1 for `.editorconfig`) once you answer the above question.
> 
> > 3. It can show update but you can't install it over your build.
> 
> ? What do you mean you can't install it over your build? Will the update fail? Or do you mean I shouldn't interact with that update button?

Because your build will have different signature from Mihon official build, so they just refuse to update over each other. You can create a build with different package name so it won't overwrite Mihon build, then use your build to migrate the downloaded chapter (using same folder for both build), maybe.

cuong-tran:
> > Because your build will have different signature from Mihon official build, so they just refuse to update over each other. You can create a build with different package name so it won't overwrite Mihon build, then use your build to migrate the downloaded chapter (using same folder for both build), maybe.
> 
> I can't do that. I need to write over each other, as I need access to the existing Mihon DB, which doesn't reside on shared storage. For the obsolete sources, the DB cannot be populated from source anymore, which is the data needed for migration.
> 
> Is there no way to force my fork over the official build?

no you can't force your fork over. But you can do this:
1. Create Mihon backup
2. Restore that backup onto your build. Set your app to using same storage path with Mihon (so that you can move downloaded file using your feature)
3. Make sure to Reindex download on your build.
4. Doing migration on your app
5. Create a backup from your app
6. Remove the entries from Mihon app then restore your backup (step 5) over it.

PR like yours probably will take really long time for it to be merged so better do the migration like that yourselves.

fatotak:
@cuong-tran 

Thank you!!!
I'll give that a shot.

My Mihon has been unusable for months, but it looks like I can make it usable again.
I'll probably transfer over to my fork and not transfer back.

I've always wanted to:
1. Address the excess write wear on SD card (Tachiyomi killed 2 over the last 10+ years, but I thought maintainers would someday put in this change... finally found time to do it myself).
2. Address the storage related slowness since Android 11.  Tachiyomi was blazing fast on Android 10, on a worse phone than now, but is slow since Android 11.  Upgrading phone didn't help at all, so I am 99% sure it is related to SAF (Storage Access Framework) introduced in Android 11.
3. fix the random crash bug that's been happening for years.

I've already addressed 1.
If I fix 2 and 3... I have no reason to come back to mainline.
I'll also remove all the useless features (privacy concerns like integration with trackers) if I go down this path.

PR #1724, opened 2025-02-13
Feature: Download CBZ chapters - tmp files in RAM

fatotak:
Split from https://github.com/mihonapp/mihon/pull/1718

This is a new opt-in feature to download chapters in RAM, rather than in storage.
This is strictly for the downloading and processing of _tmp files and folders.
It does not affect the final outcome (.cbz file on disk).

**This change does not impact non-cbz downloads.**

The current behaviour of mihon when downloading CBZ is:
1. create temp folder
2. download pages as temp files in storage
3. rename images
4. zip up the folder
5. rename the zip
6. delete the temp folder

As mobile phones and tablets are using flash storage, this download process erodes flash endurance more than necessary.

For people who have devices with sufficient RAM (RAM usage depends on number of sources allowed to concurrently download), all of the unnecessary flash endurance erosion can be avoided by:
1. Download pages into RAM
2. Zip the in-memory data to disk
3. Rename the zip file once it's ready for consumption (may not be necessary?)

This change requires users to have devices with sufficient RAM.

Changes since initial PR:
1. Hard memory limit added - 300Mb globally, 50Mb per chapter
1. Fallover routine added - if either global or per-chapter limit is exceeded, download via RAM will cancel and resume via _tmp file method instead.

Potential changes for consideration:
1. Make memory limits dynamic based on device memory (5% of device RAM as global limit, 1% of device RAM as chapter limit, if chapter limit < 50Mb, auto-turn off feature).
2. Opt-in/Opt-out default value dynamic dependent on device RAM.  (if device memory < 5Gb, default = off; else default = on)
3. Change to text description of toggle. (Download CBZ in RAM -> CBZ download optimization for high end devices)?

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-aboutlib_version = "11.5.0"
+aboutlib_version = "11.6.0"
 leakcanary = "2.14"
 moko = "0.24.5"
 okhttp_version = "5.0.0-alpha.14"


cuong-tran:
why download chapter in RAM? beside from the OOM, will the RAM downloaded cbz will be clear once app is closed and user will have to redownload again and hence more burden on the source?

Also, if the RAM-based CBZ is cleared because of app close, will the download index detect that to correctly display download status or the user will have to do a full `download reindex`?

cuong-tran:
I think is it because you are using SDCard and want it to download ahead without wearing down your SDCard? How about downloading ahead but not as CBZ but as cached of image files, in RAM or cached storage (phone storage)?

fatotak:
> why download chapter in RAM? beside from the OOM, will the RAM downloaded cbz be cleared once app is closed and user will have to redownload again and hence more burden on the source?

No.  The CBZ itself is saved to storage.
All of the temp files are handled in RAM, and cleared upon saving to storage.
If the downloader (which runs in the background) is force closed, yes, the current chapter in transit will need to start again.
However, any chapters already finished will reside in storage.
 
I am NOT expecting OOM from a majority of devices.
This processing adds ~chapter size x nr concurrent sources to RAM usage during download only.
It doesn't keep stuff in RAM indefinitely, only keeps it in RAM until the zip process is complete.

> Also, if the RAM-based CBZ is cleared because of app close, will the download index detect that to correctly display download status or the user will have to do a full `download reindex`?

The final CBZ resides on disk.
1. The image files are DOWNLOADED to RAM.
2. A stream zip to disk is created to archive the image data that is in RAM.
3. Indexes are updated after the zip stream is closed.

If the process is terminated before CBZ is saved to disk, then all changes are lost and indexes are not updated.
If the process is terminated after CBZ is saved to disk, then it's the same as legacy functionality, except that there were less writes to flash storage.

> I think is it because you are using SDCard and want it to download ahead without wearing down your SDCard? How about downloading ahead but not as CBZ but as cached of image files, in RAM or cached storage (phone storage)?

1. Lookhead doesn't help with flash wear.  Less write operations to storage is the only thing that will help.
2. Even phone internal storage is based on NAND flash technology and suffers flash wear.  It has better endurance than SD card, but ultimately, too much wear on internal storage will kill your phone altogether (as opposed to SD card which can be replaced).

This change is to minimize write to storage, not eliminate it completely.
It's not treating RAM as disk, but...

| Step | Legacy download method | This method |
| ----- | ----------------------------- | ---------------- |
| 1 | Create temp folder | skipped |
| 2 | Download image data **to disk** | download image data **to RAM** |
| 3 | Rename image files on disk | skipped |
| 4 | Zip image files **from disk** to disk | zip image files **from RAM** to disk |
| 5 | Rename zip file | same |
| 6 | Delete temp folder and temp files | skipped / image data in RAM discarded |
| 7 | Update indices to the cbz file on disk | same |

cuong-tran:
ah, so you mean the _tmp while downloading a chapter is put in RAM. The title might mislead.

fatotak:
> ah, so you mean the _tmp while downloading a chapter is put in RAM. The title might mislead.

Yes, all the _tmp stuff is in RAM, but the final output is same as before!

cuong-tran:
some titles, especially EH gallery I know can up to several hundreds MB or GB. That would be an issue.

fatotak:
> some titles, especially EH gallery I know can up to several hundreds MB or GB. That would be an issue.

For a single chapter???

AntsyLich:
I have some issues with this, app can only utilize a limited amount of ram which varies from device to device, it is volatile and affects user performance

fatotak:
> I have some issues with this, app can only utilize a limited amount of ram which varies from device to device, it is volatile and affects user performance

Then leave it off.  It's opt-in.

> I know can up to several hundreds MB or GB. That would be an issue.

I've done an incomplete change.
Change:
1. Introduced global hard coded limit of 300Mb for the chapter data cache.
2. Introduced per-chapter hard coded limit of 50Mb for the chapter data cache.
If either of the above parameters are exceeded by a chapter download, that chapter will spool downloaded files to disk, and then invoke the legacy download via storage function.  I have assumed the legacy function will pick up on the "already downloaded files" in the tmp folder and resume downloading from there.

Complete:
1. Confirmed working downloading via RAM.

Incomplete:
1. Need to somehow make the 2 limits configurable.
2. Unsure how to test if the failover is working.  Need to find a chapter big enough.

CrepeTF:
Even if OOM won't affect the majority of users, that still isn't good enough because the Mihon user-base is so large. There are a LOT of people using e-readers and cheap tablets which come with miniscule amount of ram already.

Even if it is opt-in only, most users don't explore the settings at all and wouldn't know to enable this setting (which is I'd say not strictly necessary to the reading experience itself). This would just mean another feature to maintain for oncoming versions which practically no one would use.

fatotak:
> Even if it is opt-in only, most users don't explore the settings at all and wouldn't know to enable this setting (which is I'd say not strictly necessary to the reading experience itself). This would just mean another feature to maintain for oncoming versions which practically no one would use.

As I said in the PR description, the primary purpose of the change is to minimize degradation of physical devices (phone, sd card) via reduction in write wear.

The secondary benefit is performance (SAF is SLOW, reducing storage access improves performance as a result).  Even reader performance will improve during downloads, as there is less that goes through the SAF.  But this is secondary.  Primary benefit is money.  You don't kill your phone storage - meaning you don't have to buy new phone, and you don't kill your SD card - meaning you don't have to spend money to buy new SD card; as long as you have enough RAM.

CrepeTF:
> Do you have any evidence to back this up?

I mean I am also a support staff in the server discord, and we do get a fair few people with even as low as 2GB of ram. Mihon is a pretty universally used application, especially in SEA (who are the largest demographic of Mihon users).

You are free to peruse the litany of devices that show up there.

CrepeTF:
There is a large overlap between users who have low ram and use sd cards (could say that practically all users who have low ram are a sizable subset of the users who use SD cards).

So even if they turn it on to save wear on their internal storage they are met with another problem that they do not have a lot of ram anyway.

You're also assuming that people would know about the setting to begin with if it were implemented. Which is gravely false. Vast majority of users will never explore the settings to even find out about it not, I very highly doubt that even 1% of the user-base would actively use this.

fatotak:
> You're also assuming that people would know about the setting to begin with if it were implemented. Which is gravely false. Vast majority of users will never explore the settings to even find out about it not, I very highly doubt that even 1% of the user-base would actively use this.

No, I'm not assuming anything.
I said it has the "potential" to benefit the "majority of users".
I didn't say it WILL benefit the "majority of users".
I don't claim to know how many people will turn it on.
I certainly will.

You're the one assuming people won't take it up, so it's not a useful feature.
1. It has a financial benefit to the user base who can and do use it.  This isn't an assumption.  It's fact.  It's just how flash storage works.
    1. The only people who absolutely won't benefit are:
        1. People who have <4Gb RAM
        1. People who run on emulators with images residing on magnetic drives instead of flash storage
    2. Everyone else CAN benefit, IF they turn it on.
1. Many users (whether they be majority or minority) have devices that can afford 300Mb memory usage.  This is also not an assumption.  It's fact.
1. By default, users will not be worse off, as the setting can be left off.  This is also not an assumption.  It's fact.

So there's only upside (better than before) and no downside (worse than before).
What's your issue again?

That maybe "not enough" people will use it?
1. So what?
2. What exactly is "not enough"?

I'm really not interested in arguing when you are asserting assumptions without any facts.
If you really want to continue to argue, please:
1. Read up on flash write wear.
2. Present arguments based on facts, not assumptions and opinions.

I AM keen to listen to feedback such as what cuong-tran provided:
1. problems with the code.
1. use cases that may not have been covered.
1. different approaches that solve the same problem.

AntsyLich:
If you're only issue is with flash write wear then just make a zip file and keep appending downloaded files to it. That is much more stable, less error prone and low maintenance burden compared to the current state of the PR.

fatotak:
> then just make a zip file and keep appending downloaded files to it. That is much more stable, less error prone and low maintenance burden compared to the current state of the PR.

It's something to consider, but in my view, it's not **more stable/less error prone**, but rather, just a different approach with different pros/cons.

Considering the primary vector of downloader being interrupted is the phone restarting (scheduled overnight reboot or user initiated), or user exiting wifi range...
1. keeping zip method is better than ram method for exiting wifi range.  Reason: while the user is out and about, the RAM usage won't elevate by the downloaded pages of incomplete chapters.
2. ram method is better than keeping zip method for phone restarting.  Reason: primary thing to be concerned about is the zip file header corruption during write (reboot in the middle of writing zip file header entries).  Writing in a shorter timespan (all at once) provides less probability of this happening than extending this time (over the course of the download).

But if you will merge the change if I adopt your approach, then I'll make the appropriate changes.
Please also confirm, if using zip file append method can I remove the toggle and make this the ONLY method for when CBZ download is selected?

cuong-tran:
Isn't phone restarting happening rarely comparing to phone in/out wifi range?
Also in case phone restarting, can just start zip file over to avoid corruption. My idea if you go with that method.

AntsyLich:
> But if you will merge the change if I adopt your approach, then I'll make the appropriate changes.
Please also confirm, if using zip file append method can I remove the toggle and make this the ONLY method for when CBZ download is selected?
>
We need to do some testing but yeah should be fine to make it part of the process instead of an option

fatotak:
> Isn't phone restarting happening rarely comparing to phone in/out wifi range? Also in case phone restarting, can just start zip file over to avoid corruption. My idea if you go with that method.

2 different issues, pro/con.
- In/out of wifi range pauses the download, doesn't corrupt anything.
Ram method holds onto the RAM (containing the already downloaded images) while download is paused as an extension of how it works.  i.e. if a 30mb chapter was being downloaded, and phone goes out of wifi 50% in, it will hold onto 15mb until wifi is back or allowing download on mobile data is selected by user.
- power loss corrupts stuff (i.e. phone restarting) that are mid-write.
This happens for both methods, but the dynamics of the write cycle is completely different (long period where there is no write, followed by a very short period where there is, compared to writes throughout).  During the long period, any power loss results in 0 corruption as there's no write.  The frequent writes for the zip append method makes it more likely that a write cycle and a power cycle intersect.  As for restarting the zip each time, that comes at it's own cost.

Basically every method has a cost... it's finding the cost that's most bearable.
Current method = cost is flash erosion
Ram method = cost is increase in RAM usage during download (both active and paused)
Zip append method = cost is more frequent resetting of interrupted download to cater for potential corruption.

Anyway given the above, I'll start working on zip append method tomorrow.

krysanify:
stupid question: how would this affect user who read and download a chapter at the same time?

i assume the reason we have a tmp folder and files is to differentiate between which page is ready to be displayed/read, and .tmp files are those in progress. same goes for writing the .cbz file separately, where the tmp folder is use for reading & maybe deleted after user done reading.

cuong-tran:
> > Isn't phone restarting happening rarely comparing to phone in/out wifi range? Also in case phone restarting, can just start zip file over to avoid corruption. My idea if you go with that method.
> 
> 2 different issues, pro/con.

What I meant is, preferring the method which providing pro for the cases which happens *more often* (wifi in/out vs. power loss/reboot)

fatotak:
1. lost my changes to (how do I undo this?):
fatotak [force-pushed](https://github.com/mihonapp/mihon/compare/fe731ece9789fb832ea1e6b9f16de697f584881a..8e81a5e68b61a7db36cd3ef39ac3f319c4d6e0a1) the feature/download_to_ram branch from [fe731ec](https://github.com/mihonapp/mihon/commit/fe731ece9789fb832ea1e6b9f16de697f584881a) to [8e81a5e](https://github.com/mihonapp/mihon/commit/8e81a5e68b61a7db36cd3ef39ac3f319c4d6e0a1) 
[3 hours ago](https://github.com/mihonapp/mihon/pull/1724#event-16325654497)
2. created a new PR to supercede this with the zip append method that was discussed between @AntsyLich and @cuong-tran above.  New PR: https://github.com/mihonapp/mihon/pull/1738

fatotak:
> stupid question: how would this affect user who read and download a chapter at the same time?
> 
> i assume the reason we have a tmp folder and files is to differentiate between which page is ready to be displayed/read, and .tmp files are those in progress. same goes for writing the .cbz file separately, where the tmp folder is use for reading & maybe deleted after user done reading.

I have no idea how the online reading works... best to ask an expert like @cuong-tran or @AntsyLich in this thread.

cuong-tran:
> 1. lost my changes to (how do I undo this?):

On you project folder, `git checkout fe731ece97 -b new_branch`.
the reference should still be there.

PR #1725, opened 2025-02-13
Feature: Migrate Downloaded Chapters

fatotak:
Split from https://github.com/mihonapp/mihon/pull/1718

This is a QoL improvement feature for source migration of manga.

Currently, downloaded chapters cannot be migrated, and they must either be:
1. Deleted and redownloaded
2. Migrated manually via file system operations

There are problems with both approaches.
1. Delete + Redownload
    1. Can take a long time to download again.
    2. Unnecessarily erodes flash storage endurance.
    3. Unnecessary internet bandwidth consumption.
2. Migrate Manually
    1. Very time consuming for the user.
        1. User must download a sample of chapters to figure out the file naming pattern under the new source
        2. Rename the chapters according to the new source's chapter naming pattern
        3. Move the renamed chapters to the correct folder under the new source.
    2. Cache is out of sync with application until cache is invalidated.

This feature addresses the above issues with migrate.

Sample title for migration
mihon_migrate_1

Files in storage before migration
mihon_migrate_1a

Migration interface (new checkbox for "Migrate downloaded")
mihon_migrate_2

Sample title post-migration
mihon_migrate_3

Files in storage post-migration
mihon_migrate_3a

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
@@ -1,6 +1,8 @@
 package eu.kanade.tachiyomi.data.download
 
 import android.content.Context
+import android.net.Uri
+import android.provider.DocumentsContract
 import eu.kanade.tachiyomi.data.download.model.Download
 import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.source.model.Page
@@ -16,10 +18,12 @@ import kotlinx.coroutines.runBlocking
 import logcat.LogPriority
 import tachiyomi.core.common.i18n.stringResource
 import tachiyomi.core.common.storage.extension
+import tachiyomi.core.common.storage.nameWithoutExtension
 import tachiyomi.core.common.util.lang.launchIO
 import tachiyomi.core.common.util.system.ImageUtil
 import tachiyomi.core.common.util.system.logcat
 import tachiyomi.domain.category.interactor.GetCategories
+import tachiyomi.domain.chapter.interactor.GetChaptersByMangaId
 import tachiyomi.domain.chapter.model.Chapter
 import tachiyomi.domain.download.service.DownloadPreferences
 import tachiyomi.domain.manga.model.Manga
@@ -40,6 +44,7 @@ class DownloadManager(
     private val getCategories: GetCategories = Injekt.get(),
     private val sourceManager: SourceManager = Injekt.get(),
     private val downloadPreferences: DownloadPreferences = Injekt.get(),
+    private val getChaptersByMangaId: GetChaptersByMangaId = Injekt.get(),
 ) {
 
     /**
@@ -235,6 +240,69 @@ class DownloadManager(
             // Delete manga directory if empty
             if (mangaDir?.listFiles()?.isEmpty() == true) {
                 deleteManga(manga, source, removeQueued = false)
+            }
+        }
+    }
+
+    /**
+     * Migrates manga from one source to another.
+     *
+     * @param oldManga the origin manga for migration..
+     * @param oldSource the origin source for migration..
+     * @param newManga the target manga for migration.
+     * @param newSource the target source for migration.
+     */
+    fun migrateManga(oldManga: Manga, oldSource: Source, newManga: Manga, newSource: Source) {
+        launchIO {
+            val oldMangaDir = provider.findMangaDir(oldManga.title, oldSource) ?: return@launchIO
+            val newMangaDir = provider.getMangaDir(newManga.title, newSource)
+            val downloadedChapters = oldMangaDir.listFiles()
+            val downloadedChaptersMap =
+                downloadedChapters?.associateBy({ it.nameWithoutExtension }, { it }) ?: return@launchIO
+
+            // map old chapters to new chapters.  The names may not be the same.
+            val oldChapters = getChaptersByMangaId.await(oldManga.id)
+            val newChapters = getChaptersByMangaId.await(newManga.id)
+            val oldChaptersByChapterNumber = oldChapters.associateBy { it.chapterNumber }
+
+            // If the chapter is downloaded, migrate it.
+            for (chapter in newChapters) {
+                val oldChapter = oldChaptersByChapterNumber[chapter.chapterNumber]
+                if (oldChapter != null) {
+                    val oldChapterName = provider.getChapterDirName(oldChapter.name, oldChapter.scanlator)
+                    val oldChapterFile = downloadedChaptersMap[oldChapterName]
+                    if (oldChapterFile != null) {
+                        val newChapterName = if (oldChapterFile.extension == null) {
+                            provider.getChapterDirName(chapter.name, chapter.scanlator)
+                        } else {
+                            provider.getChapterDirName(chapter.name, chapter.scanlator) + "." + oldChapterFile.extension
+                        }
+
+                        // move the file to newMangaDir, renaming the file/folder in the process
+                        var chapterUri: Uri? = oldChapterFile.uri
+
+                        // Rename the file.
+                        chapterUri = chapterUri?.let {
+                            DocumentsContract.renameDocument(
+                                context.contentResolver,
+                                it, newChapterName,
+                            )
+                        }
+
+                        // Move the file to the new source folder.
+                        chapterUri = chapterUri?.let {
+                            DocumentsContract.moveDocument(
+                                context.contentResolver,
+                                it, oldMangaDir.uri, newMangaDir.uri,
+                            )
+                        }
+
+                        // Can save chapterUri somewhere here if needed.
+
+                        cache.removeChapter(oldChapter, oldManga)
+                        cache.addChapter(newChapterName, newMangaDir, newManga)
+                    }
+                }
             }
         }
     }
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,9 @@
+# prevent user settings outside of the project from propagating into this project
+root = true
+
 [*.{kt,kts}]
 max_line_length = 120
+indent_style = space
 indent_size = 4
 insert_final_newline = true
 ij_kotlin_allow_trailing_comma = true
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/MigrationFlags.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/MigrationFlags.kt
@@ -30,24 +30,33 @@ object MigrationFlags {
     private const val CATEGORIES = 0b00010
     private const val CUSTOM_COVER = 0b01000
     private const val DELETE_DOWNLOADED = 0b10000
+    private const val MIGRATE_DOWNLOADED = 0b100000
 
     private val coverCache: CoverCache by injectLazy()
     private val downloadCache: DownloadCache by injectLazy()
 
     fun hasChapters(value: Int): Boolean {
-        return value and CHAPTERS != 0
+        return isFlagSet(value, CHAPTERS)
     }
 
     fun hasCategories(value: Int): Boolean {
-        return value and CATEGORIES != 0
+        return isFlagSet(value, CATEGORIES)
     }
 
     fun hasCustomCover(value: Int): Boolean {
-        return value and CUSTOM_COVER != 0
+        return isFlagSet(value, CUSTOM_COVER)
+    }
+
+    fun hasMigrateDownloaded(value: Int): Boolean {
+        return isFlagSet(value, MIGRATE_DOWNLOADED)
     }
 
     fun hasDeleteDownloaded(value: Int): Boolean {
-        return value and DELETE_DOWNLOADED != 0
+        return isFlagSet(value, DELETE_DOWNLOADED)
+    }
+
+    private fun isFlagSet(value: Int, flag: Int): Boolean {
+        return (value and flag) == flag
     }
 
     /** Returns information about applicable flags with default selections. */
@@ -61,6 +70,7 @@ object MigrationFlags {
                 flags += MigrationFlag.create(CUSTOM_COVER, defaultSelectedBitMap, MR.strings.custom_cover)
             }
             if (downloadCache.getDownloadCount(manga) > 0) {
+                flags += MigrationFlag.create(MIGRATE_DOWNLOADED, defaultSelectedBitMap, MR.strings.migrate_downloaded)
                 flags += MigrationFlag.create(DELETE_DOWNLOADED, defaultSelectedBitMap, MR.strings.delete_downloaded)
             }
         }
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/search/MigrateDialog.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/search/MigrateDialog.kt
@@ -212,6 +212,7 @@ internal class MigrateDialogScreenModel(
         val migrateChapters = MigrationFlags.hasChapters(flags)
         val migrateCategories = MigrationFlags.hasCategories(flags)
         val migrateCustomCover = MigrationFlags.hasCustomCover(flags)
+        val migrateDownloaded = MigrationFlags.hasMigrateDownloaded(flags)
         val deleteDownloaded = MigrationFlags.hasDeleteDownloaded(flags)
 
         try {
@@ -275,6 +276,13 @@ internal class MigrateDialogScreenModel(
         }
             .takeIf { it.isNotEmpty() }
             ?.let { insertTrack.awaitAll(it) }
+
+        // Migrate downloaded chapters
+        if (migrateDownloaded) {
+            if (oldSource != null) {
+                downloadManager.migrateManga(oldManga, oldSource, newManga, newSource)
+            }
+        }
 
         // Delete downloaded
         if (deleteDownloaded) {
--- a/i18n/src/commonMain/moko-resources/base/strings.xml
+++ b/i18n/src/commonMain/moko-resources/base/strings.xml
@@ -16,6 +16,7 @@
     <string name="manga">Library entries</string>
     <string name="chapters">Chapters</string>
     <string name="track">Tracking</string>
+    <string name="migrate_downloaded">Migrate downloaded</string>
     <string name="delete_downloaded">Delete downloaded</string>
     <string name="history">History</string>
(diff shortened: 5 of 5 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
```diff
@@ -239,6 +244,69 @@ class DownloadManager(
...
+    /**
+     * Migrates manga from one source to another.
+     *
+     * @param oldManga the origin manga for migration..
+     * @param oldSource the origin source for migration..
+     * @param newManga the target manga for migration.
+     * @param newSource the target source for migration.
+     */
+    fun migrateManga(oldManga: Manga, oldSource: Source, newManga: Manga, newSource: Source) {
+        launchIO {
```
cuong-tran, line 256: since you running `migrate downloaded` as an IO coroutine and right after that also running `delete downloaded` also using IO coroutine, if user selects both option, there is no way ensuring downloaded chapters will be deleted first or removed first.

you should do both action in same function and handle all cases where user select either 1 of those 2 options or select both.

fatotak: OK I'll make this change.

AntsyLich:
I'm against this whole feature altogether. Downloads are not the same across sources. There are also edge cases here and there. Instead we should introduce some kind of merging feature alongside say move downloads to local

fatotak:
> I'm against this whole feature altogether. Downloads are not the same across sources. There are also edge cases here and there. Instead we should introduce some kind of merging feature alongside say move downloads to local

If you're against this feature, why aren't you asking for Chapter reader marker migration to also be removed?  They share the same flaws.  A purist approach would have neither feature to cater for the risk of mismatched chapters due to the edge cases you cite.

This feature is to make migrate actually useful for if/when the user is willing to take the risk.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
fatotak: Change made... couldn't merge the 2 functions as the parameters are too different, so the change blast radius would be quite big (delete was being called in numerous places).

cuong-tran:
a few more points:
1. There is a lot cases where a manga having multiple chapters having same chapter number: different scanlators, different quality version, raw vs. translated, mis-match chapter number caused by volume number mixed in... While comparing old manga's chapter with new manga's chapter using chapter number might be the only way seems feasible, those cases I list will fail and overwrite downloaded files with each other.
2. While the chapter bookmark being migrated also has flaw like that but at least it's visible right there on the screen at a glance and easier for user to spot on, downloaded file will be hard for user to notice if it's really correct one or not unless they open & compare download files with online source (via webview?). A bit blindly maybe.
3. I also more of into the idea as Antsy (the maintainer) is moving the downloaded files to local source. Maybe still doing the rename to match new source's chapter name so user can decide to move the files into new manga's folder (at least the name is renamed for easier). Also, do a better handling of matching file name, for example if there is duplicated chapter number then keep old name instead of rename (to avoid overwrite each other), perhap?

AntsyLich:
> If you're against this feature, why aren't you asking for Chapter reader marker migration to also be removed? They share the same flaws.
> 
No it doesn't. Chapter migration doesn't try to migrate stuff related to content like reading progress or history of a chapter
> A purist approach would have neither feature to cater for the risk of mismatched chapters due to the edge cases you cite.
>
I hadn't even mentioned the edge cases you're just assuming. @cuong-tran has them in his point 1. And as you can none of them affects chapter migration in its current state
> This feature is to make migrate actually useful for if/when the user is willing to take the risk.
>
So according to you the migrate feature at its current state is useless? Ok let's just assume what you said is true but if it comes with a huge maintenance burden (edge cases says hi) and there are alternative solutions which is simpler (i.e. entry merging) then I'm not gonna merge it even if it's a revolutionary feature.

fatotak:
> > If you're against this feature, why aren't you asking for Chapter reader marker migration to also be removed? They share the same flaws.
> 
> No it doesn't. Chapter migration doesn't try to migrate stuff related to content like reading progress or history of a chapter
> 
> > A purist approach would have neither feature to cater for the risk of mismatched chapters due to the edge cases you cite.
> 
> I hadn't even mentioned the edge cases you're just assuming. @cuong-tran has them in his point 1. And as you can none of them affects chapter migration in its current state
> 
> > This feature is to make migrate actually useful for if/when the user is willing to take the risk.
> 
> So according to you the migrate feature at its current state is useless? Ok let's just assume what you said is true but if it comes with a huge maintenance burden (edge cases says hi) and there are alternative solutions which is simpler (i.e. entry merging) then I'm not gonna merge it even if it's a revolutionary feature.

1. When I placed a read marker on the chapter, and then migrated the manga, the migrated manga had that manga as having been read.  So what do you mean it doesn't come with the same flaws?
2. The existing function has the same flaws. (I.e  you didn't read this version of the chapter from different scanlator or picture quality so on but it says you did)
3. Yes. Migrate in its current state only saves a few seconds for the user, but is followed by activity taking hours or days (redownload).  Without chapter migration, it's pointless. 

I already acknowledged that it won't work for certain edge cases, but the existing function also doesn't work for the same edge cases.

AntsyLich:
> When I placed a read marker on the chapter, and then migrated the manga, the migrated manga had that chapter as having been read. So what do you mean it doesn't come with the same flaws?
> 
"related to content" I have read chapter 1 is different from I have read chapter from this source. I repeat we don't migrate reading progress or history of a chapter which is related to content. 
> The existing function has the same flaws. (I.e you didn't read this version of the chapter from different scanlator or picture quality so on but it says you did)
> 
Same as above
> Yes. Migrate in its current state only saves a few seconds for the user, but is followed by activity taking hours or days (redownload).  Without chapter file migration, it's pointless. Consider the case where there is 250gb from 1 source.  When the source dies and need to migrate... It takes 8weeks+ to download, and kills sd card if need to migrate multiple times.
> 
>  The problem gets even worse as post migration, can't use download all button as it only downloads unread, so need to click dl button 1 by 1 or clear read marker, download all, then reapply read marker.
> 
>  Look at the alternative (manual) method of migration compared to the current migrate function:
> 
So the thing is migration is useless for you in particular. Most of the user base are not data hoarders so there usually isn't a demand/request for download migration in this scale.
> What alternative solutions? What entry merging? Can you provide link? If there is an alternative solution that will solve my problem I'll be happy to use it.
> 
Quoting myself from the first comment
> Instead we should introduce some kind of merging feature alongside say move downloads to local

Look I'm not saying the problem is non-existent but we should be looking at this from a boarder view. The solution I'm proposing is a very requested feature we could be killing two birds with one stone this way.

AntsyLich:
> > "related to content" I have read chapter 1 is different from I have read chapter from this source. I repeat we don't migrate reading progress or history of a chapter which is related to content.
> 
> ??? As a user, all I see is "did I read this" chapter? yes/no. I don't know what you mean by reading progress or history of a chapter if what you are talking about is different to the read marker that I'm referring to (i.e. hold chapter, click tickbox byitself|with down arrow|with slash). When I migrate, the "did I read this"/"read marker" is set, even if the content is different.
> 
You do realize we have a history tab right?
> > Quoting myself from the first comment
> > > Instead we should introduce some kind of merging feature alongside say move downloads to local
> > 
> > 
> > Look I'm not saying the problem is non-existent but we should be looking at this from a boarder view. The solution I'm proposing is a very requested feature we could be killing two birds with one stone this way.
> 
> So the solution doesn't exist yet? Why not take this one up temporarily, and then delete it once the permanent one is available? Put warnings on the checkbox if you need to, so it's 100% at the user's risk.
>
And who is going to take the maintenance burden of this temporary feature?

fatotak:
> You do realize we have a history tab right?

Yes... what about it?
How is the history tab related to what we were talking about?

> And who is going to take the maintenance burden of this temporary feature?

I'll commit to maintain it and then delete it in 3 months time, giving you guys 3 months to figure out and implement the permanent solution.  Is that an ok resolution?

krysanify:
+1 to have a "merging to local source" instead. there are a few times when I downloaded chapters from one source, but need to download some missing chapters from a different source.

it would be a shame to tied that usefulness to a migration process since i don't use that second source very often.

Jetspectre:
The problem with this approach is that not all sources are the same. Source A could host 20 pages per chapter, but Source B has the SAME CONTENT but cut differently increasing the page numbers to 50 or even 100+. This much is TRUE for every Korean Manhwa that is available (Scanlated vs Official).

Chapter Reader marker migration is a different issue, because the core content of each chapter is consistent across sources. The main difference lies in the inclusion of scanlator pages, credits, or explanatory notes, which vary between sources but don't affect the core chapter content.

While this may work for you, it most certainly will not work for up to 9X% of users.

Some recommendations:
1. Migrate from Source A to Source B -> Delete orphaned downloads -> Download new ones
2. Host a local server with Komga/Mango/Kavita
3. Buy a new phone where you don't have to rely on SD cards, or buy a new OTG thumb drive
4. Make a personal fork with this feature implemented

fatotak:
> The problem with this approach is that not all sources are the same. Source A could host 20 pages per chapter, but Source B has the SAME CONTENT but cut differently increasing the page numbers to 50 or even 100+. This much is TRUE for every Korean Manhwa that is available (Scanlated vs Official).

You think most people actually care that pages are cut differently?
Ridiculous.

> Chapter Reader marker migration is a different issue, because the core content of each chapter is consistent across sources. The main difference lies in the inclusion of scanlator pages, credits, or explanatory notes, which vary between sources but don't affect the core chapter content.

Nope.  The cases where it actually matters, they are both broken.  For all other cases, it's fine.
- Qualitative differences (resolution, translation, etc) - both broken, but suspect minor issue (most people don't care?).
- Censorship - both broken (suspect people care about this one)
- Chapter numbering differences (source A starts at 0, source B starts at 1, source A breaks chapter into 1.1, 1.2, 1.3, source B doesn't break chapter, etc) - both broken.

> While this may work for you, it most certainly will not work for up to 9X% of users.

I think that's a ridiculous assumption.
It's been acknowledged that there certainly are cases where it won't work, but under the SAME cases, the existing chapter read marker also doesn't work.

>     1. Migrate from Source A to Source B -> Delete orphaned downloads -> Download new ones

Sigh... this change is to exactly AVOID this.

>     3. Buy a new phone where you don't have to rely on SD cards, or buy a new OTG thumb drive

And kill your $2000 phone instead of your $100 sd card?
You're such a genius.
Read up on flash wear, and read up on what kind of storage phones, sd cards and thumb drives use (hint: they are ALL flash)

fatotak:
> The more you explained why this feature is needed, the more niche I find the user demographics to be and the furthest away for it to be potentially useful for users that use mihon just to read and not hoarding data.

You are assuming everyone except me uses it like you do.
That's false by default.

> You can say, "so what? it's an opt-in.", but it's not clear why this needs to be in main project of mihon other than your personal needs.

This feature will be:
1. Useful for people like me
    - Don't particularly care about scan quality, translation quality, etc
    - Tendency to read in bulk, rather than one chapter at a time (minimum 5 chapters, preference to read 1 entire arc at a time)
    - Tendency to forget some of the content and read the previous chapters.
    - Tendency to glance at an old chapter when referenced (i.e. chap 452 might have a returning character and a foot note saying "see chap 137", triggering to re-read chap 137 - 142 for a refresher for example).
    - Don't delete chapters until a series is finished and then delete the entire series once the last page of the series has been read.
2. Useless to people who
    - Delete chapters after reading
    - Only read chapters once
    - Do care about differences in translation quality, scan quality, etc

> If I may make a suggestion: find a way for this to work on your fork, share it with the community and see how many opted in. Google Play Store policy requires dev to test their apps on at least 12 beta testers, that could be a good amount to aim even though we're not in the play store.

If I release my fork and take up the maintenance burden, why would I need to come back to mainline?
I'm trying to avoid releasing my own fork as it does come with a maintenance burden (all the other stuff apart from the few features I specifically need added).

I will be happy to support my changes in mainline for a period of time to achieve not having to maintain the rest of the project in my own fork.

taos15:
> The problem with this approach is that not all sources are the same. Source A could host 20 pages per chapter, but Source B has the SAME CONTENT but cut differently increasing the page numbers to 50 or even 100+. This much is TRUE for every Korean Manhwa that is available (Scanlated vs Official).
> 
> Chapter Reader marker migration is a different issue, because the core content of each chapter is consistent across sources. The main difference lies in the inclusion of scanlator pages, credits, or explanatory notes, which vary between sources but don't affect the core chapter content.
> 
> While this may work for you, it most certainly will not work for up to 9X% of users.
> 
> Some recommendations:
> 
>     1. Migrate from Source A to Source B -> Delete orphaned downloads -> Download new ones
> 
>     2. Host a local server with Komga/Mango/Kavita
> 
>     3. Buy a new phone where you don't have to rely on SD cards, or buy a new OTG thumb drive
> 
>     4. Make a personal fork with this feature implemented

One thing people keep forgetting is that most source just re-uploaded (aggregators) chapters from other sources. So the migration will do the same as any source do. For example, there are many cases that the first version of a chapter get updated, if you dont delete and re-download the chapter you would never know.

I think this is a very useful features. It also helps save bandwidth, and lower the chance a source  trying to break working with miho due to abuse. Imagine having a manga that is 3000+ chapters, and the source that you were using is not available anymore, why re-download everything? 

The numbering should not matter, there are evern some source that release .5 version of old chapters and we just mark them as read, so I dont see why should we be able to add any chapter that we already have.

taos15:
> I'm against this whole feature altogether. Downloads are not the same across sources. There are also edge cases here and there. Instead we should introduce some kind of merging feature alongside say move downloads to local

Actually this sounds like a good idea. A local or even a new default library like merged, where you can just `get all that match name:xyz`. That library will get the manga that match that name, and show all your downloaded chapters. This will need a source of true for it to work correctly, something like only showing manga that have a tracker, that way the tracker ID can be use to identify manga that are the same (since the naming, and the id of the source might vary).

fatotak:
> local or even a new default library like merged, where you can just `get all that match name:xyz`.

Although that may sound like a good idea, the problem is that it will be a very big change to how Tachiyomi works, as well as complicated to implement.
1. The base model of Tachiyomi has Source as the top of the food chain.  The change that you guys are proposing fundamentally changes how the model works, with Title being the top of the hierarchy, and sources and chapters being subsidiaries.
2. Then design decisions need to be made about how chapter downloads will work... more specifically, will it download the same chapter from all sources?  Or will it selectively download?  If selectively downloading, how will it choose the source to download from when there are multiple sources available with the same chapter?

I believe this feature, although not useful for every Tachiyomi user, is still useful for many users.
You can talk about ideals, but they don't exist yet, and even if someone starts designing it now, I really don't see it happening for the next 12 months to be honest, given all the regression impact it will have, so I don't see why people are against this, even as a temporary stop-gap measure.

taos15:
> > local or even a new default library like merged, where you can just `get all that match name:xyz`.
> 
> Although that may sound like a good idea, the problem is that it will be a very big change to how Tachiyomi works, as well as complicated to implement.
> 1. The base model of Tachiyomi has Source as the top of the food chain.  The change that you guys are proposing fundamentally changes how the model works, with Title being the top of the hierarchy, and sources and chapters being subsidiaries.
> 2. Then design decisions need to be made about how chapter downloads will work... more specifically, will it download the same chapter from all sources?  Or will it selectively download?  If selectively downloading, how will it choose the source to download from when there are multiple sources available with the same chapter?
> 
> I believe the feature that is implemented in this PR, although not useful for every Tachiyomi user, is still useful for many users.
> I have been unable to use the app in the last few months due to the closure of some sources and the annoyance of migrating titles manually, prompting me to work on this feature.
> 
> You can talk about ideals, **but they don't exist yet**, and even if someone starts designing it now, I really don't see it happening for the next 12 months to be honest, given all the regression impact it will have, so I don't see why people are against this simple solution, even as a temporary stop-gap measure.
> 

I understand you point, I was not suggesting a rewrite of how the current app work, I was think kore like a second `local` which I think there are very few that actually use it.  If am not wrong --I have comtributed to suwoyomibut never mihon, but there are based on the same code based-- the current logic is based on source, but dont we use source's ID? What I was proposing was not changing that logic, but instead reused that implementation to treat trackers ID like a source "only for that new library".

so the tracker will act like a source, instead of getting manga with ID 1234 from source with ID 678; that library will be pulling manga ID 2222 from tracker ID 5555.

 It will not affect any other library, and should onmy display manga that have a tracker. 

I try to comment in discord, but I think I domt have access to the programming channel.

monothread1:
What a completely idiotic discussion.
As a user, I just had a problem with the source code and basically all my downloads broke.
Then I go to github and find a PR ready to be approved with the feature I need and there are simply a bunch of people complaining about completely irrelevant things from a developer who took the time to develop a solution to the problem.
It's no wonder that open source software takes forever to become viable.
Developers with inflated egos, that's the problem.

AntsyLich:
> What a completely idiotic discussion. As a user, I just had a problem with the manga source and basically all my downloads broke. 
>
Well this PR doesn't fix your broken downloads in anyway. The "idiotic" discussion on the other hand provides ideas that can.

fatotak:
> > What a completely idiotic discussion. As a user, I just had a problem with the manga source and basically all my downloads broke.
> 
> Well this PR doesn't fix your broken downloads in anyway. The "idiotic" discussion on the other hand provides ideas that can.

But it does address this exact problem.  It just doesn't do it in a way that some people like (incl you).

AntsyLich:
> > > What a completely idiotic discussion. As a user, I just had a problem with the manga source and basically all my downloads broke.
> > 
> > 
> > Well this PR doesn't fix your broken downloads in anyway. The "idiotic" discussion on the other hand provides ideas that can.
> 
> But it does address this exact problem. It just doesn't do it in a way that some people like (incl you).

Either I'm blind or this PR doesn't fix broken downloads

fatotak:
> > > > What a completely idiotic discussion. As a user, I just had a problem with the manga source and basically all my downloads broke.
> > > 
> > > 
> > > Well this PR doesn't fix your broken downloads in anyway. The "idiotic" discussion on the other hand provides ideas that can.
> > 
> > 
> > But it does address this exact problem. It just doesn't do it in a way that some people like (incl you).
> 
> Either I'm blind or this PR doesn't fix broken downloads

"problem with the manga source" and his "downloads broke"... i.e. migrating to a different source is the solution.  This PR enhances migrate so any downloaded chapters may be retained.  Hence it does fix the issue.

PR #1726, opened 2025-02-14
Fix (PreferenceRestorer): correctly backup/restore preferences related to categories

cuong-tran:
Preferences related to category:
* Library/Default category
* Library/Global update categories
* Downloads/Excluded categoreis
* Downloads/Auto-download categories

They are being stored in preferences as category ID and hence failed to be backed up/restored correctly.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupCategory.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupCategory.kt
@@ -8,6 +8,7 @@ import tachiyomi.domain.category.model.Category
 class BackupCategory(
     @ProtoNumber(1) var name: String,
     @ProtoNumber(2) var order: Long = 0,
+    @ProtoNumber(9) var id: Long = 0,
     // @ProtoNumber(3) val updateInterval: Int = 0, 1.x value not used in 0.x
     @ProtoNumber(100) var flags: Long = 0,
 ) {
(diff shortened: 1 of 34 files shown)


AntsyLich:
Maybe we should also keep the category id when making backup and set this preferences accordingly?

AntsyLich:
> The only downside is need changing backup schema.
Older backups wouldn't have worked correctly anyway

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupCategory.kt
```diff
@@ -8,6 +8,7 @@ import tachiyomi.domain.category.model.Category
 class BackupCategory(
     @ProtoNumber(1) var name: String,
     @ProtoNumber(2) var order: Long = 0,
+    @ProtoNumber(9) var id: Long = 0,
```
AntsyLich, line 11: Why 9?

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/PreferenceRestorer.kt
AntsyLich, line 1: Extract the category preference restore bits in separate function. As is this looks quite messy.

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/backup/models/BackupCategory.kt
AntsyLich: use 3

AntsyLich (approved):

PR #1727, opened 2025-02-14
Fix chapter navigator slider being empty

cuong-tran:
Commit e8c9cb2c2e4c24443368f0d653c5283f9671ffec
forgot to update the reworked slider for chapter navigator slider.


AntsyLich:
it was intentional and i'm keeping it like that for now

PR #1733, opened 2025-02-16
Add an `Add to Favorites` button on HistoryScreen

### LINKED ISSUE #1654: History tab: see if manga is in library

It would be nice if a little icon showed if a manga is added in the library, kinda like how the cover get's a filter when browsing an extension if the manga is already in your library. As most of the manga in the history tab are probably going to be in the users library it would not make sense to use this same filter as it would not be visually appealing imo. If this feature is added it might be interesting to use this same icon when browsing extensions to keep consistency.

### Other details

_No response_

### Acknowledgements

Animeboynz:
closes #1654

Basically that minus the badge ⬇️

https://github.com/user-attachments/assets/d74d75cc-ff05-428a-b31b-6d28ace2569f

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
@@ -38,6 +38,7 @@ fun HistoryScreen(
     onSearchQueryChange: (String?) -> Unit,
     onClickCover: (mangaId: Long) -> Unit,
     onClickResume: (mangaId: Long, chapterId: Long) -> Unit,
+    onClickFavorite: (mangaId: HistoryWithRelations) -> Unit,
     onDialogChange: (HistoryScreenModel.Dialog?) -> Unit,
 ) {
     Scaffold(
@@ -84,6 +85,7 @@ fun HistoryScreen(
                     onClickCover = { history -> onClickCover(history.mangaId) },
                     onClickResume = { history -> onClickResume(history.mangaId, history.chapterId) },
                     onClickDelete = { item -> onDialogChange(HistoryScreenModel.Dialog.Delete(item)) },
+                    onClickFavorite = { history -> onClickFavorite(history) },
                 )
             }
         }
@@ -97,6 +99,7 @@ private fun HistoryScreenContent(
     onClickCover: (HistoryWithRelations) -> Unit,
     onClickResume: (HistoryWithRelations) -> Unit,
     onClickDelete: (HistoryWithRelations) -> Unit,
+    onClickFavorite: (HistoryWithRelations) -> Unit,
 ) {
     FastScrollLazyColumn(
         contentPadding = contentPadding,
@@ -126,6 +129,7 @@ private fun HistoryScreenContent(
                         onClickCover = { onClickCover(value) },
                         onClickResume = { onClickResume(value) },
                         onClickDelete = { onClickDelete(value) },
+                        onClickFavorite = { onClickFavorite(value) },
                     )
                 }
             }
@@ -152,6 +156,7 @@ internal fun HistoryScreenPreviews(
             onClickCover = {},
             onClickResume = { _, _ -> run {} },
             onDialogChange = {},
+            onClickFavorite = {},
         )
     }
 }
--- a/app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
@@ -38,7 +38,7 @@ fun HistoryScreen(
     onSearchQueryChange: (String?) -> Unit,
     onClickCover: (mangaId: Long) -> Unit,
     onClickResume: (mangaId: Long, chapterId: Long) -> Unit,
-    onClickFavorite: (mangaId: HistoryWithRelations) -> Unit,
+    onClickFavorite: (mangaId: Long) -> Unit,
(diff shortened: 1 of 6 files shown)


app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
```diff
@@ -38,6 +38,7 @@ fun HistoryScreen(
     onSearchQueryChange: (String?) -> Unit,
     onClickCover: (mangaId: Long) -> Unit,
     onClickResume: (mangaId: Long, chapterId: Long) -> Unit,
+    onClickFavorite: (mangaId: HistoryWithRelations) -> Unit,
```
AntsyLich, line 41: This doesn't look like a 'mangaId'

app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryScreenModel.kt
```diff
@@ -112,6 +137,110 @@ class HistoryScreenModel(
...
+            val selection = getMangaCategoryIds(manga)
+            mutableState.update { currentState ->
+                currentState.copy(
+                    dialog = Dialog.ChangeCategory(
+                        manga = manga,
+                        initialSelection = categories.mapAsCheckboxState { it.id in selection }.toImmutableList(),
+                    ),
+                )
+            }
+        }
```
AntsyLich, line 241: Why will there be a selected category for non-library stuff?

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryScreenModel.kt
cuong-tran: This probably inherited from existed code, non-favorited entry might already in library previously and already got its categories.

Another case is when accidently removes an entry from library and then users go to history to find and add it back, would need to keep its categories.

Animeboynz:
Ah shit need to change the screenModel too. Gonna force push 😭

app/src/main/java/eu/kanade/presentation/history/HistoryScreen.kt
Animeboynz: Fixed now

app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryScreenModel.kt
Animeboynz: Yeah. It's there cus I copied the code from another screen, thought of removing it but I saw no harm in leaving it as someone might be trying to re-add an entry they had removed previously

AntsyLich: Yeah but we don't store the info tho

Animeboynz: https://github.com/user-attachments/assets/8d8817e1-ee59-40f4-a09a-eb71a9efe946

Animeboynz: It seems to prefill with the last categories

AntsyLich: 🤔

Animeboynz: Ignore how I'm using my PC lol, I'm on a bus rn

cuong-tran: because when remove an entry from library, we simply set favorite to false, not actually remove it from DB so all info, chapters, categories, read histories is still there

AntsyLich:
Dis GitHub just eat my 3rd review 🤔

Animeboynz:
Probably cus there is only 2 reviews. What was it for?

app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryScreenModel.kt
```diff
@@ -112,6 +137,110 @@ class HistoryScreenModel(
...
+
+            if (checkDuplicate) {
+                val duplicate = getDuplicateLibraryManga.await(manga).getOrNull(0)
+
+                if (duplicate != null) {
+                    mutableState.update { it.copy(dialog = Dialog.DuplicateManga(manga, duplicate)) }
+                    return@launchIO
+                }
+            }
+
```
AntsyLich, line 193:
```suggestion
    fun addFavorite(id: Long) {
        screenModelScope.launchIO {
            val manga = getManga.await(mangaOrId) ?: return@launchIO
            
            val duplicate = getDuplicateLibraryManga.await(manga).getOrNull(0)
            if (duplicate != null) {
                mutableState.update { it.copy(dialog = Dialog.DuplicateManga(manga, duplicate)) }
                return@launchIO
            }
            addFavorite(manga)
        }
    }
    
    fun addFavorite(manga: Manga) {
        screenModelScope.launchIO {
```

AntsyLich (changes requested):

AntsyLich (approved):
Changelog

Animeboynz:
Should be all done here, hopefully

PR #1734, opened 2025-02-16
remove reference to Preferences of deleting category

cuong-tran:
When a category is deleted, preferences which are referencing to that category should also be clear to avoid null/incorrect settings

### DIFF
--- a/app/src/main/java/mihon/core/migration/migrations/CategoryPreferencesCleanupMigration.kt
+++ b/app/src/main/java/mihon/core/migration/migrations/CategoryPreferencesCleanupMigration.kt
@@ -6,8 +6,6 @@ import tachiyomi.core.common.util.lang.withIOContext
 import tachiyomi.domain.category.interactor.GetCategories
 import tachiyomi.domain.download.service.DownloadPreferences
 import tachiyomi.domain.library.service.LibraryPreferences
-import uy.kohesive.injekt.Injekt
-import uy.kohesive.injekt.api.get
 
 class CategoryPreferencesCleanupMigration : Migration {
     override val version: Float = 10f
@@ -16,7 +14,7 @@ class CategoryPreferencesCleanupMigration : Migration {
         val libraryPreferences = migrationContext.get<LibraryPreferences>() ?: return@withIOContext false
         val downloadPreferences = migrationContext.get<DownloadPreferences>() ?: return@withIOContext false
 
-        val getCategories = Injekt.get<GetCategories>()
+        val getCategories = migrationContext.get<GetCategories>() ?: return@withIOContext false
         val allCategories = getCategories.await().map { it.id.toString() }.toSet()
 
         val defaultCategory = libraryPreferences.defaultCategory().get()
--- a/app/src/main/java/mihon/core/migration/migrations/CategoryPreferencesCleanupMigration.kt
+++ b/app/src/main/java/mihon/core/migration/migrations/CategoryPreferencesCleanupMigration.kt
@@ -3,43 +3,39 @@ package mihon.core.migration.migrations
 import mihon.core.migration.Migration
 import mihon.core.migration.MigrationContext
 import tachiyomi.core.common.util.lang.withIOContext
-import tachiyomi.data.DatabaseHandler
+import tachiyomi.domain.category.interactor.GetCategories
 import tachiyomi.domain.download.service.DownloadPreferences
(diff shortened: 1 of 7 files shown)


AntsyLich (changes requested):
Inject in the class constructor also add a migration to clean existing preferences? And changelog.

cuong-tran:
> add a migration to clean existing preferences?

A migration seems overkill. These preferences can be easily fixed once user notices and re-set them again.

MajorTanya:
> > add a migration to clean existing preferences?
> 
> A migration seems overkill. These preferences can be easily fixed once user notices and re-set them again.
> 
> 

Never rely on user action to fix incorrect state.

BrutuZ:
> once user notices 

Overestimating much? :risitas:

cuong-tran:
because this referred settings issue been there since long but users didn't really have problem with it that much, meaning this issue is not a critical one.
Migration, in other hand, requires bumping versionCode and hence prevent user from going back version.

MajorTanya:
> because this referred settings issue been there since long but users didn't really have problem with it that much, meaning this issue is not a critical one.
> Migration, in other hand, requires bumping versionCode and hence prevent user from going back version.

How does it prevent uninstalling and installing an old version? It doesn't. And going back to older versions isn't a supported thing for Mihon either. We make it a point in support requests and issues here to make sure people use the latest version before raising complaints.

cuong-tran:
> > because this referred settings issue been there since long but users didn't really have problem with it that much, meaning this issue is not a critical one.
> > Migration, in other hand, requires bumping versionCode and hence prevent user from going back version.
> 
> How does it prevent uninstalling and installing an old version? It doesn't. And going back to older versions isn't a supported thing for Mihon either. We make it a point in support requests and issues here to make sure people use the latest version before raising complaints.

it prevents install over (to keep data/cache). Anw, it's opinion because I only said overkill, given the minor affect of this issue comparing to bumping versionCode.

cuong-tran:
> You do realize I'm going bump the version on new release anyway?

as a matter of fact, no I don't, because there are so little updates so I honestly couldn't notice. anw I'm doing migration now.

AntsyLich:
> > You do realize I'm going bump the version on new release anyway?
> 
> as a matter of fact, no I don't, because there are so little updates so I honestly couldn't notice. anw I'm doing migration now.

Oi that's a foul :angy:

AntsyLich (changes requested):
Also bump app version to 10

domain/src/main/java/tachiyomi/domain/category/interactor/DeleteCategory.kt
```diff
@@ -5,9 +5,15 @@ import tachiyomi.core.common.util.lang.withNonCancellableContext
...
 import tachiyomi.domain.category.repository.CategoryRepository
+import tachiyomi.domain.download.service.DownloadPreferences
+import tachiyomi.domain.library.service.LibraryPreferences
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
 
 class DeleteCategory(
     private val categoryRepository: CategoryRepository,
+    private val libraryPreferences: LibraryPreferences = Injekt.get(),
+    private val downloadPreferences: DownloadPreferences = Injekt.get(),
```
AntsyLich, line 16:
```suggestion
    private val libraryPreferences: LibraryPreferences,
    private val downloadPreferences: DownloadPreferences,
```

app/src/main/java/mihon/core/migration/migrations/CategoryPreferencesCleanupMigration.kt
```diff
@@ -0,0 +1,42 @@
...
+import uy.kohesive.injekt.api.get
+
+class CategoryPreferencesCleanupMigration : Migration {
+    override val version: Float = 10f
+
+    override suspend fun invoke(migrationContext: MigrationContext): Boolean = withIOContext {
+        val libraryPreferences = migrationContext.get<LibraryPreferences>() ?: return@withIOContext false
+        val downloadPreferences = migrationContext.get<DownloadPreferences>() ?: return@withIOContext false
+
+        val getCategories = Injekt.get<GetCategories>()
```
AntsyLich, line 19:
```suggestion
        val getCategories = migrationContext.get<GetCategories>()
```

AntsyLich (changes requested):

Issue #1735, opened 2025-02-16
Track entries privately if supported by the tracker


NarwhalHorns:
### Describe your suggested feature

Allow user to chose to track entries privately or not on adding and toggle afterwards, if supported by a tracker. Could hide functionality behind a settings toggle to not interfere with most users who are not bothered by it.

### Other details

Mostly describes itself. Don't know about other trackers but I use Anilist and that has private tracking capablities. Currently Mihon disregards private so defaults to public which means you have to then change to private manually using the site or something else.

~Already done some PoC on my own and seems mostly easy to implement. Am a beginner tho so can only hope my work is quality enough or makes sense. Am having an issue with findLibManga in AnilistApi not returning private entries despite being authenticated tho.~ (thanks to @MajorTanya )

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.17.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

MajorTanya:
> Am having an issue with findLibManga in AnilistApi not returning private entries despite being authenticated tho.

Have you checked whether the query Mihon sends to AL could fetch this data? I haven't used this private tracking you mentioned but it may require you to adjust the query Mihon sends over so that AL includes the private entries. No idea if that's the case, this is just me spitballing.

krysanify:
> Currently Mihon disregards private so defaults to public which means you have to then change to private manually using the site or **other service**.

This other service you mentioned, are they open-source? maybe we can learn how are they able to manage private entries.

Also worth checking if there's some sort of consent required from user to tell Anilist that yes, they may share private entries with third-party apps.

NarwhalHorns:
> This other service you mentioned, are they open-source? maybe we can learn how are they able to manage private entries.

oh I didn't have one in mind, my bad. Just meant you had to leave Mihon and find 'some other' thing to do it, if not the site.

krysanify:
No worries, that's just mean we're back at MajorTanya's question: does Anilist even allow third-party apps to gain access to private entries?

NarwhalHorns:
I've been able to add, edit and remove them so far so I imagine so, I'm just having some trouble with that query.

PR #1736, opened 2025-02-16
Track entries privately if supported by the tracker

NarwhalHorns:
Features in #1735

- Add a private tracking setting toggle to allow private tracking.
- Add `supportsPrivateTracking` to Trackers
- if tracking is enabled and supported, can choose between private or public tracking when adding a track on the search screen.
- Add a toggle to the track info screen to toggle between private and public tracking on entries

Current issues:
- `findLibManga` query in anilistApi doesn't seem to return private entries causing further issues down the line

### DIFF
--- a/app/src/main/java/eu/kanade/domain/track/service/TrackPreferences.kt
+++ b/app/src/main/java/eu/kanade/domain/track/service/TrackPreferences.kt
@@ -36,6 +36,8 @@ class TrackPreferences(
 
     fun anilistScoreType() = preferenceStore.getString("anilist_score_type", Anilist.POINT_10)
 
+    fun privateTracking() = preferenceStore.getBoolean("pref_private_tracking", false)
+
     fun autoUpdateTrack() = preferenceStore.getBoolean("pref_auto_update_manga_sync_key", true)
 
     fun autoUpdateTrackOnMarkRead() = preferenceStore.getEnum(
--- a/app/src/main/java/eu/kanade/presentation/track/TrackerSearch.kt
+++ b/app/src/main/java/eu/kanade/presentation/track/TrackerSearch.kt
@@ -11,6 +11,7 @@ import androidx.compose.foundation.combinedClickable
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.FlowRow
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
@@ -90,8 +91,9 @@ fun TrackerSearch(
     queryResult: Result<List<TrackSearch>>?,
     selected: TrackSearch?,
     onSelectedChange: (TrackSearch) -> Unit,
-    onConfirmSelection: () -> Unit,
+    onConfirmSelection: (private: Boolean) -> Unit,
     onDismissRequest: () -> Unit,
+    privateTracking: Boolean,
 ) {
     val focusManager = LocalFocusManager.current
     val focusRequester = remember { FocusRequester() }
@@ -164,15 +166,42 @@ fun TrackerSearch(
                 enter = fadeIn() + slideInVertically { it / 2 },
                 exit = slideOutVertically { it / 2 } + fadeOut(),
             ) {
-                Button(
-                    onClick = { onConfirmSelection() },
-                    modifier = Modifier
-                        .padding(12.dp)
-                        .windowInsetsPadding(WindowInsets.navigationBars)
-                        .fillMaxWidth(),
-                    elevation = ButtonDefaults.elevatedButtonElevation(),
-                ) {
-                    Text(text = stringResource(MR.strings.action_track))
+                if (privateTracking) {
+                    FlowRow (
+                        modifier = Modifier.fillMaxWidth()
+                    ) {
+                        Button(
+                            onClick = { onConfirmSelection(false) },
+                            modifier = Modifier
+                                .weight(1f)
+                                .padding(12.dp)
+                                .windowInsetsPadding(WindowInsets.navigationBars),
+                            elevation = ButtonDefaults.elevatedButtonElevation(),
+                        ) {
+                            Text(text = stringResource(MR.strings.action_track))
+                        }
+                        Button(
+                            onClick = { onConfirmSelection(true) },
+                            modifier = Modifier
+                                .weight(1f)
+                                .padding(12.dp)
+                                .windowInsetsPadding(WindowInsets.navigationBars),
+                            elevation = ButtonDefaults.elevatedButtonElevation(),
+                        ) {
+                            Text(text = stringResource(MR.strings.action_private_track))
+                        }
+                    }
+                } else {
+                    Button(
+                        onClick = { onConfirmSelection(false) },
+                        modifier = Modifier
+                            .padding(12.dp)
+                            .windowInsetsPadding(WindowInsets.navigationBars)
+                            .fillMaxWidth(),
+                        elevation = ButtonDefaults.elevatedButtonElevation(),
+                    ) {
+                        Text(text = stringResource(MR.strings.action_track))
+                    }
                 }
             }
         },
--- a/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
@@ -32,12 +32,15 @@ interface Track : Serializable {
 
     var tracking_url: String
 
+    var private: Boolean
+
     fun copyPersonalFrom(other: Track) {
         last_chapter_read = other.last_chapter_read
         score = other.score
         status = other.status
         started_reading_date = other.started_reading_date
         finished_reading_date = other.finished_reading_date
+        private = other.private
     }
 
     companion object {
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/Tracker.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/Tracker.kt
@@ -22,6 +22,8 @@ interface Tracker {
     // Application and remote support for reading dates
     val supportsReadingDates: Boolean
 
+    val supportsPrivateTracking: Boolean
+
     @ColorInt
     fun getLogoColor(): Int
 
@@ -82,4 +84,6 @@ interface Tracker {
     suspend fun setRemoteStartDate(track: Track, epochMillis: Long)
 
     suspend fun setRemoteFinishDate(track: Track, epochMillis: Long)
+
+    suspend fun setRemotePrivate(track: Track, private: Boolean)
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
@@ -43,6 +43,8 @@ class Anilist(id: Long) : BaseTracker(id, "AniList"), DeletableTracker {
 
     override val supportsReadingDates: Boolean = true
 
+    override val supportsPrivateTracking: Boolean = true
+
     private val scorePreference = trackPreferences.anilistScoreType()
 
     init {
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/dto/ALManga.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/dto/ALManga.kt
@@ -49,6 +49,7 @@ data class ALUserManga(
     val startDateFuzzy: Long,
     val completedDateFuzzy: Long,
     val manga: ALManga,
+    val private: Boolean,
 ) {
     fun toTrack() = Track.create(TrackerManager.ANILIST).apply {
         remote_id = manga.remoteId
@@ -60,6 +61,7 @@ data class ALUserManga(
         last_chapter_read = chaptersRead.toDouble()
         library_id = libraryId
         total_chapters = manga.totalChapters
+        private = private
     }
 
     private fun toTrackStatus() = when (listStatus) {
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackInfoDialog.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackInfoDialog.kt
@@ -40,10 +40,12 @@ import cafe.adriel.voyager.navigator.currentOrThrow
 import dev.icerock.moko.resources.StringResource
 import eu.kanade.domain.track.interactor.RefreshTracks
 import eu.kanade.domain.track.model.toDbTrack
+import eu.kanade.domain.track.service.TrackPreferences
 import eu.kanade.domain.ui.UiPreferences
 import eu.kanade.presentation.track.TrackChapterSelector
 import eu.kanade.presentation.track.TrackDateSelector
 import eu.kanade.presentation.track.TrackInfoDialogHome
+import eu.kanade.presentation.track.TrackPrivateSelector
 import eu.kanade.presentation.track.TrackScoreSelector
 import eu.kanade.presentation.track.TrackStatusSelector
 import eu.kanade.presentation.track.TrackerSearch
@@ -58,6 +60,7 @@ import eu.kanade.tachiyomi.util.system.copyToClipboard
 import eu.kanade.tachiyomi.util.system.openInBrowser
 import eu.kanade.tachiyomi.util.system.toast
 import kotlinx.collections.immutable.ImmutableList
+import kotlinx.collections.immutable.persistentListOf
 import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.distinctUntilChanged
@@ -82,6 +85,7 @@ import tachiyomi.presentation.core.components.material.padding
 import tachiyomi.presentation.core.i18n.stringResource
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
+import uy.kohesive.injekt.injectLazy
 import java.time.Instant
 import java.time.LocalDate
 import java.time.ZoneId
@@ -144,6 +148,14 @@ data class TrackInfoDialogHomeScreen(
                         track = it.track!!,
                         serviceId = it.tracker.id,
                         start = false,
+                    ),
+                )
+            },
+            onPrivateClick = {
+                navigator.push(
+                    TrackPrivateSelectorScreen(
+                        track = it.track!!,
+                        serviceId = it.tracker.id,
                     ),
                 )
             },
@@ -645,6 +657,58 @@ private data class TrackDateRemoverScreen(
     }
 }
 
+private data class TrackPrivateSelectorScreen(
+    private val track: Track,
+    private val serviceId: Long,
+) : Screen() {
+
+    @Composable
+    override fun Content() {
+        val navigator = LocalNavigator.currentOrThrow
+        val screenModel = rememberScreenModel {
+            Model(
+                track = track,
+                tracker = Injekt.get<TrackerManager>().get(serviceId)!!,
+            )
+        }
+        val state by screenModel.state.collectAsState()
+
+        TrackPrivateSelector(
+            selection = state.selection,
+            selections = persistentListOf(
+                stringResource(MR.strings.track_public),
+                stringResource(MR.strings.track_private)),
+            onSelectionChange = screenModel::setSelection,
+            onConfirm = {
+                screenModel.setPrivate()
+                navigator.pop()
+            },
+            onDismissRequest = navigator::pop,
+        )
+    }
+
+    private class Model(
+        private val track: Track,
+        private val tracker: Tracker,
+    ) : StateScreenModel<Model.State>(State(track.private)) {
+
+        fun setSelection(selection: Boolean) {
+            mutableState.update { it.copy(selection = selection) }
+        }
+
+        fun setPrivate() {
+            screenModelScope.launchNonCancellable {
+                tracker.setRemotePrivate(track.toDbTrack(), state.value.selection)
+            }
+        }
+
+        @Immutable
+        data class State(
+            val selection: Boolean,
+        )
+    }
+}
+
 data class TrackerSearchScreen(
     private val mangaId: Long,
     private val initialQuery: String,
@@ -654,6 +718,8 @@ data class TrackerSearchScreen(
 
     @Composable
     override fun Content() {
+        val trackPreferences: TrackPreferences by injectLazy()
+
         val navigator = LocalNavigator.currentOrThrow
         val screenModel = rememberScreenModel {
             Model(
@@ -673,11 +739,13 @@ data class TrackerSearchScreen(
             queryResult = state.queryResult,
             selected = state.selected,
             onSelectedChange = screenModel::updateSelection,
-            onConfirmSelection = {
+            onConfirmSelection = { private: Boolean ->
+                state.selected!!.private = private
                 screenModel.registerTracking(state.selected!!)
                 navigator.pop()
             },
             onDismissRequest = navigator::pop,
+            privateTracking = screenModel.privateTracking && trackPreferences.privateTracking().get(),
         )
     }
 
@@ -724,6 +792,8 @@ data class TrackerSearchScreen(
         fun updateSelection(selected: TrackSearch) {
             mutableState.update { it.copy(selected = selected) }
         }
+
+        val privateTracking = tracker.supportsPrivateTracking
 
         @Immutable
         data class State(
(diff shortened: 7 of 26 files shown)


NarwhalHorns:
Good luck, I'll keep digging too

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/dto/ALManga.kt
```diff
@@ -60,6 +61,7 @@ data class ALUserManga(
         last_chapter_read = chaptersRead.toDouble()
         library_id = libraryId
         total_chapters = manga.totalChapters
+        private = private
```
MajorTanya, line 64:
```suggestion
        private = this@ALUserManga.private
```
Found it. Otherwise, remoteManga will have false here, and we update our local one with the data from remote.

NarwhalHorns: ahhhh damn that was sneaky, nice job

app/src/main/java/eu/kanade/presentation/track/TrackerSearch.kt
```diff
@@ -164,15 +166,42 @@ fun TrackerSearch(
...
+                            .padding(12.dp)
+                            .windowInsetsPadding(WindowInsets.navigationBars)
+                            .fillMaxWidth(),
+                        elevation = ButtonDefaults.elevatedButtonElevation(),
+                    ) {
+                        Text(text = stringResource(MR.strings.action_track))
+                    }
                 }
             }
         },
```
MajorTanya, line 207:
```suggestion
                Row (
                    modifier = Modifier.fillMaxWidth()
                ) {
                    Button(
                        onClick = { onConfirmSelection(false) },
                        modifier = Modifier
                            .weight(1f)
                            .padding(12.dp)
                            .windowInsetsPadding(WindowInsets.navigationBars),
                        elevation = ButtonDefaults.elevatedButtonElevation(),
                    ) {
                        Text(text = stringResource(MR.strings.action_track))
                    }
                    if (privateTracking) {
                        Button(
                            onClick = { onConfirmSelection(true) },
                            modifier = Modifier
                                .weight(1f)
                                .padding(12.dp)
                                .windowInsetsPadding(WindowInsets.navigationBars),
                            elevation = ButtonDefaults.elevatedButtonElevation(),
                        ) {
                            Text(text = stringResource(MR.strings.action_private_track))
                        }
                    }
                }
            }
        },
```

No need for FlowRow or just Button. You can make this a Row with the normal button in it, then surround the private tracking button with the if so it only shows when it should. You can keep the weight on the standard button too.

For this suggestion, I'd _suggest_ looking at the diff in AS first. I'm confident I caught all the braces correctly but it would be annoying to fix it if I didn't and you just applied here.

MajorTanya: Yeah this would miss the import for FlowRow...

NarwhalHorns: oh coolio. yeah had a weird time trying to make that work. if it isn't clear this is my first endevour at anything kotlin so i appreciate any suggestions and whatnot

NarwhalHorns: indeed, done

MajorTanya:
The Row should look pretty much identical but now it's the normal Row scope used in other places. FlowRow, according to its docs, would also start overflowing if the content didn't fit and make another row (not desirable here ~right now~).

Haven't taken a super detailed look at everything but considering it worked perfectly fine once we found that sneaky little thing, it seems to be good work.

Specifically, I checked an entry that I had already added as private through the AL website and then connected that to Mihon, which got kept correctly now. (Mihon show "Private" and AL still has the checkbox ticked).
Also added an entirely new entry from within Mihon that didn't exist on my AL yet, and that also worked correctly when I selected the private tracking option. AL showed the checkbox ticked and Mihon showed "Private".

Others may have other suggestions but the feature itself appears to be working. Nice job! 👍

AntsyLich:
I don't quite like the UI. Are you able to provide any alternative designs?

NarwhalHorns:
> I don't quite like the UI. Are you able to provide any alternative designs?

Mmm, I wasn't really sure what to go with and just winged what came to me first. what aspects of it don't you like or any suggestions?

AntsyLich:
In the tracker sheet the private toggle has more visual space given it's likely a one time toggle. And there being 2 type of track button in search

NarwhalHorns:
yeah, on the 2 buttons I'm not sure what would be good. Like we could have the track button as is but have a private toggle checkbox or something on there somewhere, but it's just adding more steps. whereas 2 buttons keep both to a one step interaction.

on the setting toggle, not sure what you mean. I could move it to the bottom of the screen? or perhaps put it in advanced settings and just stick it in a tracking section?

Animeboynz:
Just my personal opinion but the selector feels overkill for just 2 options. 
I feel like it would probably be cleaner to just switch it on click
Screenshot_20250217_203250_Chrome

Animeboynz:
> In the tracker sheet the private toggle has more visual space given it's likely a one time toggle.

I guess it can also be moved to the overflow/kebab menu thing right? And show like the incognito icon as a badge on the tracker logo or next to the name

NarwhalHorns:
ohhhh i see what you mean now right. yeah good idea

MajorTanya:
> _This will include only the tracker abstraction and AniList implementation_

I've checked the other non-enhanced trackers:
Has no support:
- MAL
- Shikimori

Has support:
- Kitsu
- Bangumi

MangaUpdates only has private/public lists, not a per-title way to set things to private.

MajorTanya:
Checking [Kitsu Docs](https://kitsu.docs.apiary.io/#reference/user-libraries/library-entries/update-resource), the change for it could be extremely simple, just adding the `private: true` attribute to the request we send to them.

In KitsuApi.kt, in updateLibManga and addLibManga, you should be able to just add the private attribute to the JSON object being built.

MajorTanya:
And Bangumi is looking similarly simple. But their docs are not only incomplete and in Chinese, they also contradict themselves and are out of date in places, so no links there. Aside from DTO changes, just adding the `private` attribute to the add and update lib manga method "should be it".

NarwhalHorns:
> I guess it can also be moved to the overflow/kebab menu thing right? And show like the incognito icon as a badge on the tracker logo or next to the name
How's this?

| Privately tracked | Publicly tracked |
|----|-----|
| image | image |

NarwhalHorns:
if we can think of way to make the search private as unintrusive as this we probably don't need the full setting toggle tbh but i feel like that will be a harder task

NarwhalHorns:
how about this for the search?

image

NarwhalHorns:
tho i'm not sure the glasses get the point across that well, perhaps a like padlock?

Animeboynz:
I'm not sure how I feel about the glasses on the left of the title.
If the user has multiple trackers the titles will be misaligned cus of that

Also don't forget to create migrations 
image

MajorTanya:
Is there no Icognito mode icon in the Icons lib we use?

Animeboynz:
> Is there no Incognito mode icon in the Icons lib we use?

I don't think there is xD

Also I would recommend something like this, you can use whatever icon tho like lock or something
Screenshot_20250217_225217_Mihon

NarwhalHorns:
yeah this is nice

image

NarwhalHorns:
kinda winged adding a migration, let me know if it's wrong lol

MajorTanya:
I don't think you need to specify AniList as a separate entry in the Changelog. Not sure on the exact wording but what about "Support for private tracking with AniList"? Readers of the Changelog don't need to know that you made a baseline as well.

NarwhalHorns:
> I don't think you need to specify AniList as a separate entry in the Changelog. Not sure on the exact wording but what about "Support for private tracking with AniList"? Readers of the Changelog don't need to know that you made a baseline as well.

oh i just thought that would be good to specify cus that would then make it clear that others can be added but would need work. But yeah i suppose if the changelog is for users primarily that makes sense. Will do

NarwhalHorns:
Well i believe it is functionally done, anyone got any more gripes with the UI we can iron out?

MajorTanya:
> > I don't think you need to specify AniList as a separate entry in the Changelog. Not sure on the exact wording but what about "Support for private tracking with AniList"? Readers of the Changelog don't need to know that you made a baseline as well.
> 
> oh i just thought that would be good to specify cus that would then make it clear that others can be added but would need work. But yeah i suppose if the changelog is for users primarily that makes sense. Will do

It would make it clear, yeah, but we'd also get "when private tracking for mal???" requests every other day or something 😅

NarwhalHorns:
> Edit: Btw, I'm working on a Bangumi rework and it just so happens that private tracking will probably become trivial to add for it once I'm done. It'll probably just come down to enabling it in the Bangumi tracker class. But no promises yet. Also, don't consider that a blocker if you don't want to wait/add Bangumi and/or Kitsu. Would be cool if all trackers got it at the same time though, so just tell me if you don't plan on it and I'll prepare a follow up PR for those two.

oh sorry i missed this, idk how this works tbh, if you can just put the changes in here go for it lol or if they are truly trivial i can do it. but if its a bit of work for you might as well keep in seperate for now i guess?

MajorTanya:
> > Edit: Btw, I'm working on a Bangumi rework and it just so happens that private tracking will probably become trivial to add for it once I'm done. It'll probably just come down to enabling it in the Bangumi tracker class. But no promises yet. Also, don't consider that a blocker if you don't want to wait/add Bangumi and/or Kitsu. Would be cool if all trackers got it at the same time though, so just tell me if you don't plan on it and I'll prepare a follow up PR for those two.
> 
> oh sorry i missed this, idk how this works tbh, if you can just put the changes in here go for it lol or if they are truly trivial i can do it. but if its a bit of work for you might as well keep in seperate for now i guess?

I'll do a follow-up for Bangumi and Kitsu later then. I'll have the luxury of just slotting in to your system.

AntsyLich:
Get the icon from `Icons` class instead of adding a drawable

NarwhalHorns:
> Get the icon from `Icons` class instead of adding a drawable

I had a feelin there was a better way of doing that lol

CHANGELOG.md
AntsyLich, line 1: Please revert the formatting changes. Make a separate PR if you want

app/src/main/java/eu/kanade/domain/track/service/TrackPreferences.kt
```diff
@@ -36,6 +36,8 @@ class TrackPreferences(
 
     fun anilistScoreType() = preferenceStore.getString("anilist_score_type", Anilist.POINT_10)
 
+    fun privateTracking() = preferenceStore.getBoolean("pref_private_tracking", false)
```
AntsyLich, line 39: It's already opt-in we don't another extra toggle for this

app/src/main/java/eu/kanade/tachiyomi/ui/manga/track/TrackInfoDialog.kt
```diff
@@ -673,11 +684,13 @@ data class TrackerSearchScreen(
             queryResult = state.queryResult,
             selected = state.selected,
             onSelectedChange = screenModel::updateSelection,
-            onConfirmSelection = {
+            onConfirmSelection = { private: Boolean ->
+                state.selected!!.private = private
                 screenModel.registerTracking(state.selected!!)
```
AntsyLich, line 689:
```suggestion
            onConfirmSelection = f@{ private: Boolean ->
                val selected = state.selected ?: return@f
                selected.private = private
                screenModel.registerTracking(selected)
```

i18n/src/commonMain/moko-resources/base/strings.xml
```diff
@@ -751,6 +751,8 @@
     <!-- Tracking Screen -->
     <string name="manga_tracking_tab">Tracking</string>
     <string name="add_tracking">Add tracking</string>
+    <string name="action_track">Track</string>
+    <string name="action_private_track">Track privately</string>
```
AntsyLich, line 755: Use the other one

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
```diff
@@ -183,6 +184,10 @@ class Anilist(id: Long) : BaseTracker(id, "AniList"), DeletableTracker {
     override suspend fun bind(track: Track, hasReadChapters: Boolean): Track {
         val remoteTrack = api.findLibManga(track, getUsername().toInt())
         return if (remoteTrack != null) {
+            if (this.supportsPrivateTracking && trackPreferences.privateTracking().get()) {
```
AntsyLich, line 187: I don't see the point for doing a `this.supportsPrivateTracking` check here

app/src/main/java/eu/kanade/tachiyomi/data/track/Tracker.kt
```diff
@@ -82,4 +84,8 @@ interface Tracker {
     suspend fun setRemoteStartDate(track: Track, epochMillis: Long)
 
     suspend fun setRemoteFinishDate(track: Track, epochMillis: Long)
+
+    suspend fun setRemotePrivate(track: Track, private: Boolean)
+
+    suspend fun togglePrivate(track: Track) = this.setRemotePrivate(track, !track.private)
```
AntsyLich, line 90: toggle only happens in one single place so no need to make it a function of the tracker

AntsyLich (changes requested):

i18n/src/commonMain/moko-resources/base/strings.xml
AntsyLich: `action_toggle_private_on`

app/src/main/java/eu/kanade/domain/track/service/TrackPreferences.kt
AntsyLich: Users can already choose if they want to track privately or not. No need for an additional toggle on top of that

app/src/main/java/eu/kanade/presentation/track/TrackerSearch.kt
```diff
@@ -90,8 +91,9 @@ fun TrackerSearch(
     queryResult: Result<List<TrackSearch>>?,
     selected: TrackSearch?,
     onSelectedChange: (TrackSearch) -> Unit,
-    onConfirmSelection: () -> Unit,
+    onConfirmSelection: (private: Boolean) -> Unit,
     onDismissRequest: () -> Unit,
+    privateTracking: Boolean,
```
AntsyLich, line 96:
```suggestion
    supportsPrivateTracking: Boolean,
```

app/src/main/java/eu/kanade/presentation/track/TrackerSearch.kt
```diff
@@ -164,15 +166,33 @@ fun TrackerSearch(
...
             ) {
-                Button(
-                    onClick = { onConfirmSelection() },
-                    modifier = Modifier
-                        .padding(12.dp)
-                        .windowInsetsPadding(WindowInsets.navigationBars)
-                        .fillMaxWidth(),
-                    elevation = ButtonDefaults.elevatedButtonElevation(),
+                Row(
+                    modifier = Modifier.fillMaxWidth(),
```
AntsyLich, line 170:
```suggestion
                    modifier = Modifier
                        .padding(MaterialTheme.padding.small)
                        .windowInsetsPadding(WindowInsets.navigationBars)
                        .fillMaxWidth(),
                    horizontalArrangement = Arrangement.spacedBy(MaterialTheme.padding.small),
```

app/src/main/java/eu/kanade/presentation/track/TrackerSearch.kt
```diff
@@ -164,15 +166,33 @@ fun TrackerSearch(
...
+                                .windowInsetsPadding(WindowInsets.navigationBars),
+                            elevation = ButtonDefaults.elevatedButtonElevation(),
+                        ) {
+                            Icon(
+                                imageVector = Icons.Filled.VisibilityOff,
+                                contentDescription = stringResource(MR.strings.action_toggle_private_on),
+                            )
+                        }
+                    }
                 }
```
AntsyLich, line 196:
```suggestion
                    Button(
                        onClick = { onConfirmSelection(false) },
                        modifier = Modifier.weight(1f),
                        elevation = ButtonDefaults.elevatedButtonElevation(),
                    ) {
                        Text(text = stringResource(MR.strings.action_track))
                    }
                    if (privateTracking) {
                        Button(
                            onClick = { onConfirmSelection(true) },
                            elevation = ButtonDefaults.elevatedButtonElevation(),
                        ) {
                            Icon(
                                imageVector = Icons.Filled.VisibilityOff,
                                contentDescription = stringResource(MR.strings.action_toggle_private_on),
                            )
                        }
                    }
                }
```

AntsyLich (changes requested):

i18n/src/commonMain/moko-resources/base/strings.xml
```diff
@@ -515,12 +515,12 @@
     <string name="tracking_guide">Tracking guide</string>
     <string name="pref_auto_update_manga_sync">Update progress after reading</string>
     <string name="pref_auto_update_manga_on_mark_read">Update progress when marked as read</string>
+    <string name="pref_private_tracking">Enable private tracking when supported</string>
```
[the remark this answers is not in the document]
NarwhalHorns: i fuckin new i would forget to remove that

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
```diff
@@ -183,6 +184,8 @@ class Anilist(id: Long) : BaseTracker(id, "AniList"), DeletableTracker {
     override suspend fun bind(track: Track, hasReadChapters: Boolean): Track {
         val remoteTrack = api.findLibManga(track, getUsername().toInt())
         return if (remoteTrack != null) {
+            remoteTrack.private = track.private
+
```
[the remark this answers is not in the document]
NarwhalHorns: ah I had this to make sure we use the user track private setting over the remote setting if it is different as that should take priority if they chose it manually, letting the copy do it all would overwrite it if the remote is public and they selected private i think? not sure if there's a better way to  go about it tho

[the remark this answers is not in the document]
NarwhalHorns: hmm but actually i guess we only want the remote to override if tracker.supportsPrivate is false as there is a chance it does support but isn't implemented yet, so could risk overwriting without knowing. Probably best to just handle it per tracker instead of in the copy then?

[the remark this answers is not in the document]
NarwhalHorns: sometthing like this maybe?
image

MajorTanya:
The merge of #1748 means you could add Bangumi support with this too, as that has added a "private" attribute to the relevant DTO. If you don't wanna do that (which, fair enough), just give me a heads up and I'll do it once this PR goes through👍

NarwhalHorns:
> The merge of #1748 means you could add Bangumi support with this too, as that has added a "private" attribute to the relevant DTO. If you don't wanna do that (which, fair enough), just give me a heads up and I'll do it once this PR goes through👍

I don't use it so wouldn't really be able to test it properly, but if it's as simple as adding a couple lines yeah sure

MajorTanya:
> > The merge of #1748 means you could add Bangumi support with this too, as that has added a "private" attribute to the relevant DTO. If you don't wanna do that (which, fair enough), just give me a heads up and I'll do it once this PR goes through👍
> 
> I don't use it so wouldn't really be able to test it properly, but if it's as simple as adding a couple lines yeah sure

It's done using the same add/update methods currently used for adding/updating a library entry. Should just be one copy of `put("private", track.private)` or whatever in each of the BangumiApi.kt methods, and then handing that value over as is appropriate by your process.

NarwhalHorns:
> It's done using the same add/update methods currently used for adding/updating a library entry. Should just be one copy of `put("private", track.private)` or whatever in each of the BangumiApi.kt methods, and then handing that value over as is appropriate by your process.

I think that should do it, i'm not 100% tho. Been a week since my brain ejected the finer details of this. Can you give it a quick test if you have bangumi?

MajorTanya:
Okay, so it works fine for newly added entries, and switching between public/private in linked entries also works. What doesn't work right now is tracking something as private if it's already in the user's list as a public entry. (i.e. I have EntryA in my list as public already but when I search it in Mihon, I select private tracking. This is not reflected on Bangumi's website until I switch to public tracking, then back to private tracking.)

That's because we don't update Bangumi at all if we find remote data for a track:
https://github.com/mihonapp/mihon/blob/main/app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/Bangumi.kt#L53-L63

Not sure what the best approach would be here. When I removed the update call here, it made sense to me but apparently it wasn't that smart. Maybe I should hijack my emergeny linking fix #1770 or maybe I should make one update call anyway.

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Added option to enable incognito per extension ([@sdaqo](https://github.com/sdaqo), [@AntsyLich](https://github.com/AntsyLich)) ([#157](https://github.com/mihonapp/mihon/pull/157))
 - Add button to favorite manga from history screen ([@Animeboynz](https://github.com/Animeboynz)) ([#1733](https://github.com/mihonapp/mihon/pull/1733))
 - Add Monochrome theme (made with e-ink displays in mind) ([@MajorTanya](https://github.com/MajorTanya)) ([#1752](https://github.com/mihonapp/mihon/pull/1752))
+- Support for private tracking with AniList ([@NarwhalHorns](https://github.com/NarwhalHorns)) ([#1736](https://github.com/mihonapp/mihon/pull/1736))
+- Support for private tracking with Bangumi ([@MajorTanya](https://github.com/MajorTanya)), ([@NarwhalHorns](https://github.com/NarwhalHorns)) ([#1736](https://github.com/mihonapp/mihon/pull/1736))
```
AntsyLich, line 20:
```suggestion
- Support for private tracking with AniList and Bangumi ([@NarwhalHorns](https://github.com/NarwhalHorns)) ([#1736](https://github.com/mihonapp/mihon/pull/1736))
```

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
```diff
@@ -32,12 +32,15 @@ interface Track : Serializable {
 
     var tracking_url: String
 
-    fun copyPersonalFrom(other: Track) {
+    var private: Boolean
+
+    fun copyPersonalFrom(other: Track, useLocalPrivate: Boolean = false) {
```
AntsyLich, line 37:
```suggestion
    fun copyPersonalFrom(other: Track, copyRemotePrivate: Boolean = true) {
```

app/src/main/java/eu/kanade/tachiyomi/data/database/models/Track.kt
```diff
@@ -32,12 +32,15 @@ interface Track : Serializable {
...
         last_chapter_read = other.last_chapter_read
         score = other.score
         status = other.status
         started_reading_date = other.started_reading_date
         finished_reading_date = other.finished_reading_date
+        private = if (useLocalPrivate) this.private else other.private
```
AntsyLich, line 43:
```suggestion
        if (copyRemotePrivate) private = other.private
```

app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
```diff
@@ -183,7 +184,7 @@ class Anilist(id: Long) : BaseTracker(id, "AniList"), DeletableTracker {
     override suspend fun bind(track: Track, hasReadChapters: Boolean): Track {
         val remoteTrack = api.findLibManga(track, getUsername().toInt())
         return if (remoteTrack != null) {
-            track.copyPersonalFrom(remoteTrack)
+            track.copyPersonalFrom(remoteTrack, true)
```
AntsyLich, line 187:
```suggestion
            track.copyPersonalFrom(remoteTrack, copyRemotePrivate = false)
```

app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/Bangumi.kt
```diff
@@ -49,7 +51,7 @@ class Bangumi(id: Long) : BaseTracker(id, "Bangumi") {
     override suspend fun bind(track: Track, hasReadChapters: Boolean): Track {
         val statusTrack = api.statusLibManga(track, getUsername())
         return if (statusTrack != null) {
-            track.copyPersonalFrom(statusTrack)
+            track.copyPersonalFrom(statusTrack, true)
```
AntsyLich, line 54:
```suggestion
            track.copyPersonalFrom(statusTrack, copyRemotePrivate = false)
```

AntsyLich (changes requested):

AntsyLich (approved):

AntsyLich:
Can you also add kitsu support?

NarwhalHorns:
> Can you also add kitsu support?

no idea, can look into it but i only use anilist so can't test it and would be going in blind

MajorTanya:
Kitsu's docs are probably the best ones we have 😅 but I've heard the call, I'll make a separate one for Kitsu.

PR #1748, opened 2025-02-19
Migrate to Bangumi's newer v0 API

MajorTanya:
This comes with many benefits:
- Starting dates are now available and shown to users
- Lays groundwork to add private tracking for Bangumi, e.g. in #1736
- Mihon makes approximately 2-4 times fewer calls to Bangumi's API
- Simplified interceptor for the access token addition
  - v0 does not allow access tokens in the query string
- There is [actively maintained documentation](https://bangumi.github.io/api/) for it

Also shrunk the DTOs for Bangumi by removing attributes we have no use for either now or in the foreseeable future. Volume data remains in case Mihon wants to ever support volumes. But attributes such as user avatars, nicknames, data relating to Bangumi's tag & meta-tag systems, etc. have been removed or just not added to the DTOs.

If the volume data attributes should also be removed, I can do that. 👍

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/Bangumi.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/Bangumi.kt
@@ -9,7 +9,6 @@ import eu.kanade.tachiyomi.data.track.bangumi.dto.BGMOAuth
 import eu.kanade.tachiyomi.data.track.model.TrackSearch
 import kotlinx.collections.immutable.ImmutableList
 import kotlinx.collections.immutable.toImmutableList
-import kotlinx.serialization.encodeToString
 import kotlinx.serialization.json.Json
 import tachiyomi.i18n.MR
 import uy.kohesive.injekt.injectLazy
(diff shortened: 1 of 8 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/BangumiUtils.kt
```diff
@@ -3,10 +3,10 @@ package eu.kanade.tachiyomi.data.track.bangumi
 import eu.kanade.tachiyomi.data.database.models.Track
 
 fun Track.toApiStatus() = when (status) {
```
AntsyLich, line 5:
```suggestion
fun Track.toApiStatus() = status.toInt()
```

app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/dto/BGMCollectionResponse.kt
```diff
@@ -1,28 +1,34 @@
...
-)
+    val private: Boolean = false,
+    val subject: BGMSlimSubject? = null,
+) {
+    fun getStatus(): Long = when (type) {
+        1 -> Bangumi.PLAN_TO_READ
+        2 -> Bangumi.COMPLETED
+        3 -> Bangumi.READING
+        4 -> Bangumi.ON_HOLD
+        5 -> Bangumi.DROPPED
```
AntsyLich, line 24:
```suggestion
        in 1..5 -> type.toLong()
```

AntsyLich (changes requested):

AntsyLich (approved):

app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/Bangumi.kt
```diff
@@ -112,9 +105,13 @@ class Bangumi(id: Long) : BaseTracker(id, "Bangumi") {
     suspend fun login(code: String) {
         try {
             val oauth = api.accessToken(code)
+            // Users can set a 'username' (not nickname) once which effectively
+            // replaces the stringified ID in certain queries.
+            // If no username is set, the API returns the user ID as a strings
+            var username = api.getUsername()
```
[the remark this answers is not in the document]
MajorTanya: Leftover from when I thought I'd have to fiddle with it if the name wasn't set. Never saw that it was var after I figured out they return the stringified ID if no name is set. Thanks for the hint but it's all merged now anyway and I'm not making a third fixup PR for this.

PR #1775, opened 2025-02-25
Rework build variants and add FOSS variant

AntsyLich:


papjul:
I highly suggest you enable reproducible builds on F-Droid side. This means you will build and sign the `foss` variant with your key, and F-Droid will also build it on their side and take your APK with your signature, if the APK they built strictly matches yours (excluding the signature of course).

That's why you can include the updater in the `foss` variant because the signature will match, especially as some users may just use it outside the F-Droid store (getting updates from GitHub).

papjul:
The updater is not a prerequisite, indeed.

I would say reproducible builds kinda are, since it's hard to switch from non-reproducible to reproducible after joining the F-Droid repo. You will have a mix of non-reproducible (signed by F-Droid) and reproducible to ensure users who initially installed the non-reproducible can still update.

To know if the builds are reproducible, you will need to apply, send them the link to your builds, and they will tell you if they were able to make the exact same build. If there are differences, they will tell you which ones.

Syer10 (approved):
I like it, nice way to do analytics

linsui:
About the updater, F-Droid has a clear policy

> The software must not download additional executable binary files (e.g. addons, auto-updates, etc.) without explicit user consent. Consent means it needs to be opt-in (it must not be harder to decline than to accept or presented in a way users are likely to press accept without reading) and structured in a way that clearly explains to users that they’re choosing to bypass F-Droid’s checks if they activate it.

AntsyLich:
Updater just sends a system notification

PR #1779, opened 2025-02-26
use artist and author in duplicate search if available

NarwhalHorns:
Simple as really, if both database and query manga have artist or author, don't consider them duplicates if they are different.

### DIFF
--- a/domain/src/main/java/tachiyomi/domain/manga/interactor/GetDuplicateLibraryManga.kt
+++ b/domain/src/main/java/tachiyomi/domain/manga/interactor/GetDuplicateLibraryManga.kt
@@ -8,6 +8,11 @@ class GetDuplicateLibraryManga(
 ) {
 
     suspend fun await(manga: Manga): List<Manga> {
-        return mangaRepository.getDuplicateLibraryManga(manga.id, manga.title.lowercase(), manga.artist?.lowercase(), manga.author?.lowercase())
+        return mangaRepository.getDuplicateLibraryManga(
+            manga.id,
+            manga.title.lowercase(),
+            manga.artist?.lowercase(),
+            manga.author?.lowercase(),
+        )
     }
 }


AntsyLich:
And the justification being?

NarwhalHorns:
> And the justification being?

so manga that have the same title but are not duplicates don't get called up as duplicates? Pretty self explanatory

BrutuZ:
> I would rather rely on sources having some level of artist/author being good

In an ideal world, sure, but the reality is harsh. You might get some smidge of consistency on a single source, but the moment you start comparing different ones it all goes out the window

NarwhalHorns:
Could attempt some kinda of similarity check? but can see that being absurd to do reliably

AntsyLich:
Title are more or less common across sources but it's not really the case for author/artist names. This in my opinion will do more harm than good.

NarwhalHorns:
yeah i'm starting to agree, unless anyone else has any very bright ideas probably not worth it

cuong-tran:
maybe instead of that, showing author/artist name in duplicate dialog so user can be aware?

BrutuZ:
A little redundant with the _Show Entry_ button right there, no?

NarwhalHorns:
idk having to go and check by manually looking at the other entry every time is bothersome. If we could just get a little info on the duplicate menu saying like "artist match" or something would be nice

cuong-tran:
> A little redundant with the _Show Entry_ button right there, no?

that works, but too many actions. Show entry, go back, hit favorite again.

Also, even with that won't solve the problem if there are more than one duplicate entries, it always show the first one while that one might not be the real duplicate but the other one.

BrutuZ:
I just can't imagine a scenario where you'd have so many duplicate titles for that to be an issue.
<details><summary>extra</summary>
<p>
I'm not particularly conservative with my library (7k+) and sometimes do try to re-add the same one-shot entry, but far from being often enough to be "annoying" and I don't think it was ever a false-positive.
</p>
</details> 

And that's the last I'll speak of it here, before comments get locked 🤐. Feel free to poke me at Discord to continue

PR #1780, opened 2025-02-26
Address various IDE warnings

AntsyLich:


app/src/main/java/eu/kanade/tachiyomi/data/track/bangumi/Bangumi.kt
```diff
@@ -111,8 +111,7 @@ class Bangumi(id: Long) : BaseTracker(id, "Bangumi") {
             // Users can set a 'username' (not nickname) once which effectively
             // replaces the stringified ID in certain queries.
             // If no username is set, the API returns the user ID as a strings
-            var username = api.getUsername()
-            saveCredentials(username, oauth.accessToken)
+            saveCredentials(api.getUsername(), oauth.accessToken)
```
MajorTanya, line 114: Why this change?

AntsyLich: thought it looked better idk

PR #1794, opened 2025-02-28
Mitigate the libraryView NPE crash that occurs for large libraries when removing entries shortly after any other DB write

FlaminSarge:
Currently, users with huge libraries hit a NullPointerException when reading libraryView due to the LibraryScreenModal subscribing to the library view Flow, as unfavoriting a manga removes a row from that library view while it is still being read.

This papers over the NPE that gets throw and allows the app to recover if a user removes an entry while that view is still being read by the flow collector.


AntsyLich:
Superseded by #1799

PR #1802, opened 2025-03-01
fix "Multiple substitutions specified in non-positional format of string resource string/restore_duration"

rhjdvsgsgks:
string replaced by using following command
```bash
sed -i "s/%02d/%1\$02d/" i18n/src/commonMain/moko-resources/*/strings.xml
sed -i "s/%02d/%2\$02d/" i18n/src/commonMain/moko-resources/*/strings.xml
```

but im not sure is that the same order in every language


MajorTanya:
I'm pretty sure we use Weblate for localisation efforts, so if the base values have positional arguments set correctly, it is on those localisation efforts to ensure proper ordering.

AntsyLich:
Some of these are not proper translation and this warning is already fixed in #1550

PR #1803, opened 2025-03-01
add BUILD_TOOLS_VERSION to buildlogic

rhjdvsgsgks:
this is beneficial to reproduce the build environment as official


MajorTanya:
Is it a good idea to remove it from the workflow?

MajorTanya:
> > Is it a good idea to remove it from the workflow?
> 
> since it is same as value in buildlogic. i dont think there is any meaning to keep it in workflow

True enough I guess.

PR #1806, opened 2025-03-02
Allow extensions to load native libraries

TheUnlocked:
Currently the classloader that gets created when an extension is loaded doesn't pass a path for the extension's native libraries, which means `System.loadLibrary` will only be able to search for libraries in system directories and not for libraries packaged in the extension. This change would pass through the extension APK's native libraries directory, allowing extensions that include native libraries to load those native libraries.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/extension/util/ExtensionLoader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/extension/util/ExtensionLoader.kt
@@ -272,7 +272,7 @@ internal object ExtensionLoader {
         }
 
         val classLoader = try {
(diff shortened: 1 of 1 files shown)


AntsyLich:
Not sure if I want to allow extensions to load native libraries

TheUnlocked:
Today it's already technically possible for extensions to load native libraries by creating their own classloader and using reflection to create instances of classes provided by the extension under that classloader. I did that to work around this limitation when experimenting with running AI models in an extension (which require native libraries for performance). Including the APK's native library path in the extension classloader just makes it so that extension developers don't need to jump through those hoops.

AntsyLich:
Sure it's possible but it's still not something we support intentionally. Given future extension API/loader/binary planned iterations will lead to sandboxing I see no pint to support this.

PR #1831, opened 2025-03-06
Change foss variant application id suffix to '.foss' and more

AntsyLich:
- Remove `BuildConfig.PREVIEW`
- Rename `BuildConfig.INCLUDE_ANALYTICS` -> `BuildConfig.ANALYTICS_INCLUDED`
- Rename `BuildConfig.INCLUDE_UPDATER` -> `BuildConfig.UPDATER_ENABLED`
- Rename build property `with-analytics` -> `include-analytics`
- Rename build property `with-updater` -> `enable-updater`
- Add build property to disable code shrink
- Add build property to include dependency info in apk/app bundle

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -10,14 +10,15 @@ plugins {
     alias(libs.plugins.aboutLibraries)
 }
 
-class ConfigBacking {
+class ConfigClass {
     val includeAnalytics: Boolean = project.hasProperty("include-analytics")
     val enableUpdater: Boolean = project.hasProperty("enable-updater")
     val enableCodeShrink: Boolean = !project.hasProperty("disable-code-shrink")
     val includeDependencyInfo: Boolean = project.hasProperty("include-dependency-info")
 }
+
 @Suppress("PropertyName")
-val Config = ConfigBacking()
+val Config = ConfigClass()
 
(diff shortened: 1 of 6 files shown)


app/build.gradle.kts
```diff
@@ -10,10 +10,12 @@ plugins {
     alias(libs.plugins.aboutLibraries)
 }
 
-val includeAnalytics = project.hasProperty("with-analytics")
-val includeUpdater = project.hasProperty("with-updater")
+val configIncludeAnalytics = project.hasProperty("with-analytics")
+val configIncludeUpdater = project.hasProperty("with-updater")
+val configCodeShrink = !project.hasProperty("no-code-shrink")
+val configPackageDependenciesInfo = !project.hasProperty("include-dependencies-info")
```
MajorTanya, line 16: Why are these inverted? Also, why did you pick boolean names for the other two but not these two?

dewantawsif: The inversion is wrong

MajorTanya:
I assume you changed to a class because the object didn't properly get the attributes?

AntsyLich:
More like build fails if I make it an object
> Object Config captures the script class instance. Try to use class or anonymous object instead

PR #1835, opened 2025-03-07
Showing in-app progress bar for backup restoring & library updating

cuong-tran:
Should we also add one for backup creating?


AntsyLich:
Backup creation doesn't take enough time to justify a banner

app/src/main/java/eu/kanade/presentation/components/Banners.kt
```diff
@@ -71,12 +86,26 @@ fun AppStateBanners(
...
+                    text = when {
+                        updating -> progress?.let {
+                            stringResource(
+                                MR.strings.notification_updating_progress,
+                                percentFormatter.format(it),
+                            )
+                        } ?: stringResource(MR.strings.updating_library)
+
+                        restoring -> progress?.let {
+                            stringResource(MR.strings.restoring_backup) + " (${percentFormatter.format(it)})"
```
MajorTanya, line 104: Naive string concat is a really bad idea in a project with proper i18n

cuong-tran: Right, better create a new string for it. Thanks.

cuong-tran:
There is one option in Settings/Advance to refresh cover using a job. Does it need progress bar too?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Display staff information on Anilist tracker search results ([@NarwhalHorns](https://github.com/NarwhalHorns)) ([#1810](https://github.com/mihonapp/mihon/pull/1810))
 - Add `id:` prefix search to library to search by internal DB ID ([@MajorTanya](https://github.com/MajorTanya)) ([#1856](https://github.com/mihonapp/mihon/pull/1856))
 - Add back option to disable unread chapter badge in library ([@AntsyLich](https://github.com/AntsyLich)) ([#1871](https://github.com/mihonapp/mihon/pull/1871))
+- Showing in-app progress bar for backup restoring & library updating ([@cuong-tran](https://github.com/cuong-tran)) ([#1835](https://github.com/mihonapp/mihon/pull/1835))
```
AntsyLich, line 32: Move to unreleased

app/src/main/java/eu/kanade/presentation/components/Banners.kt
```diff
@@ -58,25 +65,44 @@ fun WarningBanner(
...
+                        updating -> progress?.let {
+                            stringResource(MR.strings.notification_updating_progress, percentFormatter.format(it))
+                        } ?: stringResource(MR.strings.updating_library)
+
+                        restoring -> progress?.let {
+                            stringResource(MR.strings.restoring_backup_progress, percentFormatter.format(it))
+                        } ?: stringResource(MR.strings.restoring_backup)
+
+                        else -> stringResource(MR.strings.download_notifier_cache_renewal)
+                    },
```
AntsyLich, line 105: Is it worth it show multiple at once instead?

AntsyLich (changes requested):

app/src/main/java/eu/kanade/presentation/components/Banners.kt
cuong-tran: They might stack up too much space on top bar. I don't think it will look good.
Also, it's not like those actions are often running together at same time.

PR #1843, opened 2025-03-09
Tweak the app updater logic and add FOSS build support

AntsyLich:


domain/src/main/java/tachiyomi/domain/release/interactor/GetApplicationRelease.kt
```diff
@@ -25,7 +25,7 @@ class GetApplicationRelease(
             return Result.NoNewUpdate
         }
 
-        val release = service.latest(arguments.repository)
+        val release = service.latest(arguments) ?: return Result.NoNewUpdate
```
MajorTanya, line 28: Does this obsolete the when-else below? L39-42
https://github.com/mihonapp/mihon/pull/1843/files#diff-83eb064e155870afb25e791c126145476f50ea159e0d6501fdb0c43df603d8acR41

domain/src/test/java/tachiyomi/domain/release/interactor/GetApplicationReleaseTest.kt
MajorTanya, line 1: Could probably do with a test or two for the `isFoss = true` behaviour

data/src/main/java/tachiyomi/data/release/ReleaseServiceImpl.kt
```diff
@@ -13,13 +15,46 @@ class ReleaseServiceImpl(
...
+            version = release.version,
+            info = release.info.replace(gitHubUsernameMentionRegex) { mention ->
+                "[${mention.value}](https://github.com/${mention.value.substring(1)})"
+            },
+            releaseLink = release.releaseLink,
+            downloadLink = downloadLink,
+        )
+    }
+
+    private fun GithubRelease.getDownloadLink(isFoss: Boolean): String? {
```
MajorTanya, line 38: Could this be a normal private function? `private fun getDownloadLink(release: GithubRelease, isFoss: Boolean): String?`-like?

data/src/main/java/tachiyomi/data/release/GithubRelease.kt
```diff
@@ -12,14 +11,17 @@ data class GithubRelease(
     @SerialName("tag_name") val version: String,
     @SerialName("body") val info: String,
     @SerialName("html_url") val releaseLink: String,
-    @SerialName("assets") val assets: List<GitHubAssets>,
+    @SerialName("assets") val assets: List<GitHubAsset>,
 )
 
 /**
  * Assets class containing download url.
```
MajorTanya, line 18:
```suggestion
 * Asset class containing asset name and download url.
```

data/src/main/java/tachiyomi/data/release/ReleaseServiceImpl.kt
```diff
@@ -13,13 +15,46 @@ class ReleaseServiceImpl(
...
+        val map = assets.associate { asset ->
+            TYPES.find { "-$it" in asset.name } to asset.downloadLink
+        }
+
+        return if (!isFoss) {
+            map[Build.SUPPORTED_ABIS[0]] ?: map[null]
+        } else {
+            map[FOSS]
+        }
```
MajorTanya, line 47: Not sure if I'm reading this correctly but does this mean FOSS will only get universal?

domain/src/main/java/tachiyomi/domain/release/interactor/GetApplicationRelease.kt
AntsyLich: no

domain/src/test/java/tachiyomi/domain/release/interactor/GetApplicationReleaseTest.kt
AntsyLich: Maybe in the future.

data/src/main/java/tachiyomi/data/release/GithubRelease.kt
```diff
@@ -12,14 +11,17 @@ data class GithubRelease(
...
 
 /**
- * Assets class containing download url.
+ * Asset class containing asset name and download url.
  */
 @Serializable
-data class GitHubAssets(@SerialName("browser_download_url") val downloadLink: String)
+data class GitHubAsset(
+    val name: String,
+    @SerialName("browser_download_url") val downloadLink: String,
```
MajorTanya, line 23: Didn't see this before but annotations go above values not in front.

MajorTanya (changes requested):
I think other than this it looks good. I have no idea how to verify it but I'm sure you will have made a FOSS build locally

domain/src/test/java/tachiyomi/domain/release/interactor/GetApplicationReleaseTest.kt
```diff
@@ -39,13 +39,14 @@ class GetApplicationReleaseTest {
             "r2000",
             "info",
             "http://example.com/release_link",
-            listOf("http://example.com/assets"),
+            "http://example.com/release_link.apk",
         )
 
         coEvery { releaseService.latest(any()) } returns release
 
         val result = getApplicationRelease.await(
```
cuong-tran, line 47: all the cast like this `(result as GetApplicationRelease.Result.NewUpdate).release` below are causing crash when trying to force wrong value (to replicate a failed test)

AntsyLich: not sure why you would want to add new test to an existing one

cuong-tran: I'm not talking about new test, but when I make change to the updating logic and run test to see if all these current tests will pass or fail. When the logic is wrong, they are supposed to send a FAILED result to the test, but the test would actually crash because of the cast failed.

MajorTanya: Why do we even cast without a check first? If we had a simple assertion first that `result is GetApplicationRelease.Result.NewUpdate` checks out, we'd avoid the crashes. Generally a must-have before casting anything.

AntsyLich: I tested and it's not crashing anything. The test is failing as it should.

PR #1849, opened 2025-03-10
Fix duplicate tracks created when migrating manga

cuong-tran:
Migration from old manga to new manga which already has trackers.


AntsyLich:
I think we should take some inputs on how to handle this. Cause I don't think people will track wrong thing so I don't feel like deleting the existing ones

cuong-tran:
It's not because people track wrong thing but if 2 entries are tracking the same one from same tracker, while doing migration it will create a new track (from old entry) into new entry. Now new entry will have 2 track of same tracker, same remote one.

On a second thought while typing this, database already has a confict-replace condition so this is no needed. My bad, will close this.

PR #1850, opened 2025-03-10
Reset manga date when removed from library

cuong-tran:


BrutuZ:
Isn't it already removed? I know I've had to restore backups a couple of times to fix old entries having their `dateAdded` updated after accidentally tapping the _In Library_ button then adding them back

cuong-tran:
> Isn't it already removed? I know I've had to restore backups a couple of times to fix old entries having their `dateAdded` updated after accidentally tapping the _In Library_ button then adding them back

it only mark `favorite = false`, not actually removed from database.

cuong-tran:
I'm not sure if this matters, because when adding entry back to library, it will update new time, while the un-favorited entry in database doesn't really matter anywhere. It's just every other unfavorite action is reseting dateAdded to 0, except this one.

BrutuZ:
Is it really an issue since it gets overwritten anyway if added back?

BrutuZ:
Personally, I'd be more interested in it _not_ updating dates when adding back. But I'm aware that would just suit my OCD and also break a workaround used by some to manually sort their libraries, so no point in opening a request

cuong-tran:
Agree. This only tried to make every place consistent.

AntsyLich:
I think it's fine as is

PR #1851, opened 2025-03-11
Rework website update workflow

scb261:
Follow up for #1818
See also: https://github.com/mihonapp/mihon-preview/pull/35

Why: the previous implementation didn't work properly because the build CI does only a draft release. This PR separates the website update into a separate workflow that is triggered whenever an actual release is published, deleted, or edited.

Also added --fail-with-body to curl so that the action fails when the event couldn't be sent successfully.

### DIFF
--- a/.github/workflows/build_push.yml
+++ b/.github/workflows/build_push.yml
@@ -122,14 +122,3 @@ jobs:
           prerelease: false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Update website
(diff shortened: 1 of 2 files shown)


.github/workflows/build_push.yml
```diff
@@ -122,14 +122,3 @@ jobs:
           prerelease: false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
AntsyLich, line 124:
```suggestion
          GITHUB_TOKEN: ${{ secrets.MIHON_BOT_TOKEN }}
```

AntsyLich (changes requested):

AntsyLich (approved):

PR #1861, opened 2025-03-16
Handle multiple duplicates on add duplicate dialogue

NarwhalHorns:
Adds functionality to show and choose between all the duplicates found when adding a manga instead of just the first one found.

image


app/src/main/java/eu/kanade/presentation/manga/DuplicateMangaDialog.kt
```diff
@@ -120,6 +146,118 @@ fun DuplicateMangaDialog(
...
+    onOpenManga: () -> Unit,
+    onMigrate: () -> Unit,
+) {
+    val shape = RoundedCornerShape(16.dp)
+    Box(
+        modifier = Modifier
+            .fillMaxWidth()
+            .clip(shape)
+            .background(MaterialTheme.colorScheme.surface)
+            .padding(12.dp)
```
cuong-tran, line 162:
```suggestion
            .padding(MaterialTheme.padding.mediumSmall)
```

NarwhalHorns: oh that doesn't exist, added it

app/src/main/java/eu/kanade/presentation/manga/DuplicateMangaDialog.kt
```diff
@@ -120,7 +146,115 @@ fun DuplicateMangaDialog(
...
+                            SManga.CANCELLED.toLong() -> Icons.Outlined.Close
+                            SManga.ON_HIATUS.toLong() -> Icons.Outlined.Pause
+                            else -> Icons.Outlined.Block
+                        },
+                        contentDescription = null,
+                        modifier = Modifier
+                            .padding(end = 4.dp)
+                            .size(16.dp),
+                    )
+                    ProvideTextStyle(MaterialTheme.typography.bodyMedium) {
```
MajorTanya, line 236: Why this wrapper instead of the style passing used basically everywhere else? 🤔

NarwhalHorns: I copied these from the MnagaInfoHeader so no idea, can swap to something more typical if you want

MajorTanya: Oh I see. You can keep it, I was just surprised since I hadn't seen this before and was wondering if there was some special benefit to using it here.

NarwhalHorns: nah, there ya go. cleaned up a bit

AntsyLich:
why not just have the entries be places similar to comfortable library display mode

NarwhalHorns:
> why not just have the entries be places similar to comfortable library display mode

hmm, and just have the author/artists underneath it or something? not sure what you mean. I wanted to try to provide as much info to the user to differentiate dupes or not but there's not actually that much lmao so yeah ended up being kinda empty on the right side

MajorTanya:
What about this artist's impression?


> Graphic design is my passion

NarwhalHorns:
> What about this artist's impression? 
> 
> > Graphic design is my passion

don't need the title as they are the same but yeah i like putting the source at the bottom like that, will try it

NarwhalHorns:
> If we go with that, make sure we have all the cards be the same height

how lol

cuong-tran:
> > If we go with that, make sure we have all the cards be the same height
> 
> how lol

truncate the title, shouldn't they have same title which already known to user?

NarwhalHorns:
> > > If we go with that, make sure we have all the cards be the same height
> > 
> > 
> > how lol
> 
> truncate the title, shouldn't they have same title which already known to user?

they don't have a title?

NarwhalHorns:
> No idea. Maybe some min height constraint or something? I'm sure there's a thing for this somewhere in Compose.

hmm, it's complicated cus they have no set height, somehow need to make them all as tall as the tallest one

NarwhalHorns:
so got this but also this
| nice | oof |
| ---- | ----- |
| image | image |

not ideal but i think it's the best we can get

MajorTanya:
Looking at this makes me wanna see titles tbh. Especially when you have some sources that may have "Official" or "Coloured" or whatever other title addendums.

NarwhalHorns:
> Looking at this makes me wanna see titles tbh. Especially when you have some sources that may have "Official" or "Coloured" or whatever other title addendums.

well, currently they wouldn't be caught. we only find duplicates by an exact match. BUT, if we are thinking about mkaing the query a more robust search then having the titles would make sense. so i guess we cna put the titles in for now and they would be identical or add it if/when we get more complex search?

NarwhalHorns:
> And if/when we make the search more fuzzy, we could just throw a good old "LIKE" clause into the SQL (assuming that's what we use for the query)

Yeah I was thinking a few different things but that would be a good first step... might try doing that now and see how it comes out actually

MajorTanya:
###### ~similar names~

NarwhalHorns:
> ###### ~similar names~
> 
> Anway yeah, this looks alright in the title department.

fuck

NarwhalHorns:
was a little worried the fuzzy search would impact performance but tried it on my library with 15000 entries and it's fast as fook booooi. so no worries there

MajorTanya:
> was a little worried the fuzzy search would impact performance but tried it on my library with 15000 entries and it's fast as fook booooi. so no worries there

SQL is fast as fuck boooiiiii

NarwhalHorns:
> > was a little worried the fuzzy search would impact performance but tried it on my library with 15000 entries and it's fast as fook booooi. so no worries there
> 
> SQL is fast as fuck boooiiiii

While i was googling stuff i saw that we could use Full-Text Search (FTS) tables which make text based searches faster somehow, so was worried it might be slow without but seemingly not. but if we add more complex searches on top it might start impacting performance on large libraries? possibility if so but considering how it is atm probably not

MajorTanya:
> > > was a little worried the fuzzy search would impact performance but tried it on my library with 15000 entries and it's fast as fook booooi. so no worries there
> > 
> > SQL is fast as fuck boooiiiii
> 
> While i was googling stuff i saw that we could use Full-Text Search (FTS) tables which make text based searches faster somehow, so was worried it might be slow without but seemingly not. but if we add more complex searches on top it might start impacting performance on large libraries? possibility if so but considering how it is atm probably not

I don't think there's anything to gain from using FTS tables (not even sure SQLite in Android supports them). In general, I find people to be way over-obsessed over "fast" and "speed" and "optimised". SQL is designed to work on systems with billions of rows in lots and lots of tables, so out somewhat small use case is a cake walk for any SQL engine.
Filtering in Kotlin would still be fine (we have to do that in some places anyway) but obviously """slower""".

JDKamalakar:
> LIKE fuzzy search
> 
> image

Hey small suggestion move extension name to top (Above cover image).

PR #1869, opened 2025-03-18
Make and include FOSS build on release

AntsyLich:


linsui:
I'll draft a build metadata for it. :)

AntsyLich:
Btw if we ever end up doing a full rewrite would that need to be under a different package name?

linsui:
That's up to you. :)

linsui:
Can you add "foss" as a flavor instead of a build type?

linsui:
There is special support for gradle in fdroidserver. It's designed to work with different flavors instead of build type and the build type is always release. If we use foss as build type then we need to write the build metadata without the gradle support. If you can use flavor instead of build type it would be much more convenient. :)

linsui:
Should we build a universal apk or split apks? If we want split apks then each of them should has a different version code, e.g. 11, 12, 13, 14 for armv7a, arm64-v8a, x86, x86_64.

MajorTanya:
We use versionCode to check for necessary migrations we need to execute against the app's database so I believe that's out of the question.

AntsyLich:
> There is special support for gradle in fdroidserver. It's designed to work with different flavors instead of build type and the build type is always release. If we use foss as build type then we need to write the build metadata without the gradle support. If you can use flavor instead of build type it would be much more convenient. :)

Having it as build type is a bit convenient and preferred for me (also why I removed flavors that were present previously)

> We use versionCode to check for necessary migrations we need to execute against the app's database so I believe that's out of the question.

Database is different. But version code is still uses
D in the app for migrations so can't have that anytime soon.

MajorTanya:
> > We use versionCode to check for necessary migrations we need to execute against the app's database so I believe that's out of the question.
> 
> Database is different. But version code is still used in the app for migrations so can't have that anytime soon.

We still use it for migrations anyway so my point is the same.

AntsyLich:
> Can you build an apk from master so that I can test reproducible build?
>

Where do you want me to put it afterwards?

> See [mikepenz/AboutLibraries#784](https://github.com/mikepenz/AboutLibraries/issues/784). It's the time when the list is generated.
>

We also include the build time in the app...

linsui:
> Where do you want me to put it afterwards?

Not sure what you mean. For the test apk you can put it here. For the release. can you put it in release?

> We also include the build time in the app...

Can you use the time of the commit instead?

AntsyLich:
> > Where do you want me to put it afterwards?
> 
> Not sure what you mean. For the test apk you can put it here. For the release. can you put it in release?
> 
I meant for the apk you wanted. As for release that's the whole point of this PR
> > We also include the build time in the app...
> 
> Can you use the time of the commit instead?
>

Yeah that also works

linsui:
I'll check the apk manually so you can simply attach it here.

AntsyLich:
> Can you build an apk from master so that I can test reproducible build? Please add this to remove the timestamp.
> 
> ```kotlin
> aboutLibraries {
>     excludeFields = arrayOf("generated")
> }
> ```

Seems like those are not included by default since v11.6.3 (https://github.com/mikepenz/AboutLibraries/pull/1056)

AntsyLich:
APK: https://wormhole.app/kz9xvl#paIj49rY5pbsnbx_lb5EiA
Was built via `gradlew assembleFoss -Penable-updater`

linsui:
Is your apk built on Windows? Will the release apk also be built on Windows?

Some resources in i18n has ordering differences. I guess you need to sort something.

[app-universal-foss.apk.html.zip](https://github.com/user-attachments/files/19332656/app-universal-foss.apk.html.zip)

AntsyLich:
Hmm that file is generated by https://github.com/icerockdev/moko-resources

linsui:
Maybe you can postprocess those files. :shrug:

AntsyLich:
We don't really need that library so I guess removing it is an option

papjul:
> Please add this to remove the timestamp.
> 
> ```kotlin
> aboutLibraries {
>     excludeFields = arrayOf("generated")
> }
> ```

FYI, it should no longer be needed, see mikepenz/AboutLibraries#1056

> APK: https://wormhole.app/kz9xvl#paIj49rY5pbsnbx_lb5EiA Was built via `gradlew assembleFoss -Penable-updater`

Was this built with the same commands as the CI? E.g. from a clean Gradle, build the regular release, then the FOSS one. By experience, building two different flavors with different library dependencies can lead to some inconsistencies on the flavors that come after (and even a `gradlew clean` between the two flavors is not enough), especially as `org.gradle.caching` is enabled.

papjul:
Would it be possible to get this merged?

I'm sure there are people that would be happy to use a FOSS build of v0.19.0 outside of the F-Droid official repo, even if the Moko resources are *technically* in a different order in the APK.

Reproducible builds are also not a requirement for the [IzzyOnDroid repo](https://gitlab.com/IzzyOnDroid/repo/-/issues/) and could serve as an intermediate step before inclusion in the F-Droid official repo.

Thanks!

Note: FYI, since the last discussions, F-Droid official repo now has a fork of Mihon available that hostilely removed all enforced rate limits, which -as people can guess- are there for a reason. It's unclear if the F-Droid team is aware of this and approve it, but as for the Mihon project, if websites start blocking the ability to scrap their resources because it put too much pressure on them, this will make many extensions stop working. Since it's the only option for F-Droid users at the moment, the users are forced to flood websites with requests whenever they download something with the app, which -I hope- most of them would not like to do.

AntsyLich:
I'm looking at replacing crashlytics with sentry. Probably after that.

AntsyLich:
Superseded by #2367

PR #1872, opened 2025-03-18
Add missing references to CHANGELOG.md

MajorTanya:
This should be the last "general CHANGELOG.md maintenance" we need. Future changes (direct commits and PRs) should be strongly encouraged to update the CHANGELOG.md file with their own contribution when they happen.


MajorTanya:
Ignored the line about "mark as read" stuff since you're updating it in #1870

PR #1924, opened 2025-03-26
Auto (Black) for Background Color

### LINKED ISSUE #215: Background Color - Auto (Black)

When the "auto" background color works it looks great, but an option to have it default to black for normal pages (anywhere it would otherwise be white or fall back to white) would be greatly appreciated.

### How it currently works
Screenshot_20240123-133851

How it currently falls back to white;
Screenshot_20240123-134121

How a standard black and white page looks on auto
Screenshot_20240123-134419

### How this proposed new option would look

Screenshot_20240123-133851

Screenshot_20240123-134130

Screenshot_20240123-134430

### Other details

I hope the pictures can help convey what I'm asking for.

Also, the issue template still has a link to the old official extension repo in the acknowledgements.

### Acknowledgements

- [X] If this is an issue with an official extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/extensions/issues/new/choose).
- [X] I have updated the app to version **[0.16.1](https://github.com/mihonapp/mihon/releases/latest)**.

01-prog:
Fixes #215 

Added a button "Auto (Black)" working just as "Auto" works but falls back to black instead of white. (#215)
I added a translation in French but not in other langages.

## How it works 
image

| Auto | Auto (Black) |
| ------- | ------- |
| image | image |
| image | image |
| image | image |

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/reader/settings/GeneralSettingsPage.kt
+++ b/app/src/main/java/eu/kanade/presentation/reader/settings/GeneralSettingsPage.kt
@@ -1,6 +1,5 @@
 package eu.kanade.presentation.reader.settings
 
-import androidx.compose.foundation.layout.ColumnScope
 import androidx.compose.material3.FilterChip
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
@@ -21,6 +20,7 @@ private val themes = listOf(
     MR.strings.black_background to 1,
     MR.strings.gray_background to 2,
     MR.strings.white_background to 0,
+    MR.strings.automatic_background to 3,
 )
 
 private val flashColors = listOf(
@@ -30,7 +30,7 @@ private val flashColors = listOf(
 )
 
 @Composable
-internal fun ColumnScope.GeneralPage(screenModel: ReaderSettingsScreenModel) {
+internal fun GeneralPage(screenModel: ReaderSettingsScreenModel) {
     val readerTheme by screenModel.preferences.readerTheme().collectAsState()
 
     val flashPageState by screenModel.preferences.flashOnPageChange().collectAsState()
@@ -44,10 +44,10 @@ internal fun ColumnScope.GeneralPage(screenModel: ReaderSettingsScreenModel) {
     val flashColorPref = screenModel.preferences.flashColor()
     val flashColor by flashColorPref.collectAsState()
 
-    SwitchItem (MR.strings.auto_background, screenModel.preferences.readerAutomaticBackground())
+    SwitchItem(MR.strings.auto_background, screenModel.preferences.readerAutomaticBackground())
 
     SettingsChipRow(MR.strings.pref_reader_theme) {
-        themes.map { (labelRes, value) ->
+        themes.filter { it.second != 3 }.map { (labelRes, value) ->
             FilterChip(
                 selected = readerTheme == value,
                 onClick = { screenModel.preferences.readerTheme().set(value) },
--- a/app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt
@@ -15,7 +15,6 @@ import androidx.core.content.getSystemService
 import androidx.core.net.toUri
 import com.hippo.unifile.UniFile
 import eu.kanade.domain.ui.UiPreferences
-import eu.kanade.domain.ui.model.ThemeMode
 import eu.kanade.tachiyomi.R
 import eu.kanade.tachiyomi.ui.base.delegate.ThemingDelegate
 import eu.kanade.tachiyomi.ui.reader.setting.ReaderPreferences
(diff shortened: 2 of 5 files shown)


MajorTanya:
No translations. Those are handled by Weblate.

AntsyLich:
Maybe we should do something like this?
image

01-prog:
I will work on this.

01-prog:
I added a switch to simplify. Works with all base themes and tablet mode.

## How it works 
image

| Auto + Black | Auto + Gray | Auto + White|
| ------- | ------- | ------- |
| image | image | image |
| image | image | image |
| image | image | image

app/src/main/java/eu/kanade/tachiyomi/util/system/ContextExtensions.kt
```diff
@@ -108,9 +108,9 @@ fun Context.createReaderThemeContext(): Context {
...
     val themeMode = preferences.themeMode().get()
-    val isDarkBackground = when (readerPreferences.readerTheme().get()) {
-        1, 2 -> true // Black, Gray
-        3 -> when (themeMode) { // Automatic bg uses activity background by default
+    val isDarkBackground = when (Pair(readerPreferences.readerTheme().get(), readerPreferences.readerThemeSwitch().get())) {
+        Pair(1,false), Pair(2,false) -> true // Black, Gray
+        Pair(0,true),Pair(1,true),Pair(2,true) -> when (themeMode) { // Automatic bg uses activity background by default
             ThemeMode.SYSTEM -> applicationContext.isNightMode()
             else -> themeMode == ThemeMode.DARK
         }
```
AntsyLich, line 116:
```suggestion
    val isDarkBackground = when (readerPreferences.readerTheme().get()) {
        1, 2 -> true // Black, Gray
```

presentation-core/src/main/java/tachiyomi/presentation/core/components/SettingsItems.kt
```diff
@@ -381,6 +384,71 @@ fun TextItem(label: String, value: String, onChange: (String) -> Unit) {
...
+                onCheckedChange = null,
+                modifier = Modifier.size(switchSize.dp),
+            )
+        },
+        onClick = onClick,
+    )
+}
+
+@Composable
+fun SettingsChipRowSwitch(labelRes: StringResource, switchPref: Preference<Boolean>, content: @Composable FlowRowScope.() -> Unit) {
```
AntsyLich, line 415: This is just unnecessary

presentation-core/src/main/java/tachiyomi/presentation/core/components/SettingsItems.kt
```diff
@@ -381,6 +384,71 @@ fun TextItem(label: String, value: String, onChange: (String) -> Unit) {
...
+    BaseSettingsItem(
+        label = label,
+        widget = {
+            Switch(
+                checked = checked,
+                onCheckedChange = null,
+                modifier = Modifier.size(switchSize.dp),
+            )
+        },
+        onClick = onClick,
```
AntsyLich, line 410:
```suggestion
fun SwitchItem(label: String, preference: Preference<Boolean>) {
    val checked by preference.collectAsState()
    BaseSettingsItem(
        label = label,
        widget = {
            Switch(
                checked = checked,
                onCheckedChange = null,
            )
        },
        onClick = { preference.toggle() },
```

AntsyLich (changes requested):

core/common/src/main/kotlin/tachiyomi/core/common/util/system/ImageUtil.kt
AntsyLich, line 1: Why is the logic being removed here?

app/src/main/java/eu/kanade/presentation/reader/settings/GeneralSettingsPage.kt
```diff
@@ -20,7 +21,6 @@ private val themes = listOf(
     MR.strings.black_background to 1,
     MR.strings.gray_background to 2,
     MR.strings.white_background to 0,
-    MR.strings.automatic_background to 3,
```
AntsyLich, line 23: You need to add a migration for the settings removal

AntsyLich (changes requested):

core/common/src/main/kotlin/tachiyomi/core/common/util/system/ImageUtil.kt
01-prog: Before the logic was that if the edges were "colorful" it tries to find a suitable color for the background according to edges. If it fails to do so it will try to find if its better to have a black or white background. To do so, if the edges were mainly white (or black) then the background will be white (respectively black). However the issue was that for mangas the edges are manly white so the background would almost always fall back to white. I removed the black/white part because this is now the user's choice (whether it should be black, gray or white if no suitable color was found).

app/src/main/java/eu/kanade/presentation/reader/settings/GeneralSettingsPage.kt
01-prog: I'm very sorry but i don't know what a migration is ToT. Could you explain what i should do ?

PR #1920, opened 2025-03-25
Fix next chapter button not changing chapter

perokhe:
Adapted from TachiyomiSY

Closes #875

PR #1936, opened 2025-03-28
Fix page number not appearing when opening chapter

perokhe:
Caused by #1920


AntsyLich:
Add a sub entry to the previous changelog

PR #1957, opened 2025-04-02
Deduplicate entries when browsing

AntsyLich:


cuong-tran:
also notice this happens recently. I did the same for my related things to filter out duplicate entries.

data/src/main/java/tachiyomi/data/source/SourcePagingSource.kt
```diff
@@ -39,6 +39,8 @@ abstract class BaseSourcePagingSource(
     private val networkToLocalManga: NetworkToLocalManga = Injekt.get(),
 ) : SourcePagingSource() {
 
+    private val seenManga = hashSetOf<String>()
```
cuong-tran, line 42: Where is the setter?

AntsyLich: What setter?

cuong-tran: I don't see entries get added to 'seen' anywhere?

AntsyLich: https://github.com/mihonapp/mihon/blob/0d35b6fdafbf5451a2743ea9bcfc735bf49374a7/data/src/main/java/tachiyomi/data/source/SourcePagingSource.kt#L58

cuong-tran: my bad, glanced too quick, thought it was only filtering.

RePod:
Moved to 2176

AntsyLich:
Make a new issue instead of commenting on closed PRs

PR #1967, opened 2025-04-04
Update non-library manga data when browsing

### LINKED ISSUE #923: Prefer new data from source instead of db when browsing

AntsyLich:
Closes #923

### DIFF
--- a/data/src/main/sqldelight/tachiyomi/data/mangas.sq
+++ b/data/src/main/sqldelight/tachiyomi/data/mangas.sq
@@ -198,7 +198,16 @@ insertNetworkManga {
 
     -- Update the title if it is not favorite
     UPDATE mangas
-    SET title = :title
+    SET
+        title = CASE WHEN :updateTitle THEN :title ELSE title END,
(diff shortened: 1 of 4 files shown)


data/src/main/sqldelight/tachiyomi/data/mangas.sq
```diff
@@ -198,7 +198,16 @@ insertNetworkManga {
 
     -- Update the title if it is not favorite
```
AwkwardPeak7, line 199: Old comment

AwkwardPeak7 (commented):
I think if existing manga in db is initialised, it should be set to not initialised if it's title/thumbnail changes. As it is possible title or thumbnail on browse and manga page are different, so should get the details again

PR #1980, opened 2025-04-08
Fix: navigation issue after migrate a duplicated entry from History tab

cuong-tran:
after migrating a duplicate entry in History tab, the new Manga screen will replace the current App screen which prevent any going back.

Pushing to the new Manga screen here also doesn't seem reasonable since user might trying to favorite several entries in the list and doesn't expect to go to the Manga screen.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryTab.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/history/HistoryTab.kt
@@ -104,7 +104,7 @@ data object HistoryTab : Tab {
                         screenModel.addFavorite(dialog.manga)
                     },
                     onOpenManga = { navigator.push(MangaScreen(it.id)) },
-                    onMigrate = { screenModel.showMigrateDialog(dialog.manga, it) },
+                    onMigrate = { },
                 )
             }
             is HistoryScreenModel.Dialog.ChangeCategory -> {


CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix mark existing duplicate read chapters as read option not working in some cases ([@AntsyLich](https://github.com/AntsyLich)) ([#1944](https://github.com/mihonapp/mihon/pull/1944))
 - Fix app bar action tooltips blocking clicks ([@Bartuzen](https://github.com/Bartuzen)) ([#1928](https://github.com/mihonapp/mihon/pull/1928))
 - Fix unintended app permissions due to Firebase misconfiguration ([@AntsyLich](https://github.com/AntsyLich)) ([#1960](https://github.com/mihonapp/mihon/pull/1960))
+- Fix navigation issue after migrate a duplicated entry in History tab ([@cuong-tran](https://github.com/cuong-tran)) ([#1980](https://github.com/mihonapp/mihon/pull/1980))
```
AntsyLich, line 38:
```suggestion
- Fix navigation issue after migrating a duplicated entry from History tab ([@cuong-tran](https://github.com/cuong-tran)) ([#1980](https://github.com/mihonapp/mihon/pull/1980))
```

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreen.kt
```diff
@@ -218,7 +218,7 @@ class MangaScreen(
                     screenModel = MigrateDialogScreenModel(),
                     onDismissRequest = onDismissRequest,
                     onClickTitle = { navigator.push(MangaScreen(dialog.oldManga.id)) },
-                    onPopScreen = { navigator.replace(MangaScreen(dialog.newManga.id)) },
+                    onPopScreen = onDismissRequest,
```
AntsyLich, line 221: If the migration was a success then this should open the new entry

AntsyLich (changes requested):

AntsyLich (approved):

PR #2009, opened 2025-04-16
slider's pill wasn't using LocalTextStyle

cuong-tran:


AntsyLich:
i'm not even sure what the correct thing is supposed to be

cuong-tran:
> i'm not even sure what the correct thing is supposed to be

it was creating a local `val style = LocalTextStyle.current` right above it but didn't use it. The style is fixed to `bodyMedium` for all places `Pill` is used instead of `LocalTextStyle`

presentation-core/src/main/java/tachiyomi/presentation/core/components/Pill.kt
```diff
@@ -57,6 +57,6 @@ fun Pill(
         modifier = modifier,
         color = color,
         contentColor = contentColor,
-        style = MaterialTheme.typography.bodyMedium.copy(fontSize = fontSize),
+        style = style.copy(fontSize = fontSize),
```
AntsyLich, line 60: Make it inline

PR #2010, opened 2025-04-17
Continuous horizontal reading mode

eliasbenb:
This PR introduces a continuous horizontal reading mode by modifying.

**What's New:**

- The reading modes menu now includes a "Strip w/gaps (horizontal)" option for continuous horizontal scrolling
- In order for the existing webtoon reading modes to be more distinct with the changes, they were renamed in `strings.xml` as follows:
  - `Long strip` -> `Strip (vertical)`
  - `Long strip with gaps` -> `Strip w/gaps (vertical)`
- The Webtoon viewer now accepts an optional `isHorizontal` flag which determines the scrolling direction and fit of the images (left-to-right and fit-to-height for True and top-to-bottom and fit-to-width for False)

### Videos

https://github.com/user-attachments/assets/8195e4be-b693-4bc7-b2ab-f5223bb61c77

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
   - Use simpler markdown flavour ([@Secozzi](https://github.com/Secozzi)) ([#2000](https://github.com/mihonapp/mihon/pull/2000))
 - Add Nord Theme ([@Riztard](https://github.com/Riztard)) ([#1951](https://github.com/mihonapp/mihon/pull/1951))
 - Option to keep read manga when clearing database ([@AwkwardPeak7](https://github.com/AwkwardPeak7)) ([#1979](https://github.com/mihonapp/mihon/pull/1979))
+- Continuous horizontal reading mode ([@eliasbenb](https://github.com/eliasbenb)) ([#2010](https://github.com/mihonapp/mihon/pull/2010))
 
 ### Improved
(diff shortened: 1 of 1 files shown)


i18n/src/commonMain/moko-resources/base/strings.xml
```diff
@@ -446,8 +446,9 @@
     <string name="left_to_right_viewer">Paged (left to right)</string>
     <string name="right_to_left_viewer">Paged (right to left)</string>
     <string name="vertical_viewer">Paged (vertical)</string>
-    <string name="webtoon_viewer">Long strip</string>
-    <string name="vertical_plus_viewer">Long strip with gaps</string>
+    <string name="webtoon_viewer">Strip (vertical)</string>
+    <string name="vertical_plus_viewer">Strip w/gaps (vertical)</string>
+    <string name="horizontal_plus_viewer">Strip w/gaps (horizontal)</string>
```
MajorTanya, line 451: Why did you change these to "Strip" instead of "Long Strip" and changed "with" to "w/"? Especially the latter change could be a problem for i18n.

BrutuZ:
How do you set the direction?

i18n/src/commonMain/moko-resources/base/strings.xml
AntsyLich: I would rather have the whole sheet redesigned

BrutuZ:
> **If you're asking from a user's perspective:**
See the video I attached, there's now two options separating vertical and horizontal continuous scroll ("strip") in the reading mode menu.

That's for the orientation, I asked about direction (Left to Right or Right to Left).
Should be especially relevant since it'd affect whether 2-page spreads weld together from the center or edges if the directions match or not.

PR #2043, opened 2025-04-24
Empty browse source screen when hiding already added entried

NarwhalHorns:
currently if using the "hide entries already in library" setting, it does not show an empty screen message or anything as it relies on the paging state error being no results error which doesn't happen if we filter them out afterwards due to the setting.

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/browse/BrowseSourceScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/browse/BrowseSourceScreen.kt
@@ -73,10 +73,20 @@ fun BrowseSourceContent(
         }
     }
 
-    if (mangaList.itemCount <= 0 && errorState != null && errorState is LoadState.Error) {
+    if (mangaList.itemCount == 0 && mangaList.loadState.refresh is LoadState.Loading) {
+        LoadingScreen(
+            modifier = Modifier.padding(contentPadding),
+        )
+        return
+    }
+
+    if (mangaList.itemCount <= 0) {
         EmptyScreen(
             modifier = Modifier.padding(contentPadding),
-            message = getErrorMessage(errorState),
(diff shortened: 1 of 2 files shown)


app/src/main/java/eu/kanade/presentation/browse/BrowseSourceScreen.kt
```diff
@@ -73,10 +73,20 @@ fun BrowseSourceContent(
...
 
-    if (mangaList.itemCount <= 0 && errorState != null && errorState is LoadState.Error) {
+    if (mangaList.itemCount == 0 && mangaList.loadState.refresh is LoadState.Loading) {
+        LoadingScreen(
+            modifier = Modifier.padding(contentPadding),
+        )
+        return
+    }
+
+    if (mangaList.itemCount <= 0) {
```
AntsyLich, line 83:
```suggestion
    if (mangaList.itemCount == 0) {
```

app/src/main/java/eu/kanade/presentation/browse/BrowseSourceScreen.kt
```diff
@@ -73,10 +73,20 @@ fun BrowseSourceContent(
         }
     }
 
-    if (mangaList.itemCount <= 0 && errorState != null && errorState is LoadState.Error) {
+    if (mangaList.itemCount == 0 && mangaList.loadState.refresh is LoadState.Loading) {
+        LoadingScreen(
+            modifier = Modifier.padding(contentPadding),
+        )
```
AntsyLich, line 79:
```suggestion
        LoadingScreen(Modifier.padding(contentPadding))
```

AntsyLich (changes requested):

AntsyLich (approved):

PR #2077, opened 2025-05-07
Fix pressing `Enter` while searching also triggering navigation back on physical keyboards

### LINKED ISSUE #1131: Physical Keyboard issue
### Steps to reproduce

- open Global Search or Browse Search
- type something
- press enter to trigger search

### Expected behavior

search

### Actual behavior

automatically selects the back button and exits out of the screen (even tho search was successfully triggered)

### Crash logs

_No response_

### Mihon version

r6808 and 0.16.5

### Android version

Android 11

### Device

Waydroid

### Other details

- this is possibly an issue with compose, since it works as expected on non-compose forks (like yokai)
- not dependent on waydroid, as can be reproduced with actual phone and keyboard
- really painful during ext development/testing 😔

### Acknowledgements

AwkwardPeak7:
On a physical/hardware keyboard, a keystroke sends two events. One of key press and one on key release, so effectively two events on a single keystroke. ([ref](https://developer.android.com/develop/ui/compose/touch-input/keyboard-input/commands#key_events))
This caused `searchAndClearFocus` in `SearchToolbar` composable to trigger twice, second time tapping the back button and exiting out of search.
To fix this, I have changed it to only trigger on Key release, which is the behavior of software keyboard when tapping the enter/search key.

Also changed it so focus is moved next, and not to the back button

Note: only tested on Waydroid, I don't know if this will have any other side effects on other devices with keyboards

closes #1131

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix empty layout not appearing in browse source screen in some cases ([@NarwhalHorns](https://github.com/NarwhalHorns)) ([#2043](https://github.com/mihonapp/mihon/pull/2043))
 - Fix Pill not following the local text style ([@AntsyLich](https://github.com/AntsyLich)) ([`f8cb506`](https://github.com/mihonapp/mihon/commit/f8cb506))
 - Fix downloader stopping after failing to create download directory of a manga ([@AntsyLich](https://github.com/AntsyLich)) ([#2068](https://github.com/mihonapp/mihon/pull/2068))
+- Fix enter key exiting out of search on physical keyboards ([@AwkwardPeak7](https://github.com/AwkwardPeak7)) ([#2077](https://github.com/mihonapp/mihon/pull/2077))
 
 ### Removed
 - Remove Okhttp networking from WebView Screen ([@AwkwardPeak7](https://github.com/AwkwardPeak7)) ([#2020](https://github.com/mihonapp/mihon/pull/2020))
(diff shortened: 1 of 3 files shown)


presentation-core/src/main/java/tachiyomi/presentation/core/util/Modifier.kt
```diff
@@ -53,8 +55,17 @@ fun Modifier.clickableNoIndication(
...
     when (it.key) {
         Key.Enter, Key.NumPadEnter -> {
-            action()
-            true
+            // Physical keyboards generate two event types:
+            // - KeyDown when the key is pressed
+            // - KeyUp when the key is released
+            // To match the behavior of software keyboards (IME),
+            // trigger [action] only on the KeyUp event.
+            if (it.type == KeyEventType.KeyUp) {
```
AntsyLich, line 63: But we're working with physical keyboard here where stuff usually triggers on `KeyDown`

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix empty layout not appearing in browse source screen in some cases ([@NarwhalHorns](https://github.com/NarwhalHorns)) ([#2043](https://github.com/mihonapp/mihon/pull/2043))
 - Fix Pill not following the local text style ([@AntsyLich](https://github.com/AntsyLich)) ([`f8cb506`](https://github.com/mihonapp/mihon/commit/f8cb506))
 - Fix downloader stopping after failing to create download directory of a manga ([@AntsyLich](https://github.com/AntsyLich)) ([#2068](https://github.com/mihonapp/mihon/pull/2068))
+- Fix enter key exiting out of search on physical keyboards ([@AwkwardPeak7](https://github.com/AwkwardPeak7)) ([#2077](https://github.com/mihonapp/mihon/pull/2077))
```
AntsyLich, line 54:
```suggestion
- Fix pressing `Enter` while searching also triggering navigation back on physical keyboards ([@AwkwardPeak7](https://github.com/AwkwardPeak7)) ([#2077](https://github.com/mihonapp/mihon/pull/2077))
```

AntsyLich (changes requested):

presentation-core/src/main/java/tachiyomi/presentation/core/util/Modifier.kt
AwkwardPeak7: shouldn't the behavior be uniform across software and physical keyboard?

AntsyLich (approved):

PR #2084, opened 2025-05-09
Add category library filter

LetrixZ:
This adds a new category filter for the library. It allows you to select categories that the library item must have and/or which it shouldn't have.

filter


fivemserverhostings:
Very clean code with good test coverage. This is a great example of best practices.

Issue #1583, opened 2024-12-17
add option to turn off "pull down" = update (refresh) current category


flfwpl:
### Describe your suggested feature

PullRefresh give it an option to turn on or off

### Other details

the cause thing like to get banned.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.17.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

amayra:
i agree with you 

i would like to control this feature

PR #2105, opened 2025-05-14
feat/Pull to Refresh toggle

brahimABD98:
This PR addresses  #1583  by introducing a **"Pull to Refresh" toggle** in the Library tab settings. When disabled, users are shown a snackbar instead of triggering a refresh.
Changes

- Added `enablePullToRefresh` preference in `LibraryPreferences`
- Hooked the preference into `LibraryTab` and `LibraryContent`
- Exposed this setting in `SettingsLibraryScreen` under the Library settings section
- Updated `LibraryContent` to:
  - Respect `enablePullToRefresh` state
  - Show a snackbar when pull-to-refresh is disabled
  - Fix refresh spinner getting stuck when disabled

### Images
| Image 1 | Image 2 |
| ------- | ------- |
| image | image |

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/library/components/LibraryContent.kt
+++ b/app/src/main/java/eu/kanade/presentation/library/components/LibraryContent.kt
@@ -6,6 +6,8 @@ import androidx.compose.foundation.layout.calculateEndPadding
 import androidx.compose.foundation.layout.calculateStartPadding
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.pager.rememberPagerState
+import androidx.compose.material3.SnackbarHost
+import androidx.compose.material3.SnackbarHostState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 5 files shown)


app/src/main/java/eu/kanade/presentation/library/components/LibraryContent.kt
```diff
@@ -84,6 +88,18 @@ fun LibraryContent(
         PullRefresh(
             refreshing = isRefreshing,
             onRefresh = {
+                if (!enablePullToRefresh) {
+
+                    scope.launch {
+                        isRefreshing = true
+                        snackbarHostState.showSnackbar(
+                            message = "pull to update is disabled",
```
MajorTanya, line 96: Please use proper i18n

i18n/src/commonMain/moko-resources/base/strings.xml
MajorTanya, line 1: What are all these unnecessary changes in this file about? Do not arbitrarily format files.

PR #2108, opened 2025-05-14
- Add setting to hide missing chapter counts

### LINKED ISSUE #50: Setting to hide "Missing [x] chapters"
I like to hide all unread chapters of a series I'm reading but now there's this "Missing [x] chapters" red error message that I can't get rid of without showing all of the unread chapters again. Could this either get removed when hiding unread chapters or add a toggle to hide it in the settings?
Screenshot_20240116_040021_Mihon

### Other details

_No response_

### Acknowledgements

- [X] If this is an issue with an official extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/extensions/issues/new/choose).
- [X] I have updated the app to version **[0.16.0](https://github.com/mihonapp/mihon/releases/latest)**.

User826:
Added a simple toggle to hide/show missing chapter counts. Closes #50


AntsyLich:
Instead of hiding in the UI just don't include them at all

User826:
So get rid of the missing chapter indicators? Just wanted to double check.

AntsyLich:
If the setting is on don't include them at all instead of hiding them on UI side

User826:
Missing chapter counts should now be excluded

app/src/main/java/eu/kanade/tachiyomi/ui/manga/MangaScreenModel.kt
```diff
@@ -1136,29 +1138,33 @@ class MangaScreenModel(
...
-                    val (lowerChapter, higherChapter) = if (manga.sortDescending()) {
-                        after to before
-                    } else {
-                        before to after
-                    }
-                    if (higherChapter == null) return@insertSeparators null
+                if (hideMissingChapters) {
+                    processedChapters
+                } else {
+                    processedChapters.insertSeparators { before, after ->
```
AntsyLich:
```suggestion
                if (hideMissingChapters) {
                    return@lazy processedChapters
                }
                
                processedChapters.insertSeparators { before, after ->
```

i18n/src/commonMain/moko-resources/base/strings.xml
```diff
@@ -316,6 +316,9 @@
     <string name="pref_mark_duplicate_read_chapter_read_existing">After reading a chapter</string>
     <string name="pref_mark_duplicate_read_chapter_read_new">After fetching new chapter</string>
 
+    <string name="pref_hide_missing_chapter_indicators">Hide missing chapter indicators</string>
+    <string name="pref_hide_missing_chapter_indicators_summary">Prevent showing messages like 'Missing 3 chapters'</string>
```
AntsyLich: Don't think this needs a subtitle

AntsyLich (changes requested):

AntsyLich (approved):

PR #2200, opened 2025-06-13
fix CloudflareInterceptor not waiting for the challenge

AwkwardPeak7:
Two changes:
1)

[`onReceivedError`](https://developer.android.com/reference/android/webkit/WebViewClient#onReceivedError(android.webkit.WebView,%20android.webkit.WebResourceRequest,%20android.webkit.WebResourceError)) doesn't give http error codes in the `error` parameter. It is instead some internal error code.
<details>
<summary>Ref</Summary>

https://developer.android.com/reference/android/webkit/WebResourceError#getErrorCode()
> Gets the error code of the error. The code corresponds to one of the ERROR_* constants in [WebViewClient](https://developer.android.com/reference/android/webkit/WebViewClient).
---
image

</details>

This means, `onPageFinish` will immediately call `latch.countDown()` (in last if block) as challenge was seemingly not found, without giving it some time.

`onReceivedHttpError` is the method which actually receives http errors, so detecting the challenge there correctly makes it wait for webview to attempt to solve the challenge

2\)
Only launch webview if there is a possible challenge, not on geo block or other restrictions by cloudflare

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix Pill not following the local text style ([@AntsyLich](https://github.com/AntsyLich)) ([`f8cb506`](https://github.com/mihonapp/mihon/commit/f8cb506))
 - Fix downloader stopping after failing to create download directory of a manga ([@AntsyLich](https://github.com/AntsyLich)) ([#2068](https://github.com/mihonapp/mihon/pull/2068))
 - Fix pressing `Enter` while searching also triggering navigation back on physical keyboards ([@AwkwardPeak7](https://github.com/AwkwardPeak7)) ([#2077](https://github.com/mihonapp/mihon/pull/2077))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -56,6 +56,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix Pill not following the local text style ([@AntsyLich](https://github.com/AntsyLich)) ([`f8cb506`](https://github.com/mihonapp/mihon/commit/f8cb506))
 - Fix downloader stopping after failing to create download directory of a manga ([@AntsyLich](https://github.com/AntsyLich)) ([#2068](https://github.com/mihonapp/mihon/pull/2068))
 - Fix pressing `Enter` while searching also triggering navigation back on physical keyboards ([@AwkwardPeak7](https://github.com/AwkwardPeak7)) ([#2077](https://github.com/mihonapp/mihon/pull/2077))
+- Fix CloudflareInterceptor not waiting for the challenge to complete ([@AwkwardPeak7](https://github.com/AwkwardPeak7)) ([#2200](https://github.com/mihonapp/mihon/pull/2200))
```
```suggestion
- Ensure app waits for Cloudflare challenge to complete before continuing ([@AwkwardPeak7](https://github.com/AwkwardPeak7)) ([#2200](https://github.com/mihonapp/mihon/pull/2200))
```

PR #2249, opened 2025-06-26
Add option to download all chapters not yet downloaded

mawmao:
This adds a new **"Not Downloaded"** option to the download menu on the manga screens. Basically, it lets you queue up all the chapters that aren’t downloaded yet  (regardless of read status). If they’re already downloaded or already in the queue, it just skips them. 

Honestly, I made this mostly for myself because I wanted to download all  the chapters, even the ones I’d already read. 

It’s based off the existing `downloadUnreadChapters()` function, but I tweaked it to include all chapters instead of just the unread ones.


Smol-Ame:
This was in Tachiyomi before they intentionally removed it sometime later since it gave users easy access to DDoS sources with requests.

Consider downloading with a conscience instead, since mass downloading (or even mass updating) could lead to sources blocking you/users, sources breaking your extensions, and/or go after whoever/wherever you get your extensions from/the app itself.

mawmao:
Thanks for the heads-up. I wasn’t aware of that. 😅 
I mainly added it for personal use so I might just keep it in my fork. Totally understand the concerns though.

PR #2266, opened 2025-07-05
Fix Unbounded Memory Growth by Disabling In-Memory Cover Caching

Lolle2000la:
This PR addresses a memory leak caused by the in-memory caching of cover images, which leads to performance degradation and eventual app crashes.

While this solution resolves the critical memory issue, it does so by disabling in-memory caching for cover images through `coil3` entirely. I'm open to feedback and alternative approaches.

## The Problem

When browsing a large library or scrolling through a source's feed, the memory allocated for cover images grows without bounds. This memory is not being reclaimed, eventually exhausting the available memory for the app. When the JVM fails its next allocation attempt, the app crashes.

This issue is most noticeable with large libraries that populate pages like the library, history, and updates tabs with many cover images. However, it also occurs when scrolling through any source for an extended period. The memory usage continuously builds up until the app is force-stopped.

On a memory-constrained Android emulator (~1.5GB available), this limit can be reached in under two minutes by scrolling through a source with many entries.

## The Fix

I have explicitly configured AsyncImage to disable in-memory caching for cover images. This ensures that memory allocated to an image is freed as soon as it is no longer on screen.

Disabling the cache immediately resolves the memory buildup and prevents the related crashes.

## Notable Consequences

With this change, previously viewed images need to be reloaded as soon as they scroll out of and back into the viewport. While this is the intended effect of the fix, it may not be the ideal user experience. During testing on an emulator, this did not introduce any noticeable additional lag.

## Method of diagnosis

1.  I initially used the memory profiler and observed a significant buildup of String objects, which I attributed to database queries.
2. To reproduce the crash, I started scrolling through a source and monitored memory usage. I noticed that native memory, in particular, was increasing rapidly. My initial attempts to fix this by ensuring streams were properly closed proved ineffective.
3. I then investigated coil3's caching mechanism. After first trying to enable the disk cache (which had no effect), I disabled the in-memory cache for the MangaCover.kt component.
4. Disabling the cache completely resolved the memory leak, preventing the app from crashing under conditions where it previously would.

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/manga/components/MangaCover.kt
+++ b/app/src/main/java/eu/kanade/presentation/manga/components/MangaCover.kt
@@ -31,8 +31,8 @@ enum class MangaCover(val ratio: Float) {
         AsyncImage(
             model = coil3.request.ImageRequest.Builder(androidx.compose.ui.platform.LocalContext.current)
                 .data(data)
-                .memoryCachePolicy(coil3.request.CachePolicy.ENABLED)
-                .diskCachePolicy(coil3.request.CachePolicy.ENABLED)
+                .memoryCachePolicy(coil3.request.CachePolicy.DISABLED)
+                .diskCachePolicy(coil3.request.CachePolicy.DISABLED)
                 .build(),
             placeholder = ColorPainter(CoverPlaceholderColor),
             error = rememberResourceBitmapPainter(id = R.drawable.cover_error),
--- a/app/src/main/java/eu/kanade/presentation/manga/components/MangaCover.kt
+++ b/app/src/main/java/eu/kanade/presentation/manga/components/MangaCover.kt
@@ -29,7 +29,11 @@ enum class MangaCover(val ratio: Float) {
         onClick: (() -> Unit)? = null,
     ) {
         AsyncImage(
-            model = data,
+            model = coil3.request.ImageRequest.Builder(androidx.compose.ui.platform.LocalContext.current)
+                .data(data)
+                .memoryCachePolicy(coil3.request.CachePolicy.ENABLED)
+                .diskCachePolicy(coil3.request.CachePolicy.ENABLED)
+                .build(),
             placeholder = ColorPainter(CoverPlaceholderColor),
             error = rememberResourceBitmapPainter(id = R.drawable.cover_error),
             contentDescription = contentDescription,
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
@@ -1,7 +1,6 @@
 package eu.kanade.tachiyomi.data.download
 
 import android.content.Context
-import com.hippo.unifile.UniFile
 import eu.kanade.tachiyomi.data.download.model.Download
 import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.source.model.Page
@@ -162,7 +161,7 @@ class DownloadManager(
     fun buildPageList(source: Source, manga: Manga, chapter: Chapter): List<Page> {
         val chapterDir = provider.findChapterDir(chapter.name, chapter.scanlator, manga.title, source)
         val files = chapterDir?.listFiles().orEmpty()
-            .filter { it.isFile && isImageFile(it) }
+            .filter { it.isFile && ImageUtil.isImage(it.name) { it.openInputStream() } }
 
         if (files.isEmpty()) {
             throw Exception(context.stringResource(MR.strings.page_list_empty_error))
@@ -172,16 +171,6 @@ class DownloadManager(
             .mapIndexed { i, file ->
                 Page(i, uri = file.uri).apply { status = Page.State.Ready }
             }
-    }
-
-    private fun isImageFile(file: UniFile): Boolean {
-        return try {
-            file.openInputStream().use { stream ->
-                ImageUtil.isImage(file.name) { stream }
-            }
-        } catch (e: Exception) {
-            false
-        }
     }
 
     /**
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadManager.kt
@@ -1,6 +1,7 @@
 package eu.kanade.tachiyomi.data.download
 
 import android.content.Context
+import com.hippo.unifile.UniFile
 import eu.kanade.tachiyomi.data.download.model.Download
 import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.source.model.Page
@@ -161,7 +162,7 @@ class DownloadManager(
     fun buildPageList(source: Source, manga: Manga, chapter: Chapter): List<Page> {
         val chapterDir = provider.findChapterDir(chapter.name, chapter.scanlator, manga.title, source)
         val files = chapterDir?.listFiles().orEmpty()
-            .filter { it.isFile && ImageUtil.isImage(it.name) { it.openInputStream() } }
+            .filter { it.isFile && isImageFile(it) }
 
         if (files.isEmpty()) {
             throw Exception(context.stringResource(MR.strings.page_list_empty_error))
@@ -171,6 +172,16 @@ class DownloadManager(
             .mapIndexed { i, file ->
                 Page(i, uri = file.uri).apply { status = Page.State.Ready }
             }
+    }
+
+    private fun isImageFile(file: UniFile): Boolean {
+        return try {
+            file.openInputStream().use { stream ->
+                ImageUtil.isImage(file.name) { stream }
+            }
+        } catch (e: Exception) {
+            false
+        }
     }
 
     /**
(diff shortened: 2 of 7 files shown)


MajorTanya:
If I understand this correctly, your changes deactivate caching of covers _entirely_, right? Does that mean Coil will fetch the images from remote every single time?

Lolle2000la:
Yes, exactly, at least from `coil3`'s side when calling fetching the image. If the image comes from disk or some other caching not managed by coil3, it should still not be a problem, but if passed the remote resource directly it will also fetch it directly.

I think this might be helped a bit by at least enabling disk-caching again (I turned off all caching). I am not deep into Android and do not know if perhaps Coil can simply be configured to not eat up all memory.

MajorTanya:
> Yes, exactly, at least from `coil3`'s side when calling fetching the image. If the image comes from disk or some other caching not managed by coil3, it should still not be a problem, but if passed the remote resource directly it will also fetch it directly.
> 
> I think this might be helped a bit by at least enabling disk-caching again (I turned off all caching). I am not deep into Android and do not know if perhaps Coil can simply be configured to not eat up all memory.

Re-requesting the cover all the time is a non-starter for Mihon. This would cause all users to request all covers again all the time. Considering the number of users and the average number of titles in a library, that seems like a really, _really_ bad idea.

Lolle2000la:
That is why I asked for feedback on better methods to solve this. I absolutely get that this is a nonstarter, but it did fix the issue so I hoped that people would look at that and know from their own experiences what the root problem might be.

... I should have probably formulated the description better to make that clear, sorry!

I looked a bit into the caching and the `ImageLoader`-singleton and see that the cache is currently not configured in App.kt. I have checked [the Coil-docs](https://coil-kt.github.io/coil/image_loaders/) and it seems that you can limit the caching to a percentage. I have taken the values from the docs and undid the change to MangaCover.

This fixes it as well (no crashing and no explosive growth), but keeps caching (just not without limits). Note that I also configured disk-caching in the app cache dir (because its in the docs example configuration). If that is unwanted I can remove that part of the config again.

nubesurrealista:
I implemented this feature in my build and it works well. On low-end devices, this should be a must.

It has improved the app's memory consumption and overall it feels better. I understand that loading the covers can be a bit annoying, but hey, that could be improved.

app/src/main/java/eu/kanade/tachiyomi/App.kt
```diff
@@ -196,6 +199,19 @@ class App : Application(), DefaultLifecycleObserver, SingletonImageLoader.Factor
                 add(MangaKeyer())
             }
 
+            memoryCache(
+                MemoryCache.Builder()
+                    .maxSizePercent(context, 0.25)
```
AntsyLich:
```suggestion
                    .maxSizePercent(context)
```
1. https://coil-kt.github.io/coil/api/coil-core/coil3.memory/-memory-cache/-builder/index.html#-1971330947%2FFunctions%2F1448608982
2. https://github.com/coil-kt/coil/blob/46130609cc03bd5313bd7d902552e372225c3783/coil-core/src/androidMain/kotlin/coil3/util/contexts.kt#L30-L37

app/src/main/java/eu/kanade/tachiyomi/App.kt
```diff
@@ -196,6 +199,19 @@ class App : Application(), DefaultLifecycleObserver, SingletonImageLoader.Factor
...
+                    .build(),
+            )
+
+            diskCache(
+                DiskCache.Builder()
+                    .directory(context.cacheDir.resolve("image_cache"))
+                    .maxSizePercent(0.02)
+                    .build(),
+            )
```
AntsyLich:
```suggestion
```
It's the same as default https://github.com/coil-kt/coil/blob/46130609cc03bd5313bd7d902552e372225c3783/coil-core/src/commonMain/kotlin/coil3/disk/DiskCache.kt#L115

AntsyLich:
Also add a changelog

app/src/main/java/eu/kanade/tachiyomi/App.kt
Lolle2000la: Ok, went with that value. The fraction is slightly smaller than the current value and does get even smaller for very constrained devices.

app/src/main/java/eu/kanade/tachiyomi/App.kt
Lolle2000la: Removed it as requested

Lolle2000la:
> Also add a changelog

Did that, worked in your suggestions and also rebased just for good measure.

app/src/main/java/eu/kanade/tachiyomi/App.kt
```diff
@@ -196,6 +199,12 @@ class App : Application(), DefaultLifecycleObserver, SingletonImageLoader.Factor
                 add(MangaKeyer())
             }
 
+            memoryCache(
+                MemoryCache.Builder()
+                    .maxSizePercent(context)
+                    .build(),
+            )
```
AntsyLich, line 204: This also seems to be the default https://github.com/coil-kt/coil/blob/46130609cc03bd5313bd7d902552e372225c3783/coil-core/src/commonMain/kotlin/coil3/ImageLoader.kt#L352-L356. How is this improving anything 🤔

Lolle2000la: I'm on mobile, struggling to read on my small display so apologies if I misread, but I think that is not the code being called. You linked the method to build a fully configured loader but the existing code creates a new one from scratch (I don't know why, might be a reason for that), so it also doesn't inherit the defaults.

Lolle2000la: Ok, I looked it again and I think you are right? Unless the `build` on line 216 resolves to another method, I don't know why this should change anything.

I do wonder why this does have an impact then. In the Komikku PR I have [recorded the effect it had on memory usage and how the native memory explodes](https://github.com/komikku-app/komikku/pull/1031#issuecomment-3275707919) without the change.

My guess: The backing `memoryCacheLazy` is somehow set somewhere else and this now sets it back to a sensible default. In that case I would be treating the symptoms. Otherwise I don't know, it just changes it I guess. 

I tried tracing the flow of the values, looked a bit around the source you referenced, gave up, asked gemini and got some "I don't know, timing issues, lazy init bugs, inherited app context setting wrong, a bug in coil, but sure bro 👍 " response back (of course in many more words), so...

Anyway, to advocate for my code, this is absolutely harmless to include and somehow fixes the bug 😅 .

app/src/main/java/eu/kanade/tachiyomi/App.kt
```diff
@@ -17,6 +17,9 @@ import androidx.lifecycle.ProcessLifecycleOwner
 import androidx.lifecycle.lifecycleScope
 import coil3.ImageLoader
 import coil3.SingletonImageLoader
+import coil3.disk.DiskCache
+import coil3.disk.directory
```
AntsyLich:
```suggestion
```

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix reader tap zones triggering after scrolling is stopped by tapping ([@NGB-Was-Taken](https://github.com/NGB-Was-Taken)) ([#2680](https://github.com/mihonapp/mihon/pull/2680))
 - Fix shizuku installer not updating installed extensions ([@NGB-Was-Taken](https://github.com/NGB-Was-Taken)) ([#2697](https://github.com/mihonapp/mihon/pull/2697))
 - Fix mass migration not using the same search queries as individual migration ([@AntsyLich](https://github.com/AntsyLich)) ([#2736](https://github.com/mihonapp/mihon/pull/2736))
+- Fix unbounded memory growth by limiting in-memory cover caching ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2266](https://github.com/mihonapp/mihon/pull/2266))
```
```suggestion
- Reduce memory usage by reducing in-memory cover cache size ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2266](https://github.com/mihonapp/mihon/pull/2266))
```
Also move it under Improved (which would be above changed)

AntsyLich (changes requested):
I'll merge this but If possible can you make an issue on coil github with the finding?

CHANGELOG.md
Lolle2000la: Applied this in [bc6f634](https://github.com/mihonapp/mihon/pull/2266/commits/bc6f63454d824323186a9f9866651f6fbae274c5)

Lolle2000la:
> I'll merge this but If possible can you make an issue on coil github with the finding?

Definitely sounds like a good idea, though I might not be able to spend much time on that until Christmas holidays. But I'm quite interested in what will turn out to be the reason for this behavior.

Lolle2000la:
> I'll merge this but If possible can you make an issue on coil github with the finding?

Just a quick update, I didn't forget this but haven't yet managed to find the time and energy to make an issue on the coil repo. I still plan to do it, just wanted to give a heads up to everyone here. If someone else wants to report it instead of me though, I would welcome it.

Currently, I'm quite mentally exhausted by upcoming exams, so I will deal with those first. I apologize in advance for the delay.

PR #2298, opened 2025-07-22
Remove null check for active category

cuong-tran:
The null check added in #9419 (6bdc1b676e5ca4c3c8255ced835aa7d6bf7999d5) is redundant after commit 405a75438a61770a12292cc87c0fa36e16668124


AntsyLich:
Per category sort is still a thing so it's not redundant

Issue #94, opened 2024-01-17
Limit amount of author names shown in library entries. 


THElegend5:
### Describe your suggested feature

Basically so that the below doesn't happen.
https://private-user-images.githubusercontent.com/63903377/257935542-0c47720d-c862-409f-9d5b-db2488c0616d.jpg?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MDU1MTk3ODEsIm5iZiI6MTcwNTUxOTQ4MSwicGF0aCI6Ii82MzkwMzM3Ny8yNTc5MzU1NDItMGM0NzcyMGQtYzg2Mi00MDlmLTlkNWItZGIyNDg4YzA2MTZkLmpwZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAxMTclMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMTE3VDE5MjQ0MVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTcwYTk0YzhjNzQ4YTFmOTZhNmJjN2ZiMWZiNTU2NzBhZjk1ZGVkYTNjMzM4ZTgxNDdlZDVhOTk2NDZjOWZkZWQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.7QmFAs2DNTnA-lHGmmA4BfcI2nuKUd6cNxhXAK10Xk0

### Other details

This is kinda an edge case but I have quite a few that take up half the screen.
I was thinking maybe something like three dots at the end of author names which can be used to expand them back could be added so that they are still accessible ?

Reference: https://github.com/tachiyomiorg/tachiyomi/issues/9794

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
[...]

PR #2299, opened 2025-07-22
Limit author and artist name lines to 1 with ellipsis (#94)

Jh0nZ:
### Summary

This pull request addresses #94  by limiting the number of lines shown for the author and artist fields in the manga info header. Long author or artist names will now be truncated with an ellipsis (`...`) to avoid overflowing and disrupting the layout.

### What was changed

In `MangaInfoHeader.kt`, I added:

```kotlin
maxLines = 1,
overflow = TextOverflow.Ellipsis,
```
to the Text components for both author and artist.

### Images

| Before | After |
|--------|-------|
| Before | After |

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/manga/components/MangaInfoHeader.kt
+++ b/app/src/main/java/eu/kanade/presentation/manga/components/MangaInfoHeader.kt
@@ -453,6 +453,8 @@ private fun ColumnScope.MangaContentInfo(
             text = author?.takeIf { it.isNotBlank() }
                 ?: stringResource(MR.strings.unknown_author),
             style = MaterialTheme.typography.titleSmall,
+            maxLines = 1,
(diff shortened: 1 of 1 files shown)


AntsyLich:
Make it expandable

BrutuZ:
I would've preferred if it were shortened like that by default, but you could expand it back with an interaction.

jh0nz:
> Make it expandable

Author and artist fields already have actions: tap opens a search, and long press copies to clipboard — so making them expandable would conflict with those. Let me know if an alternative UI is preferred.

BrutuZ:
Can't the action be conditional to the current state? Expand when collapsed, search when expanded

jh0nz:
@BrutuZ @AntsyLich ready for review — added expand/collapse and long-press menu to title/author/artist.

PR #2329, opened 2025-07-31
Optimize and cleanup library code

AntsyLich:
Should better handle big user libraries


app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
```diff
@@ -243,13 +255,31 @@ class LibraryScreenModel(
                 filterFnIntervalCustom(it) &&
                 filterFnTracking(it)
         }
+    }
 
-        return mapValues { (_, value) -> value.fastFilter(filterFn) }
+    private fun List<LibraryItem>.applyGrouping(
+        categories: List<Category>,
+    ): Map<Category, List</* LibraryItem */ Long>> {
+        val groupCache = mutableMapOf</* Category */ Long, MutableList</* LibraryItem */ Long>>()
```
MajorTanya, line 260: Any reason this isn't a typealias or something?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
```diff
@@ -762,19 +794,28 @@ class LibraryScreenModel(
...
             val count = when {
                 !showMangaCount -> null
-                !showCategoryTabs -> getMangaCountForCategory(category)
+                !showCategoryTabs -> getItemCountForCategory(category)
                 // Whole library count
-                else -> libraryCount
+                else -> favorites.size
             }
-
+            Triple(1, 2, 3)
```
MajorTanya: This has no effect here does it?

data/src/main/java/tachiyomi/data/manga/MangaMapper.kt
```diff
@@ -121,7 +121,7 @@ object MangaMapper {
             isSyncing,
             notes,
         ),
-        category = category,
+        categories = categories.split(",").map { it.toLong() },
```
MajorTanya, line 124: Is there really no better way than string concat of category IDs?

MajorTanya:
Why the renaming from "categories" to "groups" in some but not all places? Is that relevant to the main changes made?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
AntsyLich: No

data/src/main/java/tachiyomi/data/manga/MangaMapper.kt
AntsyLich: Not that I found any

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
AntsyLich: I kinda like it this way

AntsyLich:
> Why the renaming from "categories" to "groups" in some but not all places? Is that relevant to the main changes made?

I'll change all of them to groups in a follow up PR

MajorTanya:
> > Why the renaming from "categories" to "groups" in some but not all places? Is that relevant to the main changes made?
> 
> I'll change all of them to groups in a follow up PR

Could you keep that name change to that follow-up PR then?

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt
```diff
@@ -98,32 +87,52 @@ class LibraryScreenModel(
...
+                            loggedInTrackerIds = loggedInTrackerIds,
+                        )
+                    }
+                }
+        }
+
+        screenModelScope.launchIO {
+            state
+                .dropWhile { !it.isInitialized }
+                .map { Quadruple(it.categories, it.favoritesById, it.tracksMap, it.loggedInTrackerIds) }
```
Syer10: At this point make a data class

brahmkshatriya: I recommend storing this data class directly as a variable in the `State` class instead of mapping it everytime
so converting these 4 variables into variable something called `libraryStateData`
https://github.com/AntsyLich/mihon/blob/chore/library-flatten/app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryScreenModel.kt#L725-L728

Syer10 (approved):

Issue #2347, opened 2025-08-04
Ul issue: Manga titles invisible over dark covers


H1ghSyst3m:
### Steps to reproduce

1. Open Mihon version 0.19.0
2. Go to "Migrate" > "Mass Migration"
3. View the list of manga shown for migration
4. Observe the manga titles displayed over the cover images

### Expected behavior

The manga titles should be clearly visible and readable during mass migration, regardless of the cover image’s colors.

### Actual behavior

The manga titles are shown in black text directly over the cover images.
If the cover is dark (which many are), the black text becomes very hard or impossible to read.

![Image](https://github.com/user-attachments/assets/bef20c5e-0402-4e57-8031-8174d298626a)

### Mihon version

0.19.0

### Android version

Android 15

### Device

Xiaomi 14

### Other details

Suggestion: Add a text shadow, outline, or a background behind the title text to improve contrast

Alternatively, display the title below the cover instead of over it

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [x] I have updated the app to version **[0.19.0](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I have filled out all of the requested information in this form, including specific version numbers.
- [x] I understand that **Mihon does not have or fix any extensions**, and I **will not receive help** for any issues related to sources or extensions.

AntsyLich:
Include a screenshot maybe?

H1ghSyst3m:
> Include a screenshot maybe?

Thx for Info, i thought I did add.

PR #2365, opened 2025-08-06
fix black titles in mass migration

H1ghSyst3m:
This fixes the problem in #2347

added color white to the text

### DIFF
--- a/app/src/main/java/mihon/feature/migration/list/MigrationListScreenContent.kt
+++ b/app/src/main/java/mihon/feature/migration/list/MigrationListScreenContent.kt
@@ -216,6 +216,7 @@ fun MigrationListItem(
                 maxLines = 2,
                 style = MaterialTheme.typography.labelMedium
                     .merge(shadow = Shadow(color = Color.Black, blurRadius = 4f)),
+                color = Color.White,
             )
             BadgeGroup(modifier = Modifier.padding(4.dp)) {
                 Badge(text = "$chapterCount")


MajorTanya:
Doesn't this just cause the opposite problem? Making titles unreadable on white/bright covers and the "no cover" image?

H1ghSyst3m:
> Doesn't this just cause the opposite problem? Making titles unreadable on white/bright covers and the "no cover" image?

The titles already have a shadow background, like in the library, so the titles should also be readable if there is no or white color

MajorTanya:
> > Doesn't this just cause the opposite problem? Making titles unreadable on white/bright covers and the "no cover" image?
> 
> The titles already have a shadow background, like in the library, so the titles should also be readable if there is no or white color

Please provide some screenshots. Specifically a dark cover, a normal cover, a bright/white cover, and the "no cover" image so that we can see the impact this change will have.

Ideally show each version in Light and Dark themes as well.

H1ghSyst3m:
> > > Doesn't this just cause the opposite problem? Making titles unreadable on white/bright covers and the "no cover" image?
> > 
> > 
> > The titles already have a shadow background, like in the library, so the titles should also be readable if there is no or white color
> 
> Please provide some screenshots. Specifically a dark cover, a normal cover, a bright/white cover, and the "no cover" image so that we can see the impact this change will have.
> 
> Ideally show each version in Light and Dark themes as well.

You were faster than me xD.

I saw your message right after I added the screenshot xD

AntsyLich:
Thanks for the PR but I went with another approach #2370

PR #2370, opened 2025-08-07
Fix title text color in light mode on mass migration list

AntsyLich:
Fixes #2347

PR #2375, opened 2025-08-07
change library title text and background based on theme

H1ghSyst3m:
Same change like in the #2370, so the title and the background is based on the theme.

Somehow I like the colors in the migration list

So I also changed it in the library to see what it looks like there.
I like it that way, but many people are already used to the style that is right now.

That is the result:

2025-08-07 23_08_33-Mihon – CommonMangaItem kt  Mihon app main
2025-08-07 23_08_49-Mihon – CommonMangaItem kt  Mihon app main

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/library/components/CommonMangaItem.kt
+++ b/app/src/main/java/eu/kanade/presentation/library/components/CommonMangaItem.kt
@@ -32,7 +32,6 @@ import androidx.compose.ui.draw.clip
 import androidx.compose.ui.draw.drawBehind
 import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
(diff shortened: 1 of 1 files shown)


MajorTanya:
Can you add "before" screenshots for each as well to illustrate the change?

H1ghSyst3m:
> Can you add "before" screenshots for each as well to illustrate the change?

It's just a small change that is relevant to the light version.

| Before | Now |
|--------|-----|
|  |  |


| Before | Now |
|--------|-----|
|  |  |

MajorTanya:
Not sure if I personally am a fan of the covers looking like they're vanishing into the aether at the bottom but that's just me. Can you add the dark theme comparisons anyway for completeness? You can just edit the previous comment 👍

AntsyLich:
No

PR #2382, opened 2025-08-09
Fix info section height on descriptions with images

Secozzi:
Descriptions with small images would take up too much space, and descriptions with large images would take up too little. Now it adjusts the size based on the size of the images.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - `Other` - for technical stuff.
 
 ## [Unreleased]
+### Fixes
+- Fix height of description not being calculated correctly if images are present ([@Secozzi](https://github.com/Secozzi)) ([#2382](https://github.com/mihonapp/mihon/pull/2382))
 
 ## [v0.19.1] - 2025-08-07
 ### Changed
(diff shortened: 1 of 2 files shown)


MajorTanya:
Do you have a screenshot or something? I'm not sure what you're referring to

Secozzi:
> Do you have a screenshot or something? I'm not sure what you're referring to

Before:
Screenshot_20250809-190146_Mihon

After:
Screenshot_20250809-190123_Mihon

anirudhsnayak:
I think that this commit may have introduced a small bug when scrolling upward while the description is open.
| Original | Reverted |
|-------------|-------------|
| <video src="https://github.com/user-attachments/assets/dbaccdb6-a890-4039-89aa-67c7f3d3fc8a"></video>  | <video src="https://github.com/user-attachments/assets/5b44238b-489f-4a0e-9c65-ae359a44e163"></video>  |

Rabie26:
JsonDecodingException: Unexpected JSON token at offset 0: $:Expected start of the object '{', but had '<' instead at path ..."JSON input: <!DOCTYPE html><html lang="ar

Smol-Ame:
@Rabie26 why are you commenting in a merged PR? Also, that's not an app-related issue, check with whoever/wherever you get your extensions from instead, not our problem.

PR #2390, opened 2025-08-10
Adding option to export to local source

### LINKED ISSUE #2166: [Feature Reqeust] - Add option to export external sorce manga to local.

Hi, firstly, thank you all for developing and maintaining this project. I've been using it for years now.

TL;DR:
I don't know if this already exists (I haven’t found it), but I think it would be great if there were an option in the Migrate menu to "Export to Local".

What I think would be really useful is the ability to export a manga from a source—or even a merged source (if that feature gets implemented, as I saw mentioned in another issue)—directly to the Local source.

Why this?
Some manga, manhwa, manhua, etc. that I’ve finished reading are no longer available due to sources being taken down (like the recent nuke of a well-known aggregator—haha), which caused some references to be lost.

I usually download chapters as I read, and then mark them for deletion once I finish. But now, I plan to keep certain titles around for longer, since I like to re-read some stories from time to time.

So, if a manga from one of my mapped sources is completed, I’d love to be able to export it to Local, preserving its metadata—without having to do it all manually by managing folders and files.

### Other details

_No response_

### Acknowledgements

- [x] I have updated the app to version **[0.18.0](https://github.com/mihonapp/mihon/releases/latest)**.

Kyougaru:
Closes #2166 by adding an option in the manga screen toolbar to export a manga to the local source.

The issue suggests having this option in the migration menu, but I believe it makes more sense to have a dedicated option. Migration requires having a corresponding manga in the target source, whereas in this case we are essentially creating a new entry in the local source.

This feature copies every downloaded chapter of the manga, along with its cover and metadata, to the local source, without requiring the user to manually rearrange files in their file explorer.

### Images

image

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
@@ -113,6 +113,7 @@ fun MangaScreen(
     onEditFetchIntervalClicked: (() -> Unit)?,
     onMigrateClicked: (() -> Unit)?,
     onEditNotesClicked: () -> Unit,
+    onClickExportToLocal: (() -> Unit)?,
 
     // For bottom action menu
     onMultiBookmarkClicked: (List<Chapter>, bookmarked: Boolean) -> Unit,
@@ -170,6 +171,7 @@ fun MangaScreen(
             onChapterSelected = onChapterSelected,
             onAllChapterSelected = onAllChapterSelected,
             onInvertSelection = onInvertSelection,
+            onClickExportToLocal = onClickExportToLocal
         )
     } else {
         MangaScreenLargeImpl(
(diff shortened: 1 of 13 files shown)


krysanify:
Thank you for doing this, I have a couple of questions:
- Seems like the condition for checking if an entry already exists is based on the folder name? Wouldn't this prevent user from exporting the same manga but from different source? What about a use case where user downloaded some chapters from one source, and then downloaded more chapters from another sources?
- It also seems like the export will fail altogether if at least one chapter fails to be exported? This is based on how `exportChapter()` is called, but I could be wrong as I haven't test it.

Xori71:
Have you given any thought to how such an option would affect site load? We have people that download hundreds of chapters for no reason. This would be their wet dream.

Edit: Nevermind, it copies the _downloaded_ chapters. My mistake.

PR #2431, opened 2025-08-22
feat: full chapter file download

paul-arthurthiery:
This PR aims to allow extensions to use an alternate download strategy (dubbed full chapter download) where whole cbz files are downloaded instead of the traditional page-by-page download.

This feature is particularly useful for users running [Suwayomi-Server](https://github.com/Suwayomi/Suwayomi-Server) and the [suwayomi extension](https://github.com/Suwayomi/tachiyomi-extension), since cbz files can be directly downloaded, without the overheard intrinsic to page-by-page download.

The feature is fully backwards compatible, with the new functions only being used by extensions if toggled via a user preference. 

I have tested this feature with various files hosted on my Suwayomi instance by running a local build of the app and the source on my device.

As this is my first attempt at contribution, feel free to point out any issue with it (be it code style, code architecture, or others). Feedback is greatly appreciated.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
@@ -478,9 +478,8 @@ class Downloader(
         download.updateFullChapterProgress(PROGRESS_START)
         download.status = Download.State.DOWNLOADING
         notifier.onProgressChange(download)
-
-        download.updateFullChapterProgress(PROGRESS_DOWNLOAD_START)
-        notifier.onProgressChange(download)
+        // Note: we only jump to PROGRESS_DOWNLOAD_START once we know content length.
+        // See streamResponseToFile() after response validation.
     }
 
     /**
@@ -588,6 +587,12 @@ class Downloader(
         val contentLength = response.body.contentLength()
         val source = response.body.source()
 
+        // If we have a known content length, nudge progress into the streaming window early
+        if (contentLength > 0) {
+            download.updateFullChapterProgress(PROGRESS_DOWNLOAD_START)
+            notifier.onProgressChange(download)
+        }
+
         chapterFile.openOutputStream().use { outputStream ->
             val sink = outputStream.sink().buffer()
             val buffer = okio.Buffer()
@@ -597,7 +602,7 @@ class Downloader(
 
             try {
                 while (true) {
-                    val bytesRead = handleStreamRead(source, buffer, totalBytesRead)
+                    val bytesRead = handleStreamRead(source, buffer, totalBytesRead, contentLength)
 
                     if (bytesRead == -1L) break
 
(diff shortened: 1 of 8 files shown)


AntsyLich:
Stubs for the new extension API needs to be added in https://github.com/mihonapp/tachiyomix. Also for compatibility with some of the endorsed forks and having prior plans for the next tachiyomix release this feature is unlikely to be merged anytime soon.

paul-arthurthiery:
> Stubs for the new extension API needs to be added in https://github.com/mihonapp/tachiyomix.

Hi ! Thanks for the comment. I had the code, but wasn't sure about the order of the sequence of PRs. I have opened the tahiyomix PR [here](https://github.com/mihonapp/tachiyomix/pull/17).

> Also for compatibility with some of the endorsed forks and having prior plans for the next tachiyomix release this feature is unlikely to be merged anytime soon.

Totally understand that this is purely me trying to remove an issue I found in my setup, and doesn't fit in any current plans. Not sure what is meant by "compatibility with endorsed forks", do I need to opened PRs with other forks to ensure compatibility ? or should I reach out to someone ? The whole release process is a blindspot to me, apologies.

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -419,6 +449,480 @@ class Downloader(
...
+    }
+
+    /**
+     * Streams the response content to the chapter file with progress tracking.
+     *
+     * @param response the HTTP response
+     * @param chapterFile the target file
+     * @param download the download for progress tracking
+     */
+    private suspend fun streamResponseToFile(response: Response, chapterFile: UniFile, download: Download) {
```
stevenyomi, line 583: This seems to provide another progress tracking approach different from what's used in `getImage()`.

https://github.com/mihonapp/mihon/blob/83f4b486296f6998efc8ef95fa9d58b75e5de130/source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/online/HttpSource.kt#L374-L377

We're hesitant about making this `open` because the progress tracking will be broken if this is overridden. Maybe this method can be extracted to apply to that too?

app/src/main/java/eu/kanade/tachiyomi/source/SourceFullChapterExtensions.kt
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * Extension function to get the full chapter download response from a source.
+ * This handles both FullChapterSource and HttpSource implementations.
+ *
+ * @param chapter the chapter to download
+ * @return Response containing the full chapter archive
+ * @throws UnsupportedOperationException if the source doesn't support full chapter downloads
+ */
+suspend fun Source.getFullChapterResponse(chapter: eu.kanade.tachiyomi.source.model.SChapter): okhttp3.Response {
```
stevenyomi, line 88: Should deal with the cache problem. Since we can't check the availability of `cache` storage in the client, I suggest using this approach:
1. Check `response.request.cacheControl.noStore == true`, or throw an exception stating that extension should disable cache in its request.
2. Document this in the KDoc of `getFullChapter()`.

Issue #2411, opened 2025-08-15
Read Duration Not Match


farizderpanzer:
### Steps to reproduce

I have been using the Mihon app for 2 weeks, but in the reading duration statistics it was recorded only 4 hours.

![Image](https://github.com/user-attachments/assets/b6e34e00-cf94-4f71-90b1-57b52631dfe7)

### Expected behavior

whereas in a day I can spend 2 hours more reading, The app should be used when the user opens the chapter page and as long as the screen is still on. 

### Actual behavior

From a few experiments when I read the manga for 30 minutes in the statistics only recorded 5 minutes. This is what makes reading duration not match the user experience 

### Mihon version

0.19.1

### Android version

Android 11

### Device

Samsung S22

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [x] I have updated the app to version **[0.19.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I have filled out all of the requested information in this form, including specific version numbers.
- [x] I understand that **Mihon does not have or fix any extensions**, and I **will not receive help** for any issues related to sources or extensions.

Salazaar36:
Same. I face the same issue.

KotlinHero:
I am trying to fix this

KotlinHero:
The cause of this bug
The reading time of Mihon is saved by chapter
When you switch chapters, swipe right back or exit the application directly, the save action will be triggered
The expected logic should be to reset the start time of reading after recording the current reading time
However, there is concurrent code
When you switch chapters, the code for resetting the reading time will be executed before calculating the reading time code
The reading end time is correct. The reading start time is always the reading end time (or 1ms earlier).
So it's not that there is no record of time. In fact, only the time when you read the last chapter will be recorded

PR #2440, opened 2025-08-25
Fix read duration not match

KotlinHero:
See #2411 

The issue of the sequence between resetting the start time and saving the reading time caused by concurrency


AntsyLich:
Thanks for the finding. I fixed it in a slightly different way #2784

PR #2476, opened 2025-09-07
Suwayomi: Support token-based authentication

cpiber:
Since https://github.com/Suwayomi/Suwayomi-Server/pull/1498 and https://github.com/Suwayomi/Suwayomi-Server/pull/1524, Suwayomi supports alternative authentication methods in addition to basic authentication already implemented in the Suwayomi tracker.

Since https://github.com/Suwayomi/tachiyomi-extension/pull/23, the extension also supports those; this PR copies relevant code to support login in the tracker as well.
Alternatively, it might be preferrable to load `TokenManager` from the extension dynamically to share that code.


AntsyLich:
Credentials for trackers like Suwayomi are usually taken from the associated source since they can have multiple instances. Don't think it'll be a good idea to change that

cpiber:
@AntsyLich You mean that the cookie should come from the extension? They're expected to be relatively short-lived, especially for the JWT tokens, I'll have to look into how other trackers handle that.

cpiber:
@AntsyLich I assume you mean that the tracker should call the extension, similar to this: https://github.com/mihonapp/mihon/blob/060e5b2e2e8ad85886c4d3362ba1ae56fac52f79/app/src/main/java/eu/kanade/tachiyomi/data/track/kavita/Kavita.kt#L127

Though your statement is incorrect, Kavita also manages tokens on its own in the tracker (but I agree that moving it to the extension is more elegant): https://github.com/mihonapp/mihon/blob/060e5b2e2e8ad85886c4d3362ba1ae56fac52f79/app/src/main/java/eu/kanade/tachiyomi/data/track/kavita/KavitaApi.kt#L40-L80

I'll see about rewriting things to use the extension's `client`, which handles re-authentication already

cpiber:
As suggested, I have simplified the code. Now the source handles everything, including the basic auth login that was previously handled here.

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/SuwayomiApi.kt
```diff
@@ -25,26 +21,21 @@ import java.security.MessageDigest
...
+    }
+    private val client: OkHttpClient by lazy {
+        source.client
+    }
+    private val headers: Headers by lazy {
+        source.headers
+    }
+    private val baseUrl: String by lazy {
+        source.baseUrl.trimEnd('/')
     }
```
AntsyLich:
```suggestion
    private val sourceManager: SourceManager by injectLazy()
    private val source: HttpSource by lazy { (sourceManager.get(sourceId) as HttpSource) }
    private val client: OkHttpClient by lazy { source.client }
    private val headers: Headers by lazy { source.headers }
    private val baseUrl: String by lazy { source.baseUrl.trimEnd('/') }
```

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/SuwayomiApi.kt
```diff
@@ -25,26 +21,21 @@ import java.security.MessageDigest
...
-            .dns(Dns.SYSTEM) // don't use DNS over HTTPS as it breaks IP addressing
-            .build()
-
-    private fun headersBuilder(): Headers.Builder = Headers.Builder().apply {
-        if (basePassword.isNotEmpty() && baseLogin.isNotEmpty()) {
-            val credentials = Credentials.basic(baseLogin, basePassword)
-            add("Authorization", credentials)
-        }
+    private val sourceManager: SourceManager by injectLazy()
+    private val source: HttpSource by lazy {
```
AntsyLich: Have you checked what happens if you remove the source after enabling the tracker?

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/Suwayomi.kt
AntsyLich, line 1: Revert the file changes or use `logcat`

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/SuwayomiApi.kt
cpiber: Tested just now: Mihon seems to remove the tracking by itself, as the tracker becomes disabled

20250918_22h24m31s_grim

Either that or there's an exception that is silently ignored that disables the tracker. I've noticed that sources also simply vanish if something cannot be constructed (e.g. evaluating preferences in constructor instead of lazy).

Reinstalling the extension shows the tracking again.

AntsyLich:
Add a changelog entry

cpiber:
> Add a changelog entry

Is this entry sufficient? Not sure what to write

CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 ### Fixes
 - Fix height of description not being calculated correctly if images are present ([@Secozzi](https://github.com/Secozzi)) ([#2382](https://github.com/mihonapp/mihon/pull/2382))
 
+### Changed
+- Suwayomi tracker: Support new authentication modes via extension ([@cpiber](https://github.com/cpiber)) ([#2476](https://github.com/mihonapp/mihon/pull/2476))
```
```suggestion
### Changed
- Delegate Suwayomi tracker authentication to extension ([@cpiber](https://github.com/cpiber)) ([#2476](https://github.com/mihonapp/mihon/pull/2476))
```

Should go above fixes. Also rebase your changes on top of main

AntsyLich (changes requested):

AntsyLich (approved):

cpiber:
I don't think those test failures are my fault? At least I don't think anything of what I did should have influenced that

PR #2483, opened 2025-09-12
Fix stuck downloads on cancel & auto refresh extensions on repo changes

c2y5:
- Fixed an issue where cancelling an extension download too quickly could cause it to remain stuck in a pending state until the app was restarted.

- Improved extension listing by making it auto refresh whenever a new repository is added or deleted (don't need manual refresh)

| Before | After |
|--------|-------|
| <video src="https://github.com/user-attachments/assets/497f613d-8b5e-4775-8cc0-bb2eebc980bc" controls width="300"></video> | <video src="https://github.com/user-attachments/assets/bc8b042a-f59c-4a0d-bd15-9f6bc68b70de" controls width="300"></video> |


MajorTanya:
For future reference, you can use this repo for demo purposes, it only contains a single dummy extension without revealing repo or extension names:
https://github.com/MajorTanya/mihon-dummyextension

c2y5:
Sorry didn't know that existed

MajorTanya:
> Sorry didn't know that existed

That's okay, not really something I or Mihon advertise as existing.

AntsyLich:
Thanks for the fix and qol improvement. Could you add entries for both in the CHANGELOG.md file?

PR #2511, opened 2025-09-19
fix(manga-restorer): Backward compatibility for old backup without `initialized`

cuong-tran:


app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/MangaRestorer.kt
```diff
@@ -139,6 +139,7 @@ class MangaRestorer(
         manga: Manga,
     ): Manga {
         return manga.copy(
+            initialized = manga.initialized || manga.description != null,
```
AwkwardPeak7: I am pretty sure setting initialized here will not end up in db, as the `insertManga` call in next line is inserting the old object.

cuong-tran: Then it's a different issue. The line I added is actually a reverted change, adding a condition for new 'initialized' field.
I will check the subsequent insert you mentioned

PR #2517, opened 2025-09-20
Use new Kitsu search URL

cpiber:
The algolia URL changed recently, so now searching always returns "Remote host terminated the handshake"


MajorTanya:
I'm not seeing any Kitsu search issues on my end, are you sure about this?

cpiber:
Interesting, you are right, it's an issue on my Wifi, where the original URL is blocked, but this new one is not (Firewall...).

In any case, this URL is what kitsu.app is using. Not sure if the old one will keep working, it's probably better to switch now. I'll change the PR to reflect this

MajorTanya:
It's not what Kitsu.app is using:
grafik

cpiber:
20250920_11h59m03s_grim


I see, it's the fallback because it's blocked on my network, I'll close this then

MajorTanya:
finally found this URL in Algolia's documentation
https://www.algolia.com/doc/guides/scaling/infrastructure/classic/distributed-search-network-dsn/#dsn-server-urls

But since these are fallback URLs, I don't know if we can properly handle that in Mihon

MajorTanya:
Maybe you should figure out why algolia.net is blocked on your WiFi network since their service is just search and infrastructure...

PR #2518, opened 2025-09-21
fix webtoon stop scroll open menu

Naputt1:
when scrolling and tap to stop the menu open

https://github.com/user-attachments/assets/917312a3-ccb4-40bf-9312-fe175605bbda

after fix

https://github.com/user-attachments/assets/ae59204a-8545-49d6-934a-7eb089a03cc6


krysanify:
maybe dont tap to stop?

to reproduce this issue, i must tap center of screen, outside of the prev/next tap zones and firmly tap instead of a "wiggle" (short swipe, not fling) which naturally happened as i repositioned the page to a readable part.

personal opinion aside, has this changes been tested along with different configurations of tap zones?

Naputt1:
I just stop the propagation of tap if the tap stop scrolling. So the tap zone won't be called if scrolling. If your using tap zone you're probably not scrolling anyway right? 

I personally turn the tap zone off and the menu just keep open like every 5 taps.

PR #2527, opened 2025-09-26
Automatically fuse two-page spreads using image analysis

TheUnlocked:
## Summary

While Mihon has an existing option to detect when an image is wide to split it in half, it currently has no feature to go the other way, detecting when two adjacent images are connected to merge them into a single wide image. This is a desirable feature because some sources only provide spreads pre-split. This PR attempts to implement this "spread fusion" by using image analysis techniques like edge detection to identify when two adjacent pages were originally connected so it can glue them back together.

Note: #2315 also attempts to address this kind of issue by showing pages in a side-by-side format in the pager viewer, but that does not help for webtoon-style viewers, and it also requires that the pages are properly aligned to make sure the left page doesn't show up on the right. The sort of approach taken in that PR could be complementary to this feature, but would not supersede it.

### Images

| w/o fusion | w/ fusion | w/ fusion + rotate |
|--------|--------|--------|
|  |  |  |

| horiz. RTL pager w/ fusion |
|--------|
| image |

| In-reader settings | Main settings |
|--------|--------|
|  |  | 

### Detecting Spreads

Spread detection follows a special algorithm designed for this purpose and implemented (with comments) [here](https://github.com/TheUnlocked/manga-vision/blob/main/mangavision/src/main/cpp/mangavision.cpp). The algorithm is not perfect, but in my real-world testing on manga I've been reading, it has a fairly high success rate. It's tuned to favor false negatives over false positives, so occasionally you will come across a spread that is not merged, but hopefully you will never come across a non-spread which is merged. Full disclosure-- I am not a computer vision person and while I feel pretty good about its performance currently, I'm sure it could be made much more reliable if anyone wants to contribute their expertise in the field.

The reason that the detection functionality is implemented as a separate library is not because it necessarily has to be. In fact, I originally implemented it to be part of this PR (see [VisionUtil.kt](https://github.com/TheUnlocked/mihon/blob/03c65bb7e1053573c2e3b984efd12ecb2f8ec4f0/core/common/src/main/kotlin/tachiyomi/core/common/util/system/VisionUtil.kt) in an intermediate commit). The problem is that this feature relies on OpenCV, and importing OpenCV as a package brings along a huge .so file containing the entire OpenCV library, blowing up the final APK size. The only way* to get around this is to compile native code against the OpenCV SDK directly, which would add significant overhead to the build process for Mihon. Additionally, in an effort to reduce the file size impact even more, manga-vision uses a custom build of the OpenCV SDK with certain hardware optimizations and other unnecessary features disabled. All of the setup, compilation, and deployment is automated using GitHub Actions at [TheUnlocked/manga-vision](https://github.com/TheUnlocked/manga-vision).

Which direction it attempts to detect spreads in (i.e. first page on the left vs first page on the right) depends on the pager type in pager mode as well as the "invert split page placement" setting in all cases.

*Technically you could also achieve a much smaller size by compiling a custom version of the .so file without unnecessary functions while still using the Java API, but that doesn't resolve any of the project maintenance issues, and the .so is still larger than it is when compiling against the native object files directly.

### Merging Spreads

Rather than make more ad-hoc changes to the page holders to handle spread fusion, this PR introduces a high-level concept of `PageLoaderInterceptor`s, which can intercept the page loading process and modify page data as appropriate before (or after) it gets initially rendered. Existing features like wide page rotation and splitting can be migrated to this model in order to simplify the page holders, though they have not been yet (splitting would be slightly more involved than rotating since it would need a new type of page state).

`SpreadFusionInterceptor` is the interceptor that handles detecting and merging spreads, and handles merging spreads by modifying the image in the first page and setting the second page's status to a new state value, `Skip`. The handling for the `Skip` state in the webtoon reader is simple, just hiding the page holder when its `ReaderPage` has that status. For the pager reader, the approach is similar to the one used when splitting wide pages, and actually removes the skipped page from the adapter. While that works for this feature, it does mean that, like with inserted pages, a skipped page cannot be un-skipped without reloading the reader, which may (or may not) be an issue for future features. However, that bridge can be crossed later.

In earlier iterations of this feature, I tried utilizing the capabilities of the interceptor more by only marking a page as `Skip` or `Ready` once it had already been checked against its neighboring pages. While this worked alright for local chapters, it added significant latency when pages had to be downloaded, and overall degraded the reading experience. The implementation in this PR eagerly renders each page as it loads, and retroactively tries to merge it later once its neighbors have also loaded.

#### Image Pipeline

Previously image data streams and buffers were used to store intermediate image data between processing steps, for example when splitting or rotating wide pages. That meant that bitmap data had to be read from an image format, processed, and then re-compressed to an image format every time it had to be modified, adding overhead and potentially introducing re-compression artifacts. The small number of cases where that would occur previously meant it was never really a big issue, but if several interceptors were used in the future, that could be an issue. To avoid this, `ReaderPage` has been enhanced to allow `Bitmap`s to be used directly as a backing data store (though all of it is still lazily loaded to avoid holding large objects in memory).

Additionally, many of the image processing operations that occurred (and still occur) inside of the page holders now use an `ImageSource` interface, which likewise can either be backed by a bitmap or by a buffer (though unlike the `ReaderPage`, these are the actual image data, not thunks to obtain the image data).

### Misc. Changes

There are a few small issues which I incidentally encountered when working on this feature which I also resolved in this PR:

* Previously there was an inconsistency where the "rotate wide pages" and "split wide pages" settings were mutually exclusive and would each unset the other in the main settings page, but would not in the in-viewer settings. This has been fixed so the in-reader settings follows the main settings logic.
* Some existing image processing steps used `BitmapFactory.decodeStream` when they should've used the [`ImageDecoder`](https://github.com/mihonapp/image-decoder) API which supports more image formats. This has been adjusted so that image processing still tries to use `BitmapFactory.decodeStream` first because it is significantly faster, but now falls back to `ImageDecoder` when that fails.
* The way settings are loaded into viewers has been adjusted slightly to reduce unnecessary refreshes/updates.


TheUnlocked:
To address the issue of the large binary size, I spent the last week or so researching OpenCV's source code and developing a custom implementation of every behavior I used OpenCV for (with the exception of gaussian blur, where I utilized [a library](https://github.com/android/renderscript-intrinsics-replacement-toolkit) that has an optimized implementation). The result is that with the latest version of [`manga-vision`](https://github.com/TheUnlocked/manga-vision), instead of adding ~13MB to the universal APK size, it adds <1.5MB.

Additionally, as an unexpected but welcome benefit from this change, the overall spread detection performance is ~3x faster in rough e2e testing. My guess is that this is due to reduced overhead from not needing to convert to and from OpenCV data structures (It's not because my naive algorithm implementations are faster than their SIMD-optimized implementations--I've benchmarked them and they're not).

AntsyLich:
Now we're talking! But uh I would potentially like to have the repo under @mihon if you don't mind.

TheUnlocked:
That's completely fine, but you'll need to handle deployment to Maven Central (or whatever other repo you want) on your own.

AntsyLich:
can you do a rebase

TheUnlocked:
Will run spotless. I also want to do a quick spot check on functionality to make sure the rebase didn't break anything (though it was conflict-free aside from the changelog so I'm optimistic).

TheUnlocked:
Looks okay in-app.

PR #2532, opened 2025-09-28
cache browse source query result

Naputt1:
https://github.com/user-attachments/assets/e596958a-e17f-46d6-842b-3155cb5ae86f


AntsyLich:
why

Naputt1:
why not? 
the search result is unlikely to change anyway

BrutuZ:
Cached for how long?

Couldn't this pose an issue with self-hosting if you add an entry between queries?

Naputt1:
in memory cache 
just close and reopen to reload

cuong-tran:
Should done this on extension side

BrutuZ:
> in memory cache just close and reopen to reload

That's quite a step down in QoL. Should probably include a visual feedback at least or a flush cache button to avoid confusion and issue reports for the app and/or extension maintainers

AntsyLich:
HTTP requests are already cached. The 1 seconds difference in loading won't bother anyone.

PR #2533, opened 2025-09-28
feat: Added Search via Clipboard

Lelantos-sh:
Feature: Search directly from the copied text aka Directly via Clipboard

< Demo Video To Be Added >

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreen.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreen.kt
@@ -12,8 +12,10 @@ import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.outlined.Favorite
 import androidx.compose.material.icons.outlined.FilterList
 import androidx.compose.material.icons.outlined.NewReleases
+import androidx.compose.material.icons.outlined.ContentPaste
 import androidx.compose.material3.FilterChip
 import androidx.compose.material3.FilterChipDefaults
+import androidx.compose.material3.FloatingActionButton
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
 import androidx.compose.material3.MaterialTheme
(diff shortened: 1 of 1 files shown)


AntsyLich:
why

Lelantos-sh:
@AntsyLich If I have the keywords copied in my clipboard, either title or secret number, I find it much more convenient to just press one button and do the complete search for me with that keyword, rather than me pressing "Search" -> Hold Text field to get Paste option, -> trigger actual search,

It saves me from too many taps.

Lelantos-sh:
I have updated the PR description and added videos, don't know how to get them to show preview

AntsyLich:
Searching from clipboard is not that common of a action that it deserves a FAB

Lelantos-sh:
If so this can be made as Toggable feature, helpful for those who want this

AntsyLich:
hmmm i suppose i'm fine with that

Lelantos-sh:
@AntsyLich I have added the toggle control. Default experience would be same as existing one. If interested, User can enable this feature

vetleledaal:
Most people will just use the defaults, so hiding the feature behind a toggle seems counterintuitive. Could checking if the clipboard was recently updated be a decent middle-ground? Browsers use something similar.

- [ClipboardManager.getPrimaryClipDescription()](https://developer.android.com/reference/android/content/ClipboardManager#getPrimaryClipDescription())
- [ClipDescription.getTimestamp()](https://developer.android.com/reference/android/content/ClipDescription#getTimestamp())
- [Chrome implementation](https://github.com/chromium/chromium/blob/7a1dfffbee666772b953f2f3958ed0a8042da474/ui/android/java/src/org/chromium/ui/base/ClipboardImpl.java) (BSD 3-Clause license)
- [Firefox implementation](https://github.com/mozilla-firefox/firefox/blob/d061baf359c3a55c0ff0a4a8cc99259e738fcf28/mobile/android/geckoview/src/main/java/org/mozilla/gecko/Clipboard.java) (MPL license)

Lelantos-sh:
The point of moving this behind settings page was that it is not as common an action that it deserves a space in the source screen, as per Antsy

Anyway there are plenty of features behind settings and this particular toggle option will sit in Browser settings which is almost always visited to add new extension repos.

Also, what is the time clipboard history TTL? If a user has some random stuff copied in their Clipboard all the time then this your suggestion may seem irregular to user since they won't understand the behaviour outright whether it is showing up because they have clipboard item or it is always shown - may attribute as Bug.l

Lelantos-sh:
@AntsyLich Please review, and please let me know when we can merge this feature?

PR #2537, opened 2025-09-29
Fix Reader's system bars

cuong-tran:
Fix these issues:
1. In fullscreen mode, the status bar is always transparent and hard to see
2. In non-fullscreen mode, the reader's bottom bar is hidden behind the system navigation bar
3. Also in non-fullscreen mode, the page indicator is put behind the system navigation bar

These issues happen after upgrading app to target Android SDK 35 and would affect devices running Android 15 and later.


cuong-tran:
close as #1908 is more properly

PR #2584, opened 2025-10-11
Handle popups properly as separate WebView instances

TheUnlocked:
Currently because the current WebView settings have `supportMultipleWindows()` set to the default of `false`, calling `window.open()` in a WebView just causes a redirect rather than opening a new window. This is a problem for a small number of sources which perform authentication through a login popup since the calling page is being completely replaced and cannot communicate with the popup.

Unfortunately fixing this is not as simple as just calling `setSupportMultipleWindows(true)`, since each popup window needs its own WebView, and those WebViews need to be rendered in order to be interactable. This PR achieves that by modifying `WebViewScreenContent` so that it stores a stack of WebView states rather than just a single state, re-creating the `WebView()` composable whenever the current top-level window changes. The `WebView` objects themselves are not destroyed in this process (except when the window is closed), allowing them to continue to communicate between each other even though only one is displayed on screen.

I don't think this should cause any issues with existing pages, though there is a change in behavior where if a user was used to popups on a particular site being treated as redirects, they might be surprised when attempting to close the WebView only closes one layer rather than returning to the main app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Spoofing of `X-Requested-With` header to support newer WebView versions ([@Guzmazow](https://github.com/Guzmazow)) ([#2491](https://github.com/mihonapp/mihon/pull/2491))
 - Download support for chapters with the same metadata. Now a hash based on chapter's url is appended to download filename to tell them apart, letting you download both. Existing downloaded chapters will continue to work normally ([@raxod502](https://github.com/radian-software)) ([#2305](https://github.com/mihonapp/mihon/pull/2305))
 - Auto refresh extension list whenever a repository is added or removed ([@c2y5](https://github.com/c2y5)) ([#2483](https://github.com/mihonapp/mihon/pull/2483))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Spoofing of `X-Requested-With` header to support newer WebView versions ([@Guzmazow](https://github.com/Guzmazow)) ([#2491](https://github.com/mihonapp/mihon/pull/2491))
 - Download support for chapters with the same metadata. Now a hash based on chapter's url is appended to download filename to tell them apart, letting you download both. Existing downloaded chapters will continue to work normally ([@raxod502](https://github.com/radian-software)) ([#2305](https://github.com/mihonapp/mihon/pull/2305))
 - Auto refresh extension list whenever a repository is added or removed ([@c2y5](https://github.com/c2y5)) ([#2483](https://github.com/mihonapp/mihon/pull/2483))
+- The WebView activity now launches new WebViews for popups instead of treating them as redirects. The X button will close the most recent WebView in the stack. ([@TheUnlocked](https://github.com/TheUnlocked)) ([#2584](https://github.com/mihonapp/mihon/pull/2584))
```
```suggestion
- Added proper multi window support in WebView instead of treating everything as a redirect ([@TheUnlocked](https://github.com/TheUnlocked)) ([#2584](https://github.com/mihonapp/mihon/pull/2584))
```

PR #2585, opened 2025-10-11
Implement Suwayomi tracking via Extension

cpiber:
Until now, the Suwayomi tracker was still using the V1 REST API instead of the new GraphQL API. Since this is planned to be removed at some point, I took this opportunity to move all API access completely to the extension.

I have not found another tracker using this approach, so I did it in a way that makes sense to me. With Suwayomi/tachiyomi-extension#26, the extension fulfils the new interface included in this PR; the tracker can then use this dedicated interface to interact with the API without implementing anything in Mihon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Option to customize the number of concurrent source and page downloads ([@AntsyLich](https://github.com/AntsyLich)) ([#2637](https://github.com/mihonapp/mihon/pull/2637))
 
 ### Changed
-- Delegate Suwayomi tracker authentication to extension ([@cpiber](https://github.com/cpiber)) ([#2476](https://github.com/mihonapp/mihon/pull/2476))
 - Increased default concurrent page downloads to 5 ([@AntsyLich](https://github.com/AntsyLich)) ([#2637](https://github.com/mihonapp/mihon/pull/2637))
-- Update Suwayomi tracker methods to GraphQL interface ([@cpiber](https://github.com/cpiber)) ([#2585](https://github.com/mihonapp/mihon/pull/2585))
 
 ### Improved
 - Spoofing of `X-Requested-With` header to support newer WebView versions ([@Guzmazow](https://github.com/Guzmazow)) ([#2491](https://github.com/mihonapp/mihon/pull/2491))
@@ -39,7 +37,9 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix date picker not allowing the same start and finish date in negative time zones ([@AntsyLich](https://github.com/AntsyLich), [@kashish-aggarwal21](https://github.com/kashish-aggarwal21)) ([#2617](https://github.com/mihonapp/mihon/pull/2617))
 
 ### Other
+- Delegate Suwayomi tracker authentication to extension ([@cpiber](https://github.com/cpiber)) ([#2476](https://github.com/mihonapp/mihon/pull/2476))
 - Fix Kitsu tracker to conform to tracker data structure properly ([@cpiber](https://github.com/cpiber)) ([#2609](https://github.com/mihonapp/mihon/pull/2609))
+- Update Suwayomi tracker to use GraphQL API instead of REST API ([@cpiber](https://github.com/cpiber)) ([#2585](https://github.com/mihonapp/mihon/pull/2585))
 
 ## [v0.19.1] - 2025-08-07
 ### Changed


cpiber:
Just realized that this breaks the extension for readers that don't provide this new interface. I will be searching for a solution to that, any suggestions are welcome

cpiber:
Found a way, turns out that Java does not actually care if the interface or class exists as long as that type is not used. So we can instead have a method which returns a `SuwayomiExtensionInterface`; the extension loads because the main class is valid, and it's fine that it has a method of which it does not know the actual return type

This requires a bit of reflection to find that method, which I don't entirely like, but I think it's a workable solution, better suggestions welcome

cpiber:
One more solution I could think of would be to move the types to the extension. I expect that would only cause a `ClassNotFoundException` when using the interface, which would leave everything else functional like it is now

AntsyLich:
Move the extension API to GraphQL. I don't see the point of delegating to the extension

cpiber:
Understood. I assume you don't want helper libraries like Apollo?

cpiber:
Maybe a quick argument why delegating to the extension should be done:

- Further code simplification on Mihon's end. The GraphQL API is a bit more verbose, requiring three separate calls to mimic the tracker update as it is currently done
- Reduced workload for you: Any future changes to the API can be quickly implemented in the extension, further reducing your maintenance workload and faster turnover on Suwayomi's side

For example, tracker authentication (#2476 is still not released, causing confusion on our side where people use it not realizing that the tracker doesn't support those modes, but the extension does

cpiber:
Note that I rebased this on top of #2609 for the mgiration, so please ignore the Kitsu-related changes

MajorTanya:
> Note that I rebased this on top of #2609 for the mgiration, so please ignore the Kitsu-related changes

Then please make this PR a draft if it depends on he Kitsu PR (and add a note that it's blocked by that PR). Then rebase on main instead once the Kitsu PR is merged.

cpiber:
Blocked by #2609

cpiber:
FYI, that test isn't failing locally, so I assume that is a flaky test

AntsyLich:
it is

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/SuwayomiApi.kt
```diff
@@ -26,66 +30,179 @@ class SuwayomiApi(private val trackId: Long) {
...
+    suspend fun getTrackSearch(mangaId: Long): TrackSearch = withIOContext {
+        val query = """
+query GetManga(${'$'}mangaId: Int!) {
+    manga(id: ${'$'}mangaId) {
+        ...MangaFragment
+    }
+}
 
+$MangaFragment
+        """.trimMargin()
```
MajorTanya, line 45: Please take a look at how [`trimMargin`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/trim-margin.html) should be used to avoid this dedent:

https://github.com/mihonapp/mihon/blob/f9793d33233ceb7729fa9cfe5ae55b45a372c54c/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/AnilistApi.kt#L44-L52

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/SuwayomiApi.kt
```diff
@@ -26,66 +30,179 @@ class SuwayomiApi(private val trackId: Long) {
...
+        val chaptersQuery = """
+query GetMangaUnreadChapters(${'$'}mangaId: Int!) {
+  chapters(condition: {mangaId: ${'$'}mangaId, isRead: false}) {
+    nodes {
+      id
+      chapterNumber
+    }
+  }
+}
+        """.trimMargin()
```
MajorTanya, line 94: see above

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/SuwayomiApi.kt
```diff
@@ -26,66 +30,179 @@ class SuwayomiApi(private val trackId: Long) {
...
+
+        val markQuery = """
+mutation MarkChaptersRead(${'$'}chapters: [Int!]!) {
+  updateChapters(input: {ids: ${'$'}chapters, patch: {isRead: true}}) {
+    chapters {
+      id
+    }
+  }
+}
+        """.trimMargin()
```
MajorTanya, line 124: see above

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/SuwayomiApi.kt
```diff
@@ -26,66 +30,179 @@ class SuwayomiApi(private val trackId: Long) {
...
-        ).awaitSuccess()
+        val trackQuery = """
+mutation TrackManga(${'$'}mangaId: Int!) {
+  trackProgress(input: {mangaId: ${'$'}mangaId}) {
+    trackRecords {
+      lastChapterRead
+    }
+  }
+}
+        """.trimMargin()
```
MajorTanya, line 151: see above

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/SuwayomiApi.kt
```diff
@@ -26,66 +30,179 @@ class SuwayomiApi(private val trackId: Long) {
...
+        fetchedAt
+    }
+    latestReadChapter {
+        lastReadAt
+        chapterNumber
+    }
+    unreadCount
+    downloadCount
+}
+        """
```
MajorTanya: use [`trimMargin`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.text/trim-margin.html) to avoid the deep dedent (see above)

MajorTanya (dismissed):

AntsyLich:
What happens if the user's suwayomi instance doesn't support graphql api

Syer10:
Thats very unlikely. The GraphQL API has been out for over a year and a half now, and the last stable before GQL didn't even have support for extension repos.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 ### Changed
 - Delegate Suwayomi tracker authentication to extension ([@cpiber](https://github.com/cpiber)) ([#2476](https://github.com/mihonapp/mihon/pull/2476))
 - Increased default concurrent page downloads to 5 ([@AntsyLich](https://github.com/AntsyLich)) ([#2637](https://github.com/mihonapp/mihon/pull/2637))
+- Update Suwayomi tracker methods to GraphQL interface ([@cpiber](https://github.com/cpiber)) ([#2585](https://github.com/mihonapp/mihon/pull/2585))
```
AntsyLich: Not a UI/UX thing so goes under Other. Same for the Suwayomi auth entry I suppose

PR #2591, opened 2025-10-13
Set library update time

tempacct-mihon:
Currently the method for scheduling library updates at a given time is tedious to set up as it just uses the time the first update was scheduled. It also uses PeriodicWorkRequests which aren't very precise as they will update within a window resulting in the update time steadily drifting from what it was originally set from.

This PR adds a new setting in the library settings menu to set the time the library update will be triggered. It calculates the delay in minutes until the specified update time and then enqueues a LibraryUpdateJob OneTimeWorkRequest to run after that many minutes. After each LibraryUpdateJob a new OneTimeWorkRequest is enqueued for the user specified time.

These changes have been tested on every update delay interval, and the UI changes have been tested with all base themes and tablet mode

### Images

| Menu option | Time picker |  Tablet mode |
| ------- | ------- | ------- |
| menu item | time picker | time picker tablet |

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsLibraryScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsLibraryScreen.kt
@@ -43,6 +43,9 @@ import tachiyomi.presentation.core.i18n.stringResource
 import tachiyomi.presentation.core.util.collectAsState
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
+import android.text.format.DateFormat
+import java.time.LocalTime
+import java.time.format.DateTimeFormatter
 
 object SettingsLibraryScreen : SearchableSettings {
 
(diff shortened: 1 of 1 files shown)


MajorTanya:
What does that look like in 24 hour format? I refuse to use 12 hour format for anything like this.

BrutuZ:
Should probably include a disclaimer about workers being inaccurate / drifting over time. Or at the very least an "approximately"
Once you add a clock people will have even more reason to expect alarm-like precision down to the minute.

tempacct-mihon:
@MajorTanya It now adapts to whatever time format your device is using
### Images

| Menu option | Time picker |
| ------- | -------|
| menu item 24 | time picker 24 |

tempacct-mihon:
> Should probably include a disclaimer about workers being inaccurate / drifting over time. Or at the very least an "approximately" Once you add a clock people will have even more reason to expect alarm-like precision down to the minute.

@BrutuZ This will fix the drifting issue and in my experience testing this it is precise to the minute. With restricted battery it may lose some precision tho.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
```diff
@@ -510,5 +479,73 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
...
+                .setConstraints(constraints)
+                .setBackoffCriteria(BackoffPolicy.LINEAR, 10, TimeUnit.MINUTES)
+                .build()
+
+            // Enqueue request
+            context.workManager.enqueueUniqueWork(
+                WORK_NAME_AUTO,
+                ExistingWorkPolicy.REPLACE,
+                request,
+            )
```
NGB-Was-Taken, line 548: Was this AI generated? Why is there a comment on pretty much every line?

tempacct-mihon: i like to be thorough with my comments 🤷‍♂️tho ig i did go a bit overboard my b

PR #2609, opened 2025-10-23
[#2605] Migrate Kitsu to use library_id and remote_id properly

### LINKED ISSUE #2605: Kitsu tracker uses `remote_id` incorrectly
### Steps to reproduce

1. Add Kitsu tracking to a title
2. Open DB
3. Notice that the entry has `remote_id` set, but not to the remote manga ID, but rather to the library entry ID, which should be `library_id` (which is empty)

### Expected behavior

`remote_id` should be the remote manga ID
`library_id` should be the handle to update the user's library data

### Actual behavior

The entry has `remote_id` set, but not to the remote manga ID, but rather to the library entry ID, which should be `library_id` (which is empty)

### Crash logs

_No response_

### Mihon version

v0.19.1

### Android version

Android 11

### Device

-

### Other details

Only Kitsu appears to have this problem. Other trackers appear to either properly use `library_id`, or not require this in the first place. The MAL API for example expects the manga ID, not a separate library ID for updating user data. Which is also likely where this bug came from, since `bind` is suspiciously similar between those two.

This was originally noticed in Suwayomi. Mihon does not appear to use those IDs outside of the respective tracker. I would work on a PR to update the usage (with a migration to properly set those fields for existing track entries).

### Acknowledgements

cpiber:
Closes #2605

Tested:
- Migrating existing track entries with both Kitsu and other trackers enabled
- Adding tracker for entries not already tracked on Kitsu
- Adding tracker for entries already tracked on Kitsu
- Refreshing Kitsu data (pull & push)
- Removing Kitsu data (remove remotely)

Note that I also included setting the remote_id for Suwayomi tracker, which also currently does not use it, and the migration is identical. Since I have started an API migration in #2585, I don't think it makes sense to update the tracker here, but let me know what you think. I'd rebase the other PR on top of this and have it use remote_id there as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix migration progress not updating after manual search ([@Secozzi](https://github.com/Secozzi)) ([#2484](https://github.com/mihonapp/mihon/pull/2484))
 - Fix category migration flag being ignored due to incorrect check against chapter flag ([@Secozzi](https://github.com/Secozzi)) ([#2484](https://github.com/mihonapp/mihon/pull/2484))
 - Fix disabling incognito mode from notification ([@NGB-Was-Taken](https://github.com/NGB-Was-Taken)) ([#2512](https://github.com/mihonapp/mihon/pull/2512))
+
+### Other
 - Fix Kitsu tracker to conform to tracker data structure properly ([@cpiber](https://github.com/cpiber)) ([#2609](https://github.com/mihonapp/mihon/pull/2609))
 
 ## [v0.19.1] - 2025-08-07


AntsyLich:
Add a changelog entry

cpiber:
Sorry about that, completely forgot, added it now

AntsyLich:
Should go under "Other" in my opinion

data/src/main/sqldelight/tachiyomi/migrations/8.sqm
```diff
@@ -0,0 +1,7 @@
+-- Save the current remote_id as library_id, since old Kitsu tracker did not use this correctly
+UPDATE manga_sync SET library_id = remote_id WHERE sync_id = 3;
+
+-- Kitsu and Suwayomi aren't using the remote_id field properly, but for both the ID is present in the URL
```
AntsyLich, line 4: Does the suwayomi stuff need #2585?

cpiber: No, currently the field is unused, so this just sets the field to a sensible value, but it's not used in code. #2585 then uses it in code (to not need a duplicate migration)

Issue #2400, opened 2025-08-13
Show "Chapters Remaining"


Kisev2:
### Describe your suggested feature

When reading a manga or mahwa. it would be really helpful to see how many chapters are left without having to calculate it manually. For example, if I'm on Chapter 37 of 50, the app could display "13 chapters remaining" alongside the chapter number.

I know this is a very subtle change, but small quality-of-life features like this make a big difference to the reading experience

Personally I think it will keep readers immersed into reading, no pausing no need to go back to home page which will mostlikely ruin the mood. And most of the time i forgot how many chapters the manhwa im reading. 

And it can help people who are tight in time. So they know how many chapters they have left. If they can finish it in this time, qnd just feel thoughtful that the app is anticipating the user needs.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.19.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

drishtiisharma:
@Exspress1onscript kindly assign me this issue

quoipi:
Working on this - submitting a PR shortly

PR #2614, opened 2025-10-25
Issue #2400: Show "Chapters Remaining"

quoipi:
Added a "Chapters remaining" indicator below the existing Chapter Navigator based on the current chapter being viewed.

image

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/reader/components/ChapterNavigator.kt
+++ b/app/src/main/java/eu/kanade/presentation/reader/components/ChapterNavigator.kt
@@ -77,81 +77,80 @@ fun ChapterNavigator(
             verticalArrangement = Arrangement.spacedBy(4.dp)
         ) {
             // We explicitly handle direction based on the reader viewer rather than the system direction
-            CompositionLocalProvider(LocalLayoutDirection provides LayoutDirection.Ltr) {
-                Row(
-                    modifier = Modifier
-                        .fillMaxWidth()
-                        .padding(horizontal = horizontalPadding),
-                    verticalAlignment = Alignment.CenterVertically,
-                ) {
-                    FilledIconButton(
-                        enabled = if (isRtl) enabledNext else enabledPrevious,
-                        onClick = if (isRtl) onNextChapter else onPreviousChapter,
-                        colors = buttonColor,
-                    ) {
-                        Icon(
-                            imageVector = Icons.Outlined.SkipPrevious,
-                            contentDescription = stringResource(
-                                if (isRtl) MR.strings.action_next_chapter else MR.strings.action_previous_chapter,
-                            ),
-                        )
-                    }
 
-                    if (totalPages > 1) {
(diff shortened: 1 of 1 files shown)


NGB-Was-Taken:
That displays pages remaining and not chapters remaining though? Pages =/= Chapters. If you were going for pages instead of chapters, I believe it to be unnecessary as there is already `Settings > Reader > Show page number` which shows current page and total pages.

quoipi:
> That displays pages remaining and not chapters remaining though? Pages =/= Chapters. If you were going for pages instead of chapters, I believe it to be unnecessary as there is already `Settings > Reader > Show page number` which shows current page and total pages.

Oh, that's true! I misunderstood the purpose of this bit of UI - working to fix this now

quoipi:
Should be fixed now!

app/src/main/java/eu/kanade/presentation/reader/components/ChapterNavigator.kt
```diff
@@ -67,80 +71,109 @@ fun ChapterNavigator(
...
-                CompositionLocalProvider(LocalLayoutDirection provides layoutDirection) {
-                    Row(
-                        modifier = Modifier
-                            .weight(1f)
-                            .clip(RoundedCornerShape(24.dp))
-                            .background(backgroundColor)
-                            .padding(horizontal = 16.dp),
-                        verticalAlignment = Alignment.CenterVertically,
+            // We explicitly handle direction based on the reader viewer rather than the system direction
+            CompositionLocalProvider(LocalLayoutDirection provides LayoutDirection.Ltr) {
```
NGB-Was-Taken: Why is there a second LTR wrapper? Is this necessary, or a mistake?

app/src/main/java/eu/kanade/presentation/reader/components/ChapterNavigator.kt
```diff
@@ -67,80 +71,109 @@ fun ChapterNavigator(
...
+                        "Chapter"
+                    } else {
+                        "Chapters"
+                    }
+
+                Text(
+                    modifier = Modifier
+                        .weight(1f)
+                        .padding(horizontal = 8.dp),
+                    text = "$remainingChapters $chapterVsChapters Remaining",
```
NGB-Was-Taken: Should be done through i18n plurals and not hard-coded here.

quoipi: Ok! I'll commit changes to fix this soon :)

app/src/main/java/eu/kanade/presentation/reader/components/ChapterNavigator.kt
quoipi: Oh - this might be a mistake I missed, thanks for catching that :) I'll remove it shortly

quoipi:
Should be up to date now 👍

PR #2641, opened 2025-11-03
Update CONTRIBUTING.md

nguyd1:
added local development and debugging sections and added verbose instructions


AntsyLich:
Seems rather pointless. If you're not familiar with with basics of Android development why are you even here

PR #2648, opened 2025-11-03
add cloudflare help button to manga screen

nguyd1:
added cloudflare help button to manga screen similar to webview screen
added more info to cloudflare fail message

Screenshot_20251103_140709_Mihon


AntsyLich:
I think it's fine as is. Support people can probably chip in

PR #2649, opened 2025-11-04
Fix webview crash

bapeey:
Closes #2643

[Screen_recording_20251103_224434.webm](https://github.com/user-attachments/assets/ce09bcd4-4c35-4737-9164-08224a53e69d)

PR #2662, opened 2025-11-06
Improve WebView multi-window UX

TheUnlocked:
#2584 included a change that made popups actually open a new WebView instance to resolve issues with some sites. I expected popups to be fairly niche and so I didn't put a huge amount of effort into UX around it, but it turns out that `target="_blank"` in links is fairly common and also opens a new popup/window/tab, so this PR tries to address the pain points around that and generally make the experience better.

**Changes:**
- Navigation history for lower windows is preserved when a popup is opened.
- When a popup window has no previous page, the back button/gesture will now close that window rather than the entire WebView activity.
- The leftmost close button now closes the entire activity like it did prior to #2584. There is instead a new button to close just the current window.
- Fixed an issue where when closing a popup window, an indeterminate progress bar would appear for the window below it.

The new button uses [Tab Close](https://fonts.google.com/icons?icon.query=tab+cl&icon.size=24&icon.color=%23e3e3e3&icon.platform=web&icon.set=Material+Symbols&selected=Material+Symbols+Outlined:tab_close:FILL@0;wght@400;GRAD@0;opsz@24) from Material Symbols and looks like this:
| No popups open | Currently showing popup window |
|--------|--------|
| image | image |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - `Other` - for technical stuff.
 
 ## [Unreleased]
+### Improved
+- Improved various aspects of the WebView multi window support added in 0.19.2 ([@TheUnlocked](https://github.com/TheUnlocked)) ([#2662](https://github.com/mihonapp/mihon/pull/2662))
 
 ## [v0.19.3] - 2025-11-04
 ### Fixed


MajorTanya:
Why not use the icon directly? We have the Material Icons library in the project...

TheUnlocked:
This icon doesn't exist in Material Icons, only Material Symbols. I checked for any alternatives that might work instead, but unless I used the same close icon as is shown on the left there weren't any great options.

MajorTanya:
> This icon doesn't exist in Material Icons, only Material Symbols. I checked for any alternatives that might work instead, but unless I used the same close icon as is shown on the left there weren't any great options.

It really doesn't exist in the library we have already? Ugh

Smol-Ame:
Just curious, reminder, am not programmer/contributor, but how about merged #2649? (Or does it work alongside with that fix?)

TheUnlocked:
> It really doesn't exist in the library we have already? Ugh

They also don't have any other library for material symbols, their instructions are just to directly include the XML files in the project. There is an open feature request for an equivalent library for material symbols, but who knows if that will ever happen: https://issuetracker.google.com/issues/447545316

> Just curious, reminder, am not programmer/contributor, but how about merged https://github.com/mihonapp/mihon/pull/2649? (Or does it work alongside with that fix?)

Works alongside it. That fixed the crash, this improves UX.

MajorTanya:
> > It really doesn't exist in the library we have already? Ugh
> 
> They also don't have any other library for material symbols, their instructions are just to directly include the XML files in the project. There is an open feature request for an equivalent library for material symbols, but who knows if that will ever happen: https://issuetracker.google.com/issues/447545316
> 

Classic Google move right there. Thanks for the link to that 👍

PR #2663, opened 2025-11-06
Fix flaky migration tests

NGB-Was-Taken:


MajorTanya:
Well it passed here at least 😅

Issue #1575, opened 2024-12-15
Suwayomi Tracking Issues


s4ik0s4n:
### Steps to reproduce

I've noticed 2 issues with the Suwayomi Enhanced Tracking. Not sure if its just me or if anyone else is having these issues.

1. The first issue has to do with manga on MANGA Plus by SHUEISHA extension. Since the website usually only hosts the first 3 and last 3 chapters of any running series, the tracking seems to freak out when I read the most recent chapter, then marks the series as complete. Any further updates (like on weekly series) don't show up as new chapters in the app, but my Suwayomi server will still download them. Is this an issue anybody else is having?

The second issue I'm seeing is tracking that involves chapters that are split into multiples, such as Chapter 76.1 and 76.2. The tracking doesn't appear to update to this non-whole numbers, so my Suwayomi server doesn't update and delete the chapters after being or read, or update the most recently read chapter on the server.

I wasn't sure if these issues should have been posted here or on the Suwayomi github. I figured you guys were the ones managing the tracker extension, so I decided to post here.

### Expected behavior

1. Update tracker to the most recent read chapter while keeping the status as "Reading"

2. Update the tracker to most recent read chapter such as "Chapters 76.2" and delete read chapters on server.

### Actual behavior

1. Tracker is updated to complete and further chapters are not downloaded unless manually unmarked as read and downloaded.

2. Tracker does not update on app or server unless chapter number is a whole number without a decimal.

### Mihon version

Stable 0.17.1

### Android version

Android 14

### Device

Samsung Galaxy Z Fold 5

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [X] I have updated the app to version **[0.17.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

zerebos:
I have also been experiencing the second issue (tracking split chapters) with Mihon and other forks. But tracking for split chapters actually works properly with TachiyomiJ2K and its fork Yokai.

[...]

PR #2673, opened 2025-11-07
[#1575] Delete downloaded chapters when marking as read

cpiber:
Also reduce the amount of data queried, since we don't actually read it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 
 ## [Unreleased]
 ### Added
-- Suwayomi: Support deleting downloads after reading, changeable via extension setting ([@cpiber](https://github.com/cpiber)) ([#2673](https://github.com/mihonapp/mihon/pull/2673))
+- Automatically remove downloads on Suwayomi after reading, configurable via extension settings ([@cpiber](https://github.com/cpiber)) ([#2673](https://github.com/mihonapp/mihon/pull/2673))
 
 ### Improved
 - Minimize memory usage by reducing in-memory cover cache size ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2266](https://github.com/mihonapp/mihon/pull/2266))


cpiber:
I originally opened this, because I thought this was a regression introduced by #2585. Now I see this seems to have never been the case? The server never deleted chapters, and there's a case to make that this is not desired. So I'm leaving this open for discussion @Syer10 @s4ik0s4n

Syer10:
This is probably something that should be optional.

cpiber:
WebUI has a setting for it, but I don't think it's a good idea to reference that. Extension setting? Or a new server setting

Syer10:
Putting it in the extension settings would probably be best.

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix mass migration not using the same search queries as individual migration ([@AntsyLich](https://github.com/AntsyLich)) ([#2736](https://github.com/mihonapp/mihon/pull/2736))
 - Fix reader not saving read duration when changing chapter ([@AntsyLich](https://github.com/AntsyLich), [@KotlinHero](https://github.com/KotlinHero)) ([#2784](https://github.com/mihonapp/mihon/pull/2784))
 
+### Other
+- Suwayomi: Support deleting downloads after reading ([@cpiber](https://github.com/cpiber)) ([#2673](https://github.com/mihonapp/mihon/pull/2673))
```
AntsyLich: Should be under added. Also doesn't explain how it works given the option is not in the app.

app/src/main/java/eu/kanade/tachiyomi/data/track/suwayomi/Suwayomi.kt
```diff
@@ -23,6 +23,9 @@ class Suwayomi(id: Long) : BaseTracker(id, "Suwayomi"), EnhancedTracker {
         const val UNREAD = 1L
         const val READING = 2L
         const val COMPLETED = 3L
+
+        private const val TRACKER_DELETE_KEY = "Tracker Delete"
```
AntsyLich, line 27: this is not how a key is supposed to be :|

cpiber: That's how keys work in the Suwayomi extension, someone did it like this and I'm not going to change it (because it breaks existing installs)

CHANGELOG.md
cpiber: @AntsyLich done

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - `Other` - for technical stuff.
 
 ## [Unreleased]
+### Added
+- Suwayomi: Support deleting downloads after reading, changeable via extension setting ([@cpiber](https://github.com/cpiber)) ([#2673](https://github.com/mihonapp/mihon/pull/2673))
```
```suggestion
- Automatically remove downloads on Suwayomi after reading, configurable via extension settings ([@cpiber](https://github.com/cpiber)) ([#2673](https://github.com/mihonapp/mihon/pull/2673))
```

PR #2675, opened 2025-11-07
[#1575] Suwayomi: Allow for a small epsilon difference in chapter number

cpiber:
Due to a `Float->Double->Float` conversion somewhere inside Mihon, the tracker sees 2.1 as 2.0999999046325684, which means this filter ignores the 2.1 chapter (which we just tried to mark as read). This small epsilon is small enough to never bother any serious usage, but large enough to ignore any such conversion errors.

See https://github.com/mihonapp/mihon/issues/1575#issuecomment-3504955974

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 
 ### Changed
 - Update tracker icons ([@AntsyLich](https://github.com/AntsyLich)) ([#2773](https://github.com/mihonapp/mihon/pull/2773))
-- Suwayomi tracker: Allow a small difference in chapter number for matching to avoid incorrect chapter progress sync ([@cpiber](https://github.com/cpiber)) ([#2657](https://github.com/mihonapp/mihon/pull/2675))
+- Add a small increment to chapter number before comparison to fix progress sync issues for Suwayomi ([@cpiber](https://github.com/cpiber)) ([#2657](https://github.com/mihonapp/mihon/pull/2675))
 
 ### Fixed
 - Fix reader tap zones triggering after scrolling is stopped by tapping ([@NGB-Was-Taken](https://github.com/NGB-Was-Taken)) ([#2680](https://github.com/mihonapp/mihon/pull/2680))


zerebos:
Alternative to using an epsilon you can virtually "restore" the precision since we're dealing with small clean decimals.

```java
Math.round((float) track.last_chapter_read * 1000.0) / 1000.0
```

which works fairly well

image

AntsyLich:
i think it's better to migrate the codes to ditch the legacy tracker models which use float instead of double

cpiber:
@AntsyLich I'm not quite sure what you mean. I looked a bit through it, and both `Track` and `TrackSearch` use `Double` for `last_chapter_read`. I see that `DelayedTrackingStore` and `ChapterRepository` use `Float`, so do you mean to update stuff there?

Though I'm not sure why the latter uses float, I think this line is bogus:

https://github.com/cpiber/mihon/blob/412815af067cdf343a6b5c7b5cd38eeb8190d543/app/src/main/java/eu/kanade/domain/track/interactor/SyncChapterProgressWithTrack.kt#L39

Because `tachiyomi.domain.chapter.model.Chapter` uses `Double`, but this line looks like it's using `Float` (just an observation by looking through github, need to check more thoroughly later)

In any case, I think it's valuable to include a small epsilon any time floating point numbers are compared, precision can get lost in so many places, so why force it?

cpiber:
@AntsyLich Please see my previous message, I'd rather avoid a PR poking around with types deep in Mihon that may have side-effects in other places, but I'll do it if you meant that

AntsyLich:
@Syer10 thoughts?

Syer10:
I think a quick fix is better for now. Just add a comment about why its there.

This will likely be fixed in general once a tracker rewrite is done to use the domain Tracker model

AntsyLich:
resolve conflict thanks

cpiber:
@AntsyLich done, thanks

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix mass migration not using the same search queries as individual migration ([@AntsyLich](https://github.com/AntsyLich)) ([#2736](https://github.com/mihonapp/mihon/pull/2736))
 - Fix reader not saving read duration when changing chapter ([@AntsyLich](https://github.com/AntsyLich), [@KotlinHero](https://github.com/KotlinHero)) ([#2784](https://github.com/mihonapp/mihon/pull/2784))
 
+### Other
+- Suwayomi tracker: Allow a small difference in chapter number for matching ([@cpiber](https://github.com/cpiber)) ([#2657](https://github.com/mihonapp/mihon/pull/2675))
```
AntsyLich: This should be under change given it changes a user facing behavior. Also include why it's being changed.

cpiber: @AntsyLich done

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 
 ### Changed
 - Update tracker icons ([@AntsyLich](https://github.com/AntsyLich)) ([#2773](https://github.com/mihonapp/mihon/pull/2773))
+- Suwayomi tracker: Allow a small difference in chapter number for matching to avoid incorrect chapter progress sync ([@cpiber](https://github.com/cpiber)) ([#2657](https://github.com/mihonapp/mihon/pull/2675))
```
```suggestion
- Add a small increment to chapter number before comparison to fix progress sync issues for Suwayomi ([@cpiber](https://github.com/cpiber)) ([#2657](https://github.com/mihonapp/mihon/pull/2675))
```

PR #2697, opened 2025-11-13
Fix shizuku installer not updating installed extensions

### LINKED ISSUE #2694: Shiuzuku installer option failed to install nor update
### Steps to reproduce

1. Start the Shizuku service via ADB/Wireless debugging or root (auto-start)
2. Check to make sure Shizuku is used as the install method (More > Settings > Advanced > Installer)
3. Go to Browse > Extensions and either install a new extension or if there's new update for some of them, tap the "Update all" option. 

### Expected behavior

Extension should be installed or updated

### Actual behavior

Installing new extensions are going to be stuck in "Downloading" or updating existing extensions doesn't install:

Example image of installing new extensions getting stuck in "Downloading" instead of "Installing"

Image

Example video of updating an existing extension that refuse to update (in frame is also Shizuku running in the background):

https://github.com/user-attachments/assets/2e07a72a-e7cb-4b8b-af9c-fa7fc2c4212f

### Crash logs

_No response_

### Mihon version

r7453

### Android version

Android 11 (phone), Android 13 (Emulator: WSA)

### Device

Realme 6i (phone), WSA 2407.40000.4.0 (emulator)

### Other details

Tried other options:

- PackageInstaller - Works. It prompts for the package installer (to install or update)

- Installing via ADB/Wireless debugging via a PC works as well (install log included):

```
PowerShell 7.5.4
PS C:\Users\User> adb connect <IP address and port number>
* daemon not running; starting now at tcp:5037
* daemon started successfully
connected to <IP address and port number>
PS C:\Users\User> adb install .\Downloads\APKs\<extension name>-<version>.apk
Performing Streamed Install
Success
PS C:\Users\User>
```

Note:
- WSA is rooted with Magisk meanwhile the phone doesn't have root, just used Wireless debugging for Shizuku to start.
- Mihon is already granted when it comes to Shizuku's app management screen:

Image

### Acknowledgements

NGB-Was-Taken:
Closes #2694

### DIFF
--- a/app/src/main/java/mihon/app/shizuku/ShellInterface.kt
+++ b/app/src/main/java/mihon/app/shizuku/ShellInterface.kt
@@ -111,14 +111,15 @@ class ShellInterface : IShellInterface.Stub() {
             .getMethod("openSession", Int::class.java)
             .invoke(packageInstaller, sessionId)
 
-        (
-            session::class.java.getMethod(
-                "openWrite",
-                String::class.java,
-                Long::class.java,
-                Long::class.java,
-            ).invoke(session, "extension", 0L, apk.length) as ParcelFileDescriptor
-            ).let { fd ->
+        session::class.java.getMethod(
+            "openWrite",
+            String::class.java,
+            Long::class.java,
+            Long::class.java,
+        )
+            .invoke(session, "extension", 0L, apk.length)
+            .let { it as ParcelFileDescriptor }
+            .let { fd ->
                 val revocable = Class.forName("android.os.SystemProperties")
                     .getMethod("getBoolean", String::class.java, Boolean::class.java)
                     .invoke(null, "fw.revocable_fd", false) as Boolean


app/src/main/java/mihon/app/shizuku/ShellInterface.kt
```diff
@@ -75,6 +75,12 @@ class ShellInterface : IShellInterface.Stub() {
             .invoke(pmInterface)
 
         val params = PackageInstaller.SessionParams(PackageInstaller.SessionParams.MODE_FULL_INSTALL).apply {
+            val installFlags = this::class.java.getField("installFlags")
+            installFlags.set(
+                this,
+                installFlags.getInt(this) or 0x00000002, // android.content.pm.PackageManager.INSTALL_REPLACE_EXISTING
```
MajorTanya: I assume this can't be done with the constant itself considering you commented it next to it?

NGB-Was-Taken: The constant is hidden from the public SDK by a `@hide` annotation in the source which you can see [here](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/content/pm/PackageManager.java;l=1682-1689). I guess I could replace it with a constant val if that would be more readable.

MajorTanya: > The constant is hidden from the public SDK by a `@hide` annotation

ew

> I guess I could replace it with a constant val if that would be more readable.

I think it would be beneficial. Add a comment with that link for future reference.

PR #2701, opened 2025-11-13
Update Kavita tracking to match current api & extension.

dear-clouds:
Previous code was not working with the last Kavita changes. 
This fixes issues with volumes, as well as the Comic library type items not being tracked/marked as read and counted correctly.


BrutuZ:
Does it still work with previous versions of the server/API?

dear-clouds:
Only obsolete versions would break, I believe, the api changed over a year ago. The issue with comics was there from the start. 

Still, just in case, non-existing types are treated as normal chapters, it should avoid breaking tracking.

PR #2728, opened 2025-11-20
Feature: Add per-manga reading time statistics with visualization in statistics view

marty4455:
## Description
Adds reading time statistics at global (statistics) and per-manga (in library) level, with toggle in statistics view to switch between scaling time by manga and total.

### Statistics screen
- New "Read Duration" section at the bottom, displaying top 30 most-read manga
- Shows manga cover thumbnails, bars, and reading time
- Items in list are clickable and navigate to the manga detail page (in library)

### Per-Manga statistics
- New "Reading Stats" option in manga detail page overflow menu
- Displays total reading time for the specific manga

### Customization Options
- **Cover Style**: Toggle between Square (compact) and Book (full) aspect ratios in Appearance settings (compact also reduces vertical padding)
- **Progress Bar Scale**: Inline toggle on stats screen to switch between:
  - **By Manga**: Bars relative to most-read manga (highlights differences between them)
  - **Total**: Bars as percentage of total reading time (shows distribution)

## Technical Implementation
- Added SQL queries: `getReadDurationByManga` (top 30) and `getReadDurationForManga` (single manga)
- Created new domain models: `ReadDurationByManga` with cover data integration
- Implemented interactors: `GetReadDurationByManga` and `GetReadDurationForManga`
- New UI components: `MangaReadTimeChart` (with segmented button switch) and `MangaReadingStatsDialog`
- Integrated with existing history tracking system

## Testing
Tested the views on my physical phone (S24 ultra)
- Verified stats displaying correctly with varying read times (under 1hr, above 1hr)
- Tested navigation from stats page to manga detail pages
- Verified cover style preferance persistance
- Tested clicking through without any reading stats

### Statistics Cover Style Setting
<table>
  <tr>
    <td>Cover style preference in settings</td>
  </tr>
</table>

### Statistics Screen
<table> Book
  <tr>
    <td>Stats screen</td>
    <td>Stats screen</td>
    <td>Stats screen</td>
  </tr>
</table>
<table> Compact
  <tr>
    <td>Stats screen - Square/compact view</td>
  </tr>
</table>

### Per-Manga Reading Stats
<table>
  <tr>
    <td>Manga detail - overflow menu with Reading Stats option</td>
    <td>Reading stats dialog showing total time</td>
  </tr>
</table>

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/manga/components/MangaReadingStatsDialog.kt
+++ b/app/src/main/java/eu/kanade/presentation/manga/components/MangaReadingStatsDialog.kt
@@ -8,16 +8,29 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.unit.dp
+import eu.kanade.presentation.util.toDurationString
 import tachiyomi.i18n.MR
 import tachiyomi.presentation.core.i18n.stringResource
+import kotlin.time.DurationUnit
+import kotlin.time.toDuration
 
 @Composable
 fun MangaReadingStatsDialog(
     totalReadDuration: Long,
     onDismissRequest: () -> Unit,
 ) {
+    val context = LocalContext.current
+    val none = stringResource(MR.strings.none)
+    val durationString = remember(totalReadDuration) {
+        totalReadDuration
+            .toDuration(DurationUnit.MILLISECONDS)
+            .toDurationString(context, fallback = none)
+    }
+
     AlertDialog(
         onDismissRequest = onDismissRequest,
         title = {
@@ -34,7 +47,7 @@ fun MangaReadingStatsDialog(
                     color = MaterialTheme.colorScheme.onSurfaceVariant,
                 )
                 Text(
-                    text = formatDuration(totalReadDuration),
+                    text = durationString,
                     style = MaterialTheme.typography.headlineMedium,
                     color = MaterialTheme.colorScheme.primary,
                 )
@@ -46,21 +59,4 @@ fun MangaReadingStatsDialog(
             }
         },
     )
-}
-
-private fun formatDuration(milliseconds: Long): String {
-    val minutes = milliseconds / 60000
-    return when {
-        minutes == 0L -> "Less than 1 minute"
-        minutes < 60 -> "$minutes minutes"
-        else -> {
-            val hours = minutes / 60
-            val remainingMinutes = minutes % 60
-            when {
-                remainingMinutes == 0L -> "$hours hours"
-                hours == 1L -> "1 hour $remainingMinutes minutes"
-                else -> "$hours hours $remainingMinutes minutes"
-            }
-        }
-    }
 }
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/stats/MangaReadTimeChart.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/stats/MangaReadTimeChart.kt
@@ -17,19 +17,25 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
 import eu.kanade.domain.ui.UiPreferences
 import eu.kanade.domain.ui.model.StatsCoverStyle
 import eu.kanade.domain.ui.model.StatsProgressBarStyle
 import eu.kanade.presentation.manga.components.MangaCover
+import eu.kanade.presentation.util.toDurationString
 import tachiyomi.domain.history.model.ReadDurationByManga
 import tachiyomi.domain.manga.model.MangaCover as MangaCoverData
+import tachiyomi.i18n.MR
 import tachiyomi.presentation.core.components.material.padding
+import tachiyomi.presentation.core.i18n.stringResource
 import tachiyomi.presentation.core.util.collectAsState
 import tachiyomi.presentation.core.util.secondaryItemAlpha
 import uy.kohesive.injekt.Injekt
 import uy.kohesive.injekt.api.get
+import kotlin.time.DurationUnit
+import kotlin.time.toDuration
 
 @Composable
 fun MangaReadTimeChart(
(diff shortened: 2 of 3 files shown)


KraXen72:
Can you please provide a test apk? I'd like to try this feature out! thanks! :)

marty4455:
@KraXen72 I've added a debug APK for you. Note that it's based on the codebase from November so it's not fully up to date with current main. You'll need to enable "Install from unknown sources" on your device to install it.
https://github.com/marty4455/mihon/releases/tag/v1

KraXen72:
thanks, @marty4455. I have tried it out, it's a nice insight. I have noticed, these issues:
- due to some migrations, some manga entries are duplicated. If this is to be expected, feel free to keep it, otherwise, some form of de-duplication, at least a lowercase + trim + match could be employed to merge the statistics
- since i've read manga from missing (obsolete/uninstalled/defunct) sources, they are missing pictures and when I click on it, i only get some information that's still left in db. I don't really know what a good solution for this would be, to be honest.

Screenshot_20260223_172402_Mihon

marty4455:
Thanks @KraXen72, I'll have a look at it

marty4455:
@KraXen72 I've changed it to merge duplicate entries and sum the read times.
I think the missing source is intended behavior. It correctly shows a warning icon and displays the stored info. Or at least I also don't know a better way to handle this.
Do you want to re-check it with the new debug APK?
https://github.com/marty4455/mihon/releases/tag/v2

KraXen72:
i'll test it later, thx

KraXen72:
It's better now, however, previously, I had three entries for greatest state developer, and one of them was valid in my library, that is, it was from a source which still works. However, after the merging, if I tap on the top item now, it will show me one of the unresolved entries. I think you should make it so that when at least one of the different entries which are being merged is valid, it should always prefer to open that one instead of an arbitrary one.

marty4455:
@KraXen72 oh i pick sorted by: favorite -> cover url not null -> total read time

Maybe then the cover url isn't null but just doesn't point to a valid url... i had no example for this on my device so assumed it worked.
I'll try again, thanks for testing!!

marty4455:
@KraXen72 Okay, I now use the sourcemanager to try to get the entry in hashmap by sourceid, which should give the best result as "main" manga which the others should be merged into.
If there is a broken image and it has been migrated already, it's probably because you did not yet read in this new source. I just saw it happen on my device on one piece, but after i opened the reader in the new source, it correctly added it to the history db so it changed the "main" entry and correctly displays the image now
Also changed the "read time" button on the manga screen since that used mangaId, it now also uses title to search.
I see one manga where the title has been changed on a different source, which is now duplicated, but there isn't much to do there i think
I had a max of 50 manga loaded from the db for merging statistics (then took 30 in view), but if you just migrated, the new "good" manga source would have read time 1 min or less, so it wouldn't be included. I just changed the limit to 500 now and doesn't seem to really be causing performance issues on the large manga db of a friend (i only had 65).

Do you want to try the new version? thx :)
https://github.com/marty4455/mihon/releases/tag/v3

KraXen72:
Hi, i tried it. this time, only one entry failed to resolve, even though i have it in the libary (kaguya sama). it's likely due to it not having any read time, since i read it a few years ago and migrated it after that source died.

My suggestion for the main manga when merging would be:
1. it should try to find any entry which is currently in the library (by title), take one with the most read time
2. if there is no such entry, use the one in the database with non-null cover image, sorted by  read time desc



i'm not quite sure what the by manga/total toggles do -- If it's as I assume, and the progress bars in the total should add up for all manga into one full progress bar, while the "by manga" progress bars take the most read manga as a full progress bar -- I'm not sure the total mode is even needed, I don't find it that intuitive, but I'll leave that up to you.

I have one more suggestion and that is remove the card/container, which wraps the manga entries. at least with the default color scheme,  (which may be just my monet theme), It isn't really visible, and it feels unnecessary. it makes all of the list items for the manga narrower (by the padding on each side).

I'm also uploading my backup, to help you test quicker. I didn't include any sensitive settings and I don't really care if people see what I read. The link expires in three days, so please download it before then:

https://send.vis.ee/download/e713e972efe98922/#7T4QekNviwhIfZZuZNc81g

---

Also, mihon really needs a "trust all sources" or trust by extension repo system (like in az). it's annoying to trust all exts one by one when reinstalling. (i reinstalled the test apks you sent, didn't want to make assumptions about whether you bumped version nums for debug builds)

marty4455:
@KraXen72 Yeah, these are fetched from the history database, so if you have not opened the new source in the reader, there is no entry and will not be selected.
By manga compares the manga to each other, based off of the one with the highest read time.
Total compares to the total hours read (all manga combined). This was requested by a few ppl who said it was useful/fun, I also thought it wasn't that useful when you have a lot of manga since the bars just get small, but oh well
You can change the card cover style in settings -> appearance -> statistics cover style -> change to book

thanks!

KraXen72:
I'm not talking about the style of the items for each manga. I think that's fine, whether configurable or not. I just think the wrapper around them (the card) is unnecessary.
I would make it so that the manga items are exactly as wide as the different cards above, in the statistics.
But I'll leave that up to you.

marty4455:
Ohh I see.
I removed the card:
Screenshot_20260225_153204_Mihon

I don't really have a preference. I guess without is more real estate

app/src/main/java/eu/kanade/presentation/manga/components/MangaReadingStatsDialog.kt
```diff
@@ -0,0 +1,66 @@
...
+        minutes < 60 -> "$minutes minutes"
+        else -> {
+            val hours = minutes / 60
+            val remainingMinutes = minutes % 60
+            when {
+                remainingMinutes == 0L -> "$hours hours"
+                hours == 1L -> "1 hour $remainingMinutes minutes"
+                else -> "$hours hours $remainingMinutes minutes"
+            }
+        }
```
MajorTanya: These must not be hardcoded. Use proper string resources or it'll be English in every app language.

app/src/main/java/eu/kanade/tachiyomi/ui/stats/MangaReadTimeChart.kt
```diff
@@ -0,0 +1,155 @@
...
+private fun formatDuration(milliseconds: Long): String {
+    val minutes = milliseconds / 60000
+    return when {
+        minutes < 60 -> "${minutes}m"
+        else -> {
+            val hours = minutes / 60
+            val remainingMinutes = minutes % 60
+            if (remainingMinutes == 0L) "${hours}h"
+            else "${hours}h ${remainingMinutes}m"
+        }
```
MajorTanya: This as well cannot be hardcoded for i18n purposes.

MajorTanya (commented):
Just saw these while exploring the code.

app/src/main/java/eu/kanade/presentation/manga/components/MangaReadingStatsDialog.kt
marty4455: Thanks! Replaced with toDurationString() using LocalContext, which takes sec/min/hr/days from resources (my method only went up to hr)

app/src/main/java/eu/kanade/tachiyomi/ui/stats/MangaReadTimeChart.kt
marty4455: Thanks! Replaced with toDurationString() using LocalContext, which takes sec/min/hr/days from resources (my method only went up to hr)

CarrotRub:
Any news on this PR ?

Seems pretty interesting and promising !

PR #2739, opened 2025-11-24
press and hold for favoriting on global search

Dorumin:
Title. Most of the code is straight from BrowseSourceScreen, which means there's a lot of duplication between those models now. Should probably lift those values up into SearchScreenModel and inherit from that on BrowseSourceScreenModel, but I just wanted to get it working.

I don't think there's an issue open for this, and I never touched java or kotlin (that much should be self evident), so I didn't put much time into cleaning up the code unless there's intent to merge.

Feedback?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,13 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - `Other` - for technical stuff.
 
 ## [Unreleased]
+### Added
+- Make long pressing global search results favorite the entry ([@Dorumin](https://github.com/Dorumin)) ([#2739](https://github.com/mihonapp/mihon/pull/2739))
+
 ### Fixed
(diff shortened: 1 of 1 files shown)


app/src/main/java/eu/kanade/tachiyomi/ui/browse/source/browse/BrowseSourceScreen.kt
```diff
@@ -105,7 +105,7 @@ data class BrowseSourceScreen(
         val haptic = LocalHapticFeedback.current
         val uriHandler = LocalUriHandler.current
         val snackbarHostState = remember { SnackbarHostState() }
-
+
```
cuong-tran: Always run `./gradlew spotlessApply` before committing to avoid this blank indent.

Dorumin: Applied

AntsyLich:
Add a changelog entry

Dorumin:
Appended it under Fixed

AntsyLich:
this isn't fixing anything... should be under added

Dorumin:
Very well. I personally saw it as making behavior more consistent between source results and the global screen

PR #2753, opened 2025-12-02
Add an option to restart the app without extensions in the crash screen

### LINKED ISSUE #2445: Safe mode to troubleshoot extensions from "Private" installer

Originally https://github.com/tachiyomiorg/tachiyomi/issues/10389 ([archive](https://git.mihon.tech/tachiyomi/tachiyomi/issues/10389))

If an extension installed in the system crashes the app, you can uninstall it without starting the app. We need a similar feature to get rid of problematic privately-installed extensions.

A possible way is to provide a separate "uninstall private extensions" screen that can be opened with a button in the crash handler screen. Just in case any uncaught exceptions might crash the app directly, we can also export this activity and add it to the launcher shortcuts, maybe replace "Library" as it's already the default screen.

### Other details

_No response_

### Acknowledgements

NGB-Was-Taken:
Also adds a shortcut to do the same (replaces the library shortcut).

| Image 1 | Image 2 |
| ------- | ------- |
| image | image |

Need someone to design the shortcut icon. ~~I'm too lazy~~
Closes #2445


presentation-core/src/main/java/tachiyomi/presentation/core/screens/InfoScreen.kt
```diff
@@ -42,6 +42,8 @@ fun InfoScreen(
     canAccept: Boolean = true,
     rejectText: String? = null,
     onRejectClick: (() -> Unit)? = null,
+    thirdButtonText: String? = null,
+    onThirdButtonClick: (() -> Unit)? = null,
```
AntsyLich: Find a better name for this

i18n/src/commonMain/moko-resources/base/strings.xml
```diff
@@ -888,6 +889,7 @@
     <string name="crash_screen_title">Whoops!</string>
     <string name="crash_screen_description">%s ran into an unexpected error. We suggest you share the crash logs in our support channel on Discord.</string>
     <string name="crash_screen_restart_application">Restart the application</string>
+    <string name="crash_screen_restart_safe_mode">Restart the application without loading extensions</string>
```
AntsyLich:
```suggestion
    <string name="crash_screen_restart_safe_mode">Restart in safe mode</string>
```

nnnn20430:
Does this let you reset the settings of the extensions? If not you should add it, cause I'm currently having a problem where a setting is causing an extensions and the app to crash, and only way to make it stop is uninstalling the extension apk from system. And if I attempt to reinstall the extension it just crashes again, no way to fix the broken setting. So the extension is soft bricked for me.

NGB-Was-Taken:
> Does this let you reset the settings of the extensions? If not you should add it

No it does not, and no it will not be added. This is built with malicious actors in mind and avoids execution of any extension provided code, which includes getting the ID that is needed for accessing its preferences. 

> I'm currently having a problem where a setting is causing an extensions and the app to crash

If the problem can be isolated to a single preference, there are probably ways to get around the crash to change that preference. Safe mode is not intended for this. This is a last resort option meant to remove a totally broken or malicious extension, similar to going into **Settings > Installed apps** and uninstalling the problematic extension there.

app/src/main/java/eu/kanade/presentation/browse/ExtensionsScreen.kt
```diff
@@ -209,6 +209,7 @@ private fun ExtensionContent(
                         is Extension.Untrusted -> "extension-untrusted-${item.hashCode()}"
                         is Extension.Installed -> "extension-installed-${item.hashCode()}"
                         is Extension.Available -> "extension-available-${item.hashCode()}"
+                        is Extension.NotLoaded -> "extension-notLoaded-${item.hashCode()}"
```
AntsyLich:
```suggestion
                        is Extension.NotLoaded -> "extension-not-loaded-${item.hashCode()}"
```

app/src/main/java/eu/kanade/presentation/more/MoreScreen.kt
```diff
@@ -70,6 +73,16 @@ fun MoreScreen(
...
 
+            item {
+                SwitchPreferenceWidget(
+                    title = stringResource(MR.strings.label_safe_mode),
+                    subtitle = stringResource(MR.strings.safe_mode_summary),
+                    icon = Icons.Outlined.Shield,
+                    checked = safeMode,
+                    onCheckedChanged = onSafeModeChange,
+                )
+            }
```
AntsyLich: I'm not sure about having this option here. Maybe show only when it's enabled?

app/src/main/java/eu/kanade/presentation/crash/CrashScreen.kt
```diff
@@ -25,7 +25,7 @@ import tachiyomi.presentation.core.screens.InfoScreen
 @Composable
 fun CrashScreen(
     exception: Throwable?,
-    onRestartClick: () -> Unit,
+    onRestartClick: (Boolean) -> Unit,
```
AntsyLich: Add a separate `onSafeModeClick`

AntsyLich (changes requested):

app/src/main/java/eu/kanade/presentation/more/MoreScreen.kt
NGB-Was-Taken: I did think about simply making it not persist across app restarts, but did not like the idea at that moment, not sure why. Should the safe mode be non-persistent, or should we just to hide the toggle when it is disabled?

AntsyLich (changes requested):
NotLoaded extensions should have a reason attached to it (for this pr's case "Due to safe mode" or something along those lines). Tapping on a not loaded extension should show the reason in a dialog and also use the warning icon.

NGB-Was-Taken:
> NotLoaded extensions should have a reason attached to it (for this pr's case "Due to safe mode" or something along those lines). Tapping on a not loaded extension should show the reason in a dialog and also use the warning icon.

image

Also, there's this issue currently where if the app is not closed before launching it in safe mode (from the shortcut), the extensions stay loaded. Not sure what we can do in this situation. Maybe launch a different activity that kills and relaunches the main activity?

domain/src/main/java/eu/kanade/tachiyomi/extension/model/Extension.kt
```diff
@@ -71,5 +74,22 @@ sealed class Extension {
         val signatureHash: String,
         override val lang: String? = null,
         override val isNsfw: Boolean = false,
+        override val isShared: Boolean,
     ) : Extension()
+
+    data class NotLoaded(
```
AntsyLich, line 80:
```suggestion
    data class Detected(
```
thoughts?

NGB-Was-Taken: I don't think `Detected` properly conveys the state of the extension. It sounds more like it would be a parent class to everything except `Available`

AntsyLich: What if we do
```
AVAILABLE   // Can be downloaded
INSTALLED   // Present locally
LOADED      // Successfully loaded
```
doesn't have to be part of this PR

NGB-Was-Taken: Still no. `Detected` describes a part of the loading process (discovery), whereas `NotLoaded` is a state. If you dislike that it is a combination of two words, `Unloaded` may be a choice but even that implies that it was loaded at some point and then unloaded.

AntsyLich: ...i didn't mention detect the second time

NGB-Was-Taken: I thought that was implied..? You originally suggested renaming `NotLoaded`, I said I wasn't convinced and then you suggested renaming the others instead. Since it was continued in the same thread, to me it seemed like the context was still about `NotLoaded` and `Detected`.

AntsyLich: Yeah just ignore Untrusted I'll merge it in the current NotLoaded. Otherwise thoughts on the naming of the other 3?

NGB-Was-Taken: Sure. I don't have any objections to those. 

Should the Untrusted and NotLoaded merge be done in this PR though?

AntsyLich: no

AntsyLich:
> Also, there's this issue currently where if the app is not closed before launching it in safe mode (from the shortcut), the extensions stay loaded. Not sure what we can do in this situation. Maybe launch a different activity that kills and relaunches the main activity?
>
rerun extension loader when safe mode value changes

AntsyLich:
rebase thanks

Issue #2738, opened 2025-11-24
Support full display of manga chapter names


WeMingT:
### Describe your suggested feature

I found that it seems impossible to fully display manga chapter names, which is unfriendly for longer titles, such as in the following case (plugin name: `Hanime1.me`, manga title: `[H9] Kanojo ni Kokuhaku Suru Mae ni Tomodachi ni Nakadashi Sareta... Ch. 0-15 Collection [Chinese] [Digital]`).  

![Image](https://github.com/user-attachments/assets/6c1b29ea-abdb-4eed-828c-02798c2e8e9f)

I hope the full manga chapter name can be displayed completely.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.19.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

NathanBap:
Hi, i'm working on this issue

PR #2754, opened 2025-12-02
Add expandable chapter titles option

NathanBap:
#2738 Adds the option to expand long chapter names : 
- Added a checkbox to enable or disable the expansion of long chapter titles. When it's not checked, it's like the usual and when it's checked, the title expands on a new line.
- This option is accessible by doing the following : 
    - Click on the 3 lines on the top right of a manga chapters page
    - Got to the "Display" tab
    - The option is called "Expand chapter titles"

Capture d’écran 2025-12-02 à 12 18 43
Capture d’écran 2025-12-02 à 12 19 01

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix shizuku installer not updating installed extensions ([@NGB-Was-Taken](https://github.com/NGB-Was-Taken)) ([#2697](https://github.com/mihonapp/mihon/pull/2697))
 - Fix mass migration not using the same search queries as individual migration ([@AntsyLich](https://github.com/AntsyLich)) ([#2736](https://github.com/mihonapp/mihon/pull/2736))
 
+### Added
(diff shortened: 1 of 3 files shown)


AntsyLich:
The new option should be a switch and the other 2 option needs to be grouped under a header

AntsyLich:
and add a changelog entry

BrutuZ:
Would it look more consistent if it was a marquee (sliding text) instead of multi-line with toggle?

NathanBap:
> Would it look more consistent if it was a marquee (sliding text) instead of multi-line with toggle?

I can try to add another option for that. I'm working on the first comment and then i'll look into it.

NathanBap:
@AntsyLich I changed the checkbox to a switch and added a changelog entry as requested.

Capture d’écran 2025-12-08 à 11 05 31

AntsyLich:
Ungrouped options should go above grouped stuff

NathanBap:
@AntsyLich It's done.
Capture d’écran 2025-12-22 à 15 33 48

PR #2765, opened 2025-12-05
make the in‑chapter page preload amount configurable and exposes it as a user setting in the reader settings

okami-horo:
This PR makes the in‑chapter page preload amount configurable and exposes it as a user setting
  in the reader settings. Previously, HttpPageLoader always preloaded 4 pages ahead; now users can
  choose how many pages to preload, including disabling it entirely.
Screenshot_20251205-173015


AntsyLich:
Superseded by #2986

PR #2773, opened 2025-12-07
Update tracker icons

AntsyLich:
| | Before | After |
| --- | --- | --- |
| Light |  |  |
| Dark |  |  |


MajorTanya:
Can we have "before" screenshots as well for each theme?

Also, where did you get the SVGs from (are they all official SVGs or did you recreate them)?

AntsyLich:
> Can we have "before" screenshots as well for each theme?
> 
Added

> Also, where did you get the SVGs from (are they all official SVGs or did you recreate them)?
>
All of them are based on Offcial SVGs. I slightly tweaked each of them to make them look nicer in Mihon

MajorTanya:
Not sure if I'm a fan of the white background Shikimori logo but that is their favicon after all...

AntsyLich:
Shikimori and Bangumi now uses dark background

PR #2779, opened 2025-12-10
Fix display of pre-1970 dates

### LINKED ISSUE #2777: Chapter dates with Komga extension before 1970 show as N/A
### Steps to reproduce

1. Connect to Komga server
2. Go to series with comics with metadata that says they released before January 1st 1970
3. See the date on Mihon under the chapter title

### Expected behavior

The date should show as the release date that is set in Komga

### Actual behavior

Date is shown as N/A (Strangely, for issue 80 here which has a release date of January 1st 1970, it shows as February 2nd 2014)

Image

Image

### Crash logs

_No response_

### Mihon version

0.19.3

### Android version

Android 16

### Device

Google Pixel 9a

### Other details

The only other comic book media server I know of, Kavita, shows dates in Mihon as the upload date to the server, not the release date stored in the metadata of a given comic. For this reason, I am not sure if this error is caused by Mihon being unable to parse a negative unix epoch or if this is because the Komga extension or server is where the problem lies.

### Acknowledgements

MajorTanya:
A user in #2777 was using the ComicInfo.xml Year/Month/Day fields to indicate date of publication for some American comics, which often predate the UNIX Epoch of 1970.

They were seeing "N/A" displays because this line of code discarded date information for any time before Jan 1st, 1970.

The `toRelativeString` extension function used in the other `relativeDateText` function already accounts for very distant dates (anything >7 days away turns into full date, not relative, regardless of setting, though disabling the relative timestamp setting would always show the full date). ~~Removing this line should not cause any issues as it is purely a display difference and the use case of backdating comics to pre-1970 is worth it in my opinion.~~

Edit: I chose to instead replace the condition with a more strict `!= 0L` since 0L serves as the sentinel value for "now"/"today". This means anyone who dates their chapter to January 1st, 1970 will see "Today" or the current date (current behaviour) since Mihon does not take time into consideration for the display. But anyone with a date other than this in UTC should be able to see proper dates in the past beyond 1970.

Would close #2777.

| Before | After (with the yyyy-MM-dd date format selected for clarity) |
|-|-|
|image|image|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix shizuku installer not updating installed extensions ([@NGB-Was-Taken](https://github.com/NGB-Was-Taken)) ([#2697](https://github.com/mihonapp/mihon/pull/2697))
 - Fix mass migration not using the same search queries as individual migration ([@AntsyLich](https://github.com/AntsyLich)) ([#2736](https://github.com/mihonapp/mihon/pull/2736))
 - Fix reader not saving read duration when changing chapter ([@AntsyLich](https://github.com/AntsyLich), [@KotlinHero](https://github.com/KotlinHero)) ([#2784](https://github.com/mihonapp/mihon/pull/2784))
-- Fix display of pre-1970 dates in chapter list ([@MajorTanya](https://github.com/MajorTanya)) ([#2779](https://github.com/mihonapp/mihon/pull/2779))
+- Fix pre-1970 upload date display in chapter list ([@MajorTanya](https://github.com/MajorTanya)) ([#2779](https://github.com/mihonapp/mihon/pull/2779))
 
 ## [v0.19.3] - 2025-11-07
 ### Improved


MajorTanya:
No changelog yet because I don't know if I missed some edge case where this check was actually needed.

For what it's worth, I tried to track down the origin of this check and lost track of it somewhere in 2022, so it's been here a very long time, even surviving the removal and re-adding of the relative timestamp setting back in Tachiyomi days.

BrutuZ:
Would this affect the behavior of `0L` and `-1L` becoming "date.now"?

MajorTanya:
> Would this affect the behavior of `0L` and `-1L` becoming "date.now"?

Good question, I didn't remember that behaviour.

MajorTanya:
In a really, really dirty way, we could keep this (annoying) behaviour if instead of removing the takeIf like I did, we constrain it a lot more to something like
```diff
- .takeIf {dateEpochMillis > 0L }
+ .takeIf { dateEpochMillis != 0L || dateEpochMillis != -1L }
```
this way we could keep the current behaviour without massively inconveniencing people and sources which rely on this behaviour (which, again, is annoying). It would technically make it impossible to date something January of 1970 without a day, or _exactly_ January 1st, 1970, but it _should_ allow any other predating like the user is doing.

Edit:
Or, to be extra silly, why not
```diff
- .takeIf { dateEpochMillis > 0L }
+ .takeIf { dateEpochMillis !in -1L..0L }
```

Unless someone can definitively tell me whether -1L has any special interpretation or not, in which case just simple equality to 0L would be enough

Edit 2:
fixed inverse logic, oops

BrutuZ:
1L would be 1970-01-01 00:00:01 (or even :001 depending on the accuracy it uses), which is technically still the same day :clueless:

MajorTanya:
> 1L would be 1970-01-01 00:00:01 (or even :001 depending on the accuracy it uses), which is technically still the same day :clueless:

ComicInfo.xml spec only allows day-level resolution so

Accuracy is milliseconds for some reason

BrutuZ:
> Unless someone can definitively tell me whether -1L has any special interpretation or not, in which case just simple equality to 0L would be enough

Brainfart on my end, I seem to have saved the defaults for chapter date and chapter number in the same neuron
https://github.com/mihonapp/mihon/blob/876c3f951b7e9782054d8f788ab39772ae6cf440/source-api/src/commonMain/kotlin/eu/kanade/tachiyomi/source/model/SChapterImpl.kt#L11-L13

MajorTanya:
Changed it to the `!= 0L` approach now which should be working better.

Edit: changed PR title as this has nothing to do with relative timestamps, contrary to my original assumption.

AntsyLich:
> Edit: I chose to instead replace the condition with a more strict `!= 0L` since 0L serves as the sentinel value for "now"/"today". This means anyone who dates their chapter to January 1st, 1970 will see "Today" or the current date (current behaviour) since Mihon does not take time into consideration for the display. But anyone with a date other than this in UTC should be able to see proper dates in the past beyond 1970.
> 

Shouldn't that only be the case until we put the chapter in database?

MajorTanya:
> > Edit: I chose to instead replace the condition with a more strict `!= 0L` since 0L serves as the sentinel value for "now"/"today". This means anyone who dates their chapter to January 1st, 1970 will see "Today" or the current date (current behaviour) since Mihon does not take time into consideration for the display. But anyone with a date other than this in UTC should be able to see proper dates in the past beyond 1970.
> 
> Shouldn't that only be the case until we put the chapter in database?

I saw this for my testing chapters even after adding them to the library and such. It was stuck with the current date for publishing when it had the sentinel 0L value.

MajorTanya:
Should I add a changelog entry

AntsyLich:
yes

MajorTanya:
damn I forgot how far behind my branch was oops

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix shizuku installer not updating installed extensions ([@NGB-Was-Taken](https://github.com/NGB-Was-Taken)) ([#2697](https://github.com/mihonapp/mihon/pull/2697))
 - Fix mass migration not using the same search queries as individual migration ([@AntsyLich](https://github.com/AntsyLich)) ([#2736](https://github.com/mihonapp/mihon/pull/2736))
 - Fix reader not saving read duration when changing chapter ([@AntsyLich](https://github.com/AntsyLich), [@KotlinHero](https://github.com/KotlinHero)) ([#2784](https://github.com/mihonapp/mihon/pull/2784))
+- Fix display of pre-1970 dates in chapter list ([@MajorTanya](https://github.com/MajorTanya)) ([#2779](https://github.com/mihonapp/mihon/pull/2779))
```
AntsyLich: Maybe?
```suggestion
- Fix pre-epoch upload date display in chapter list ([@MajorTanya](https://github.com/MajorTanya)) ([#2779](https://github.com/mihonapp/mihon/pull/2779))
```

MajorTanya: I would rather keep the 1970 explicit here but we can keep the rest of your wording. I want to keep it because not everyone might know what "pre-epoch" means and the 1970 part is very obvious.

AntsyLich: sure

MajorTanya: ~~nevermind this comment lol I tried to cheat with the suggestion thing~~

PR #2802, opened 2025-12-19
Show source chapter names in reading history

### LINKED ISSUE #994: Use the source names for chapters in History

Mihon should use the chapter names used by sources when displaying stuff you’ve read in History. That would mitigate issues where you may have read something that is in the same manga/comic but outside of the main continuity of the series, and thus uses its own name and numbering system.

An example is “Annuals”. Various comics have them, but the app makes it seem like you’ve read the third *issue* instead of the third annual, since both are displayed as “Ch.”.

### Other details

None.

### Acknowledgements

MuhamadSyabitHidayattulloh:
## Description

This PR improves the reading history by displaying source-provided chapter names while ensuring the read time is always visible, even when chapter titles are long.

---

### Problem 1: Generic chapter titles in history
The history screen previously relied on generic chapter numbering (e.g. “Ch. 3”), making it difficult to distinguish between regular chapters, annuals, specials, or spin-offs that use different naming schemes.

---

### Problem 2: Long chapter titles can hide the read time
After switching to source chapter names, long titles could push the read time out of view, making it harder to quickly see when an entry was read.

---

#### Source chapter names
- History now prefers the original chapter name provided by the source (extension).
- Falls back safely to the existing generic chapter number format when the chapter name is missing.
- No database schema or view changes are required.

#### Improved subtitle layout
- The history subtitle layout has been updated to **always show the read time first**, followed by the chapter name.
- A dot separator (`•`) is used for consistency with other lists.
- Long chapter names are ellipsized, ensuring the time remains visible.

**Format logic:**
- `Time • Chapter name`
- If only one value exists, it is shown alone.

---

### Result (Screenshot)

History shows read time before source chapter name

---

### Impact
- Reading history is clearer and more informative.
- Read time is never hidden by long chapter titles.
- No additional settings or configuration required.
- Changes are isolated to history queries and UI layout with no side effects.

---

### Affected Areas
- History database queries
- History item UI layout

---

### Testing
- History entries display source chapter names correctly
- Read time remains visible with long chapter titles
- Generic chapter numbering still appears as fallback
- No crashes or database schema issues

Closes #994

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/history/components/HistoryItem.kt
+++ b/app/src/main/java/eu/kanade/presentation/history/components/HistoryItem.kt
@@ -12,6 +12,7 @@ import androidx.compose.material.icons.outlined.FavoriteBorder
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.ProvideTextStyle
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


BrutuZ:
Wouldn't it be safer to invert the order of elements in that format (or both for consistency)? A long title could hide the time

Could also be a setting, like for titles _not_ prefixed with the number or because it's easy enough to swap between displaying titles or numbers on the chapters list

PR #2807, opened 2025-12-21
Discard the volume component earlier when parsing chapter number (#2492)

### LINKED ISSUE #2492: false positive for "Missing chapters" for bonus / side story chapters
### Steps to reproduce

As the title says, when there's a bonus / side story chapter the missing chapters detection gives a false positive, even if there's nothing missing according to the regular numbering.

### Expected behavior

No missing chapters gap should be detected since the numbering increment is not broken

### Actual behavior

Image

The "volume 10 bonus chapter" is being detected as chapter 10 overall, so the app is expecting a chapter 11.

### Crash logs

_No response_

### Mihon version

0.19.1

### Android version

15

### Device

galaxy s21 fe

### Other details

Also reproduced this with Komikku in a tablet (the screenshot is from that one, I installed Mihon 0.19.1 in my phone to verify and it does the same)

### Acknowledgements

nobbele:
Now parses the chapter correctly instead of 60, 10, 61

image

Closes #2492

### DIFF
--- a/domain/src/main/java/tachiyomi/domain/chapter/service/ChapterRecognition.kt
+++ b/domain/src/main/java/tachiyomi/domain/chapter/service/ChapterRecognition.kt
@@ -22,7 +22,13 @@ object ChapterRecognition {
      * Regex used to remove unwanted tags
      * Example Prison School 12 v.1 vol004 version1243 volume64 -R> Prison School 12
      */
-    private val unwanted = Regex("""\b(?:v|ver|vol|version|volume|season|s)[^a-z]?[0-9]+""")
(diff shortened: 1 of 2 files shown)


BrutuZ:
What if all you have are volumes (tankoubons or western media)? They'll all be identified as Chapter 1, breaking the Duplicate Chapter features (Skip, Mark as Read)

nobbele:
Right.. maybe if you could base it off of the previous/other chapter titles perhaps.. not sure what the right solution would be

AntsyLich:
next major release will have volume support we can revisit this then

PR #2814, opened 2025-12-28
Add Select All / Deselect All button and selection count to Migration screen

aSimpleGuy:
This PR adds a Select All / Deselect All button in the migration screen AppBar 
and updates the title to show how many manga are selected.


MajorTanya:
It is intentional that this screen doesn't have those options. It is batch migration, not mass migration. You're supposed to pick a smaller amount at once, migrate those, then continue with the next batch afterwards. It's a compromise between users who kept asking for this and third-party sources which incur a lot of traffic from this procedure in the first place.

I'll leave it to @AntsyLich to decide. 

I do like the display of how many were selected though.

(and as a general piece of feedback, independently of this specific PR: emojis have no place in code comments & those toasts are redundant because users would see the selection results)

AntsyLich:
1. I don't want AI slop
2. I have others plan for mass selection

PR #2828, opened 2026-01-05
Filter Scanlator by priority

### LINKED ISSUE #777: Filter scanlators by priority

Copied from scb261 for TACHIYOMI
TLDR: User can set up which scanlators he prefers over others. If first priority is not available, it would fallback to the second. If not, than third. And so on.

For example there is a manga with following chapters:

ch. 1, group A
ch. 2, group A
ch. 2, group B
ch. 2, group C
ch. 3, group B
Current implementation ([#8803](https://git.mihon.tech/tachiyomi/tachiyomi/issues/8803)) allows to filter out only whole groups from manga, so user would end up without chapter 1 or without chapter 3.

With priorities the user can select that he prefers groups in specific order, for example A > B > C. Then the filter would show each chapter only once like this:

ch. 1, group A
ch. 2, group A
ch. 3, group B

### Other details

_No response_

### Acknowledgements

### LINKED ISSUE #1001: Hide entries whose scanlators are excluded from updates tab

As the title says, if I have a manga where I've excluded some scanlators and if new chapter with those scanlators appears, then they should not pop up in the updates tab, since I don't care about those scanlators.

### Other details

Trying to click on an entry whose scanlator is excluded in filters from the updates tab will result in "Requested chapter of id ... not found in chapter list"

### Acknowledgements

AwkwardPeak7:
closes #777
closes #1001

Implement Scanlator priority system. 
Users can define a priority list of scanlators and chapter list will be deduplicated according to it. `priority = 0` is highest, and it decreases with increasing number.
It also still supports excluding a scanlator altogether, their chapters will be hidden regardless. `priority = -1`

I have tested and confirmed working, db migration from old version, backup restore from old backup.

[Screencast_20260105_150143.webm](https://github.com/user-attachments/assets/55f5da53-95a2-4320-a40e-cc1c6bb7afee)

**Disclaimer: Ai assisted**

### DIFF
--- a/app/src/main/java/eu/kanade/domain/chapter/interactor/SyncChaptersWithSource.kt
+++ b/app/src/main/java/eu/kanade/domain/chapter/interactor/SyncChaptersWithSource.kt
@@ -233,14 +233,14 @@ class SyncChaptersWithSource(
         val updatedFilters = if (currentFilters.isNotEmpty()) {
             val validFilters = currentFilters.filter { (it.scanlator ?: "") in currentScanlators }
             val existingScanlators = validFilters.map { it.scanlator ?: "" }.toSet()
-            
+
             val allChapters = dbChapters + updatedToAdd
             val scanlatorStats = allChapters.groupBy { it.scanlator }
-            
+
             val newScanlators = currentScanlators.minus(existingScanlators)
                 .sortedWith(
                     compareByDescending<String> { scanlatorStats[it]?.size ?: 0 }
-                        .thenByDescending { scanlatorStats[it]?.maxOfOrNull { c -> c.dateUpload } ?: 0 }
+                        .thenByDescending { scanlatorStats[it]?.maxOfOrNull { c -> c.dateUpload } ?: 0 },
                 )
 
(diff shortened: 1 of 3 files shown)


BrutuZ:
Very nice, but this removes the Multi Selection QoL.

When I mused about this before, instead of individual toggles it would have an extra sortable item as a "visibility threshold", everything below that item would be hidden

AwkwardPeak7:
> Multi Selection QoL.

99% of the time, you dont even need to hide any scanlator. Chapter list gets deduplicated based on the priority. 

Exclusion imo is only useful in cases where a sniper uploads a very far along chapter.

BrutuZ:
Suppose it's an aggregator and I only want to keep Official releases, which are almost always behind scanlators

AwkwardPeak7:
@BrutuZ tap and hold on eye icon to exclude all others, or unexclude if all others are excluded. looks good?

https://github.com/user-attachments/assets/1172f4d0-aa4e-4fcd-bd2d-c1579749f182

BrutuZ:
Fair compromise, albeit less intuitive.
It addresses the issue I anticipated, just will probably have to be explained to users like the range selection

PR #2832, opened 2026-01-07
Optimise MAL search queries by ~11x

MajorTanya:
Previously, the app made one request for the search, and then fired off 1 request per search result to obtain additional data, such as each title's synopsis, etc.

However, MAL's search allows field selection during the initial query, which will return all the data in that first response, avoiding the massive bunch of requests (and alleviating some pressure on MAL from our userbase).

By combining the selected fields into one constant, I was able to also get rid of the `MALUserListSearch` entirely because it was redundant. This allows for a unified `MALManga`->`TrackSearch` helper, further reducing complexity.

I got to my "11x" improvement because on page of search results has 10 elements, and this change turns 11 (1+10 for results) requests into 1. (There are actually more requests being fired for `my:` queries since we do the filtering ourselves and thus pull the entire userlist page by page).
Edit: I checked and no, `/users/@me/mangalist` does not support querying...

Also, the `list_status` data was never used for `my:` queries, which is why it also uses the general `SEARCH_FIELDS`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 
 ### Improved
 - Minimize memory usage by reducing in-memory cover cache size ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2266](https://github.com/mihonapp/mihon/pull/2266))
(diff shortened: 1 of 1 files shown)


MajorTanya:
Yes this is missing a changelog entry. Will make one once reviewed 👍 

Also, once this is merged, I will make a PR to add author names to the MAL search results, since that's actually what I was trying to do; this optimisation stuff was a happy little accident.

PR #2860, opened 2026-01-16
Don't migrate duplicate selected entries in library screen

cuong-tran:
It was using `state.selection` which will include duplicated entries of same manga from different categories.


AntsyLich:
This would've been an issue before the rework I did for library. `state.selection` is a set of long it can't have duplicates.

cuong-tran:
Hm, true. I was mistaken this with a different bug I encountered previously.

PR #2915, opened 2026-01-29
Created a new filter to remove white Strips/Borders.

ArjiAmin:
Adds a new configurable Strip/Border filter to remove white margins and gaps from webtoons. The filter automatically detects the current reading mode and applies different detection algorithm.

For context i wanted this feature mainly to read webtoons on my phone at night without white being blasted onto my eyes. I've been testing this out for months on my Samsung phone, i've been using the debug app on my phone, so i have been very extensively testing out the changes.

A few known issues with the current implementation :
- Filter takes a while to load, doesn't seem to be hardware-linked, it's quite noticable.
- Manga detections is incredibly poor, just like specified in a comment, it really is "hit or miss", some mangas work great, but most of the time it just doesn't work at all, and changing the parameters enough that the narrow strips are detected will cause the filter to remove actual content. I tried various fixes but i just don't know how to improve it.
- I have not deteted any changes in performance, but in extremely rare cases i'd get an "OutOfmemory" error, this has happened twice in over a few hundred hours of reading, and the fix is simply to remove the filter for that specific chapter. I sadly have not been able to reproduce this bug ever since.

Notes :
- These are the parameters i've found out to work in the absolute majority of cases :
    - Replacement color (RGBA)
    - White pixel detection threshold (0-255)
    - Gutter width detection (0-100%)
    - Tall margin height detection (0-100%)
    - Big chunk area detection (0-100%)
- I tried implementing a "reset" button to go back to them, but failed.
- "Replacement color" should perhaps not be rgba, it would be easier to just have options such as : Oled Black, Black, Dark Grey, Grey
- I've placed the `StripFilter.kt` in `app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/webtoon/` because it's more fitting, and because it's not an UI filter but one integrated in the rendering process.
- I did my absolute best to respect coding standards by following examples on neighboring functions and files, but i'm not a kotlin dev and had the code translated from languages i'm familiar with before integrating it manually, so i'm sure there's a lot of room for a lot of improvements to be made.
- I translated French only.

https://github.com/user-attachments/assets/1e0a5777-63ef-48b3-a0c5-2b4f3c9a36f6

### DIFF
--- a/i18n/src/commonMain/moko-resources/fr/strings.xml
+++ b/i18n/src/commonMain/moko-resources/fr/strings.xml
@@ -221,12 +221,6 @@
     <string name="default_category_summary">Toujours demander</string>
     <string name="pref_read_with_volume_keys_inverted">Inverser les boutons de volume</string>
     <string name="pref_crop_borders">Rogner les bordures</string>
-    <string name="strip_filter">Supprimeur de bande/bordure</string>
-    <string name="strip_filter_color">Couleur de remplacement</string>
-    <string name="strip_filter_white_threshold">Seuil de détection du blanc</string>
-    <string name="strip_filter_gutter_width">Largeur d'espacement détectée</string>
-    <string name="strip_filter_tall_margin_height">Hauteur des marges détectée</string>
-    <string name="strip_filter_big_chunk_area">Détection des grands espaces blancs</string>
     <string name="action_global_search">Recherche globale</string>
     <string name="other_source">Autre</string>
     <string name="action_global_search_hint">Recherche globale…</string>


MajorTanya:
(this is just a general note on Contributing to Mihon, not anything about the feature itself)

We generally don't want translations in PRs because translation is handled via [Mihon's Weblate](https://hosted.weblate.org/engage/mihon), where the community translates the base language (English) into the target languages (including French in your case).

Just a note on that very specific part of your PR, I can't speak on whether or not the main dev will accept the feature.

ArjiAmin:
> (this is just a general note on Contributing to Mihon, not anything about the feature itself)
> 
> We generally don't want translations in PRs because translation is handled via [Mihon's Weblate](https://hosted.weblate.org/engage/mihon), where the community translates the base language (English) into the target languages (including French in your case).
> 
> Just a note on that very specific part of your PR, I can't speak on whether or not the main dev will accept the feature.

Ooops, i've reverted the French translation and i've only kept the "../base/strings.xml" file as modified.

PR #2928, opened 2026-02-03
Add `src:local` search alias for Local Source

MajorTanya:
Literally just allows users to search `src:local` in the library. Carelessly forgot to make #2927 a draft before having a discussion about whether I should add it.

### Image
image

### Solved Questions:
- ~~Currently, this code compares to `LocalSource.ID` (which is just a constant of `0L`). But since the `commonMain` `expect class LocalSource` does not have this attribute (only exists on `androidMain`'s `actual class LocalSource`) and Android Studio can get very confused about it, the question is:~~

  _Should I be using the constant or just `0L` directly?_
  => No, we'll be using `LocalSource.ID`.

As context: the commonMain / androidMain and expect/actual problems are largely due to the fact the project uses only one effective source set (Android target), which makes Kotlin/Gralde think there is only one source set during analysis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Display author & artist name in MAL search results ([@MajorTanya](https://github.com/MajorTanya)) ([#2833](https://github.com/mihonapp/mihon/pull/2833))
 - Add filter options to Updates tab ([@MajorTanya](https://github.com/MajorTanya)) ([#2851](https://github.com/mihonapp/mihon/pull/2851))
 - Add `src:` prefix to search the library by source ID ([@MajorTanya](https://github.com/MajorTanya)) ([#2927](https://github.com/mihonapp/mihon/pull/2927))
+  - Add `src:local` as a way to search for Local Source entries ([@MajorTanya](https://github.com/MajorTanya)) ([#2928](https://github.com/mihonapp/mihon/pull/2928))
 
 ### Improved
 - Minimize memory usage by reducing in-memory cover cache size ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2266](https://github.com/mihonapp/mihon/pull/2266))


MajorTanya:
Also not sure if I'm happy with the implementation itself, feels... inelegant (it's the `compareTo` that kinda looks off to me)

MajorTanya:
<details>
<summary>
Android Studio complaint can be ignored, my grade state was borked
</summary>

Here is the Android Studio complaint:

image

</details>

here is the entire commonMain LocalSource file
https://github.com/mihonapp/mihon/blob/1abf01c4a030b089e6300cb284cc93f7ed323a22/source-local/src/commonMain/kotlin/tachiyomi/source/local/LocalSource.kt#L1-L6

and the `actual` class with the attribute lives here:
https://github.com/mihonapp/mihon/blob/32257e438e2f0bea7e0bfe84dc72135795d620ad/source-local/src/androidMain/kotlin/tachiyomi/source/local/LocalSource.kt#L365

app/src/main/java/eu/kanade/tachiyomi/ui/library/LibraryItem.kt
```diff
@@ -28,6 +29,9 @@ data class LibraryItem(
         if (constraint.startsWith("id:", true)) {
             return id == constraint.substringAfter("id:").toLongOrNull()
         } else if (constraint.startsWith("src:", true)) {
+            if (constraint.substringAfter("src:").compareTo("local", ignoreCase = true) == 0) {
+                return source.id == LocalSource.ID
+            }
             return source.id == constraint.substringAfter("src:").toLongOrNull()
```
AntsyLich:
```suggestion
            val querySource = constraint.substringAfter("src:")
            // "local" should be a constant
            return if (querySource.equals("local", ignoreCase = true)) {
                source.id == LocalSource.ID
            } else { 
                source.id == querySource.toLongOrNull()
            }
```

AntsyLich (changes requested):
For the AS error. Just add a expect variable in `commonMain`

MajorTanya:
> For the AS error. Just add a expect variable in `commonMain`

The error went away with a purged gradle cache and such, also not sure if this is the best PR to go clean up the mess that are the expect/actual discrepancies.

MajorTanya:
@AntsyLich I think we should ignore the commonMain problem one more time and just go ahead with this / your code suggestion (will apply that later when I have some time). Making commonMain coherent with androidMain should be a new PR since there is a lot of Inconsistency between them. (the ID attribute wouldn't even be the only thing in LocalSource anyhow)

AntsyLich (approved):

PR #2929, opened 2026-02-03
Clean up some build warnings

MajorTanya:
I tried my best to split the changes into atomic-ish commits, where each commit is either a specific change, or a grouping of extremely similar changes.

### - Context Receivers are now Context Parameters (26e61580edec41e4a59c8400bc027c5b631132db)
Apparently, the biggest difference to us is that the new thing has to have names (`_` can be used for anonymous ones).

### - Opted in to the new annotation default rules (35cf4ba1dbc16aacf32945c4a879a904763b5f5a)
I'll be completely honest and say that I don't 100% understand [the KEEP for this](https://github.com/Kotlin/KEEP/blob/change-defaulting-rule/proposals/annotation-target-in-properties.md). But the 4 annotations the warning showed up for are all `val`s and in all cases, we _do_ want the field to be adhering to the thing the annotation is about, so I believe this makes sense for us to opt in to now (the flag I used is the upcoming default behaviour).

## Deprecation replacements
- `TooltipDefaults.rememberPlainTooltipPositionProvider` -> `rememberTooltipPositionProvider(TooltipAnchorPosition.Above)` (c3b44e78930501d66f39bb7c378c2b555bbbf756)
- AboutLibraries's `LibrariesContainer` _without_ `Libs` -> `LibrariesContainer` with `Libs` (d923c06e8c80b6d38b860c2ae40e4e78e13c3fc1)
  - requires us to call `produceLibraries` ourselves now; the replacement is what's written in their Readme
- `ClipboardManager` -> `Clipboard` (56e34439ff702d80f4483c1e7057e3e4a424720c)
  - uses suspend functions now
  - also yes, it is that complicated now...
- `MenuAnchorType` -> `ExposedDropdownMenuAnchorType` (2dd5aba9b362c53c592bf4ed4eac7563a45b8374)
  - literally just a typealias deprecation

## Other
- Use multi-dollar strings for GraphQL query strings to AniList and Suwayomi (d89b250c70ea2aec6eb71d0095fb0d421988bf26)
  - saves having to do `${'$'}` for dollar literals all the time (introduced in Kotlin 2.1 apparently)
- remove explicit enabling of `OptimizeNonSkippingGroups` since it's enabled by default now (5ccda24347b90e474a088cc32b1aa98aed81ee68)
- `ArchiveInputStream.getNextEntry`: changed to block body due to an upcoming deprecation of "return in function with expression body and without explicit return type" ([KTLC link](https://youtrack.jetbrains.com/issue/KTLC-288)) (353240f99498705cae22ceee100eeb2395b5432f)
- Suppress deprecation warning for Left and Right KeyboardArrow icons (43a96383e58b37da80dca25aadc86adc39b6b9d2)
  - we use them for calendar navigation, so AutoMirrored doesn't make sense for us
  - this warning has annoyed me for literal years now, time to make it shut up
- Various redundant casts & conversions & other minor stuff (87fdacb09f763cba8392d1d228ea65f0aaa98095, 31309abd5c9b540bd77daa7b08b952e52a55ad67, a8fca9a44845d5fa7fa0f88ccb13fc4ed2ccd1ab)

## Unchanged
I decided to _not_ touch the warnings caused by the DelicateCoroutinesApi annotation on the several launchX functions we have defined in the project since they are, well, delicate and can often be replaced by better things (not the current uses though I believe).

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -166,6 +166,7 @@ kotlin {
             "-opt-in=kotlinx.coroutines.FlowPreview",
             "-opt-in=kotlinx.coroutines.InternalCoroutinesApi",
             "-opt-in=kotlinx.serialization.ExperimentalSerializationApi",
+            "-Xannotation-default-target=param-property",
         )
     }
 }


MajorTanya:
Open to roll back some of the changes but I tried my best to find the most sensible solution to each of these warnings

MajorTanya:
Also yes, no changelog (though I don't think this change should have one)

MajorTanya:
@AntsyLich merge ready

PR #2940, opened 2026-02-09
Allow LibraryUpdateJob to run with VPN (#2477)

### LINKED ISSUE #2477: Library doesn't automatically refresh over VPN connections
### Steps to reproduce

Connect to a VPN, then wait whatever duration you have it set to automatically refresh at.

### Expected behavior

It refreshes every whatever hours without manual intervention.

### Actual behavior

Instead, the library won't refresh unless you either manually do it, or you disconnect from your VPN (in which case it updates immediately upon disconnection). There's no error or anything; it just acts as if it was never supposed to refresh automatically to begin with.

### Crash logs

_No response_

### Mihon version

0.19.1

### Android version

Android 12

### Device

Google Pixel 3

### Other details

This problem occurs with Mullvad; unsure about other VPNs.

Maybe an old device issue, but my previous Google Pixel 3a worked fine, so I'm unsure. Currently also testing the beta, since I believe I used to run that; will edit when the update's supposed to happen. UPDATE: The beta won't sync automatically either.

### Acknowledgements

jeremiejig:
Hello,

To fix #2477, I took the least intrusive change because all other default capabilities might make sense to have (like BANDWIDTH_CONSTRAINED).

I also need some help testing the change (I never got around to installing any tooling for android)

The fix seems easy and I hope testing does not reveal some more complex mechanism I did not see, because otherwise I cannot promise a follow-up on this issue.

Again any help welcomed, I did not realise so much months passed since I last looked at this issue.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
@@ -436,6 +436,7 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
                     NetworkType.CONNECTED
                 }
                 val networkRequestBuilder = NetworkRequest.Builder()
+                networkRequestBuilder.removeCapability(NetworkCapabilities.NET_CAPABILITY_NOT_VPN)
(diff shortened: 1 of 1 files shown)


AntsyLich:
Add a changelog entry

PR #2950, opened 2026-02-12
Add missing indexes to improve database query performance

Lolle2000la:
Adds a number of indexes for columns that are queried against but currently have no index yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,6 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fix pre-1970 upload date display in chapter list ([@MajorTanya](https://github.com/MajorTanya)) ([#2779](https://github.com/mihonapp/mihon/pull/2779))
 - Fix crash when trying to install/update extensions while shizuku is not running ([@NGB-Was-Taken](https://github.com/NGB-Was-Taken)) ([#2837](https://github.com/mihonapp/mihon/pull/2837))
 - Fix Add Repo input not taking up the full dialog width ([@cuong-tran](https://github.com/cuong-tran)) ([#2816](https://github.com/mihonapp/mihon/pull/2816))
-- Fix Add missing indexes for columns that are queried against but have no existing indexes ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2950](https://github.com/mihonapp/mihon/pull/2950))
 
 ### Other
 - Enable logcat logging on stable and debug builds without enabling verbose logging ([@NGB-Was-Taken](https://github.com/NGB-Was-Taken)) ([#2836](https://github.com/mihonapp/mihon/pull/2836))


MajorTanya:
Have you noticed a change in performance before & after? Also, what was the motivation for this PR? (aside from "indices are good for speed")

Lolle2000la:
I definitely noticed a difference in responsiveness on my library in the emulator, but I also tried writing a synthetic test, which showed speedup of 100x on tables in excess of thousands of entries (like say, having many entries in the library and many chapters), though I realized that I really only tested sqlite, because, yeah, queries are faster with an index, I don't know what I expected. Note that I ran the synthetic test not in the emulator under the assumption of sqlite-is-sqlite, and instead did it directly on my machine.

Basically, there are some queries that are run somewhat frequently and need to do a full scan given the missing indexes, which slower the slower the device is. Aside from performance, this also impacts battery life since way more data needs to be loaded from storage than necessary.

I also have an experiment I might try to contribute later where I use FTS4 to speedup various library searches, but I don't currently see any improvements even if they should theoretically be there (and it seemed to be way bigger to implement than this, which was quite simple to produce, just look what indexes are missing) so I reduced my scope to just contributing this.

The emulator ran on a Samsung 980 Pro SSD, so the improvements should be bigger on phones with eMMC etc.

MajorTanya:
I mean yeah, sqlite will be faster with indices 🤣

How large is the library on your emulator, for reference? My real library on my phone is something like 2k entries but I don't _think_ I ever noticed the app being slow. What areas of the app did you see most of the improvements in?

Also, do you reckon this would affect app startup time?

Lolle2000la:
I basically added half of webtoon, so its 8863 :joy: 

> What areas of the app did you see most of the improvements in?

It basically just has a slight impact when just navigating. Its not dramatic, but noticable in my case. Especially the search bar sometimes just never does anything (I can wait for a literal minute before it begins to process), though that is the thread starvation issue mostly.

> Also, do you reckon this would affect app startup time?

Oof, not negatively I guess? :sweat_smile:  I haven't really looked at what the app does at startup, if it queries a lot, it should be better, if it creates a lot of new entities at every start it should be worse, otherwise I haven't really noticed app startup time in general (like, never thought about it).

MajorTanya:
Yeah I was referring to positively affecting startup 😅

Lolle2000la:
I integrated a small piece of test code that measures the time from startup to arriving on the library. I used [testing_changes.patch](https://github.com/user-attachments/files/25272177/testing_changes.patch) for this.

* 2948 ms without the changes from the PR
* 2851 ms with the changes from the PR

Between runs, this jumps by around 100 ms, so I would say the difference is negligible.

MajorTanya:
Yeah that seems statistically insignificant. Would've been cool if it had had some effects but at the same time, I don't think I really expected it to.

Now to wait for @AntsyLich to take a look.

AntsyLich:
Remove the changelog entry. This isn't affecting any UI/UX

Lolle2000la:
> Remove the changelog entry. This isn't affecting any UI/UX

Done :+1:

PR #2951, opened 2026-02-12
Prevent thread starvation in TransactionContext.kt

Lolle2000la:
A lot of places seem to be eating threads (kotlin task parallelism as a language feature is a bit of a mess to be honest), but the `TransactionContext.kt` seems to be the worst offender.

This change resolves that in two ways:

1. It serializes the requests before it hits sqlite (prequisite for 2)
2. It only grabs a thread once it actually can do anything with it.

If the lock is hit, the thread can yield, meaning that threads can't pile up waiting for the database. This primarily helps responsiveness.

It is one of the places I identified that has this problem. In general, a lot of IO seems to block the threads instead of either yielding or using the IO pool, causing responsiveness issues. Every one of those changes might have unintended consequences though, so I will only open these PRs in small portions.


Lolle2000la:
#2955 fully contains and builds on this, but this PR alone also provides improvements.

PR #1752, opened 2025-02-20
Add "Monochrome" theme

MajorTanya:
This theme is mainly geared towards e-Ink displays with limited/no colour capabilities. Previous themes like Yin & Yang would make heavy use of greyscale colours which could look off on some devices.

This theme is probably not conformant to Material Design 3 colour scheme guidelines, but it does boast some amazing WebAIM contrast ratios (#FFFFFF text on #000000 background gets a ratio of 21:1, vice versa too).

Initially, this was intended as a purely black and white theme but some contrast issues arose, such as the download badges (tertiary background, onTertiary text colour) having the same colour as unread badges (primary/onPrimary), or the step indicators (stops) not being visible on sliders (since they use the colours of the opposite state track (active region stops are the colour of the inactive region track and vice versa).

To mitigate this, each variant (dark/light) of the theme has one additional grey mixed in for their tertiary and secondaryContainer colours each. For the dark variant, this is a #A0A0A0 background for #000000 text (8.03:1 contrast ratio) and for the light variant, it is a #505050 background for #FFFFFF text (8.06:1 contrast ratio). This results in distinct unread vs download badges and visible steps in the sliders.

Closes #1747.

### Images
| Dark mode | Light mode |
| ------- | ------- |
|  |  |
|  |  |
|  |  |

PR #2953, opened 2026-02-13
Add E-Ink Theme

MatiasDesuu:
Added a new theme called E-Ink that only uses White (#FFFFFF) and Black (#000000) colors.
Useful for devices with eink screens like e-readers.

Like Monochrome but Monochrome still has some gray variants that doesn't look too good on eink.

Screenshot_20260213_125057)

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsReaderScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsReaderScreen.kt
@@ -62,6 +62,10 @@ object SettingsReaderScreen : SearchableSettings {
                 preference = readerPref.pageTransitions(),
                 title = stringResource(MR.strings.pref_page_transitions),
             ),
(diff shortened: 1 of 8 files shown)


MajorTanya:
Monochrome uses those greyscales for good reasons, which you would know if you actually bothered to check the PR in which I added it: #1752

MatiasDesuu:
I checked it, I just don't like how it looks the Monochrome mode with the gray in some parts, even if that means some sliders looks "weird" with full black/white, that's why I made this one.

Smol-Ame:
The themes follow Material Design color scheme, it's not meant to just cater to e-ink devices. Monochrome is basically the middle ground for all devices.

MajorTanya:
> The themes follow Material Design color scheme, it's not meant to just cater to e-ink devices.

Monochrome doesn't. To provide the minimum feasible amount of contrast, it uses one additional grey for light/dark mode, that's all.

PR #2955, opened 2026-02-14
Fix thread starvation caused by not yielding or using an inappropriate thread pool

Lolle2000la:
Builds on (and fully includes) #2951 

Currently, especially on bigger libraries, the app gets quite unresponsive. A lot of stuff, like updating the library, worsens the problem while it runs. Basically, a lot of interactions with IO and network are blocking threads instead of yielding them back to the thread pool, or use the much more limited main thread pool instead of the IO thread pool (which has 64 threads). 

On my library I get significantly better responsiveness upon many different tasks. Example: before this, I had around a second of wait before a library entry opens. With these changes, this happens almost instantly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Minimize memory usage by reducing in-memory cover cache size ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2266](https://github.com/mihonapp/mihon/pull/2266))
 - Optimize MAL search queries ([@MajorTanya](https://github.com/MajorTanya)) ([#2832](https://github.com/mihonapp/mihon/pull/2832))
 - Reword download reindexing message to avoid confusion ([@MajorTanya](https://github.com/MajorTanya)) ([#2874](https://github.com/mihonapp/mihon/pull/2874))
+- Rework internals for better performance ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2955](https://github.com/mihonapp/mihon/pull/2955))
 
 ### Changed
 - Update tracker icons ([@AntsyLich](https://github.com/AntsyLich)) ([#2773](https://github.com/mihonapp/mihon/pull/2773))


AntsyLich:
I'm a bit concerned about this 2 changes https://github.com/mihonapp/mihon/pull/2955/changes/bd3cdbab39db5afa5cad8a7aef820e0956c1c58a https://github.com/mihonapp/mihon/pull/2955/changes/0170c75754a57244cb85b7e41ede86753d05ea42. And can you make https://github.com/mihonapp/mihon/pull/2955/changes/7f0effb07bc4dec7280bc11fcad27e6692cb9e7b a separate PR

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderViewModel.kt
```diff
@@ -145,84 +143,92 @@ class ReaderViewModel @JvmOverloads constructor(
...
-                                            it.url,
-                                            manga.title,
-                                            manga.source,
-                                        )
-                                    ) ||
-                                (manga.bookmarkedFilterRaw == Manga.CHAPTER_SHOW_BOOKMARKED && !it.bookmark) ||
-                                (manga.bookmarkedFilterRaw == Manga.CHAPTER_SHOW_NOT_BOOKMARKED && it.bookmark)
+    private var chapterListCache: List<ReaderChapter>? = null
+    private suspend fun getChapterList(): List<ReaderChapter> {
+        if (chapterListCache == null) {
```
MajorTanya: I think you could save a level of indentation if you invert this and return `chapterListCache` early here (would also save the !! assertion). Then the rest of the function can be dedented one to save us some vertical space.

MajorTanya (commented):
just an observation from a quick read-through

Lolle2000la:
> And can you make https://github.com/mihonapp/mihon/commit/7f0effb07bc4dec7280bc11fcad27e6692cb9e7b a separate PR

I think I will just keep it out then and look at it later. I mostly did it, because I wanted to remove all the instances of the `runBlocking` import, the rest came along with it. I will just revert it for now and make a PR later once this is done.

> I'm a bit concerned about this 2 changes https://github.com/mihonapp/mihon/commit/bd3cdbab39db5afa5cad8a7aef820e0956c1c58a https://github.com/mihonapp/mihon/commit/0170c75754a57244cb85b7e41ede86753d05ea42.

What specificially, aside from the small teeny tine little bit of being extremely at the core of things possibly breaking strongly? :sweat_smile: 

Jokes aside, the lazy property replacement is fairly easy to test, regarding the `awaitAndRelease`, I mostly tested the migration feature to see if it worked (since it appears to be the only consumer), and it works without a problem on my library. I tried migrating three entries.

app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderViewModel.kt
Lolle2000la: @MajorTanya Applied this, thank you!

AntsyLich:
> > I'm a bit concerned about this 2 changes [bd3cdba](https://github.com/mihonapp/mihon/commit/bd3cdbab39db5afa5cad8a7aef820e0956c1c58a) [0170c75](https://github.com/mihonapp/mihon/commit/0170c75754a57244cb85b7e41ede86753d05ea42).
> 
> What specificially, aside from the small teeny tine little bit of being extremely at the core of things possibly breaking strongly? 😅
> 
Exactly that :p

Lolle2000la:
> Exactly that :p

I call that a krill issue :shrimp: 

---

What I tested for either of those:

For [bd3cdba](https://github.com/mihonapp/mihon/commit/bd3cdbab39db5afa5cad8a7aef820e0956c1c58a):

* Opening a manga and opening a chapter with no filter
* Doing so with filters (favorite, read, scanlator, etc.)
* Opening a manga with no chapters (empty chapter list)

No problems here!

For [0170c75](https://github.com/mihonapp/mihon/commit/0170c75754a57244cb85b7e41ede86753d05ea42):

* I migrated a manga for all the cases above

No problems here either!

Anything else I should test? That were the test cases I came up with.

AntsyLich:
Add a changelog entry

Lolle2000la:
[Oh so *now* you want a changelog entry smh](https://github.com/mihonapp/mihon/pull/2950#issuecomment-3902337947) /s

---

 Would "Improved responsiveness by fixing thread starvation issues" be good? Or is this a fix? Like "Fix thread starvation causing app to sometimes become unresponsive"?

AntsyLich:
Needs to be user friendly if possible so maybe "Rework internals for better performance"

Lolle2000la:
> Needs to be user friendly if possible so maybe "Rework internals for better performance"

Ok, added exactly that!

MajorTanya:
I've been using a nightly build for a few hours now and it seems that the download indexing that sometimes happens on startup now fails silently somehow. Opening the app from the background death in the way that usually triggers the Checking Downloads banner & reindexing now only shows that banner but the downloaded chapters are not discovered. I have to manually go to reindex downloads from the Advanced settings.

I'm pretty sure it's this PR causing the issue because my previous build was from the day before and that did not exhibit this issue (unless I was seriously blind but I doubt it).

Lolle2000la:
That sounds quite critical, can you inspect the logcat to see what might be the error? The fact that it happens in startup but not when manually indexing boggles my mind. 

I looked at my changes again and didn't see any immediate candidate for causing this. Since today is a normal Monday, I can't promise to be quick, but I will take a look at it!

Lolle2000la:
@MajorTanya I think I found the issue. I think there was timing bug that due to prior ordering from lack of concurrency didn't come up. #2970 should fix that.

MajorTanya:
While I haven't gotten to test your download index fix yet (apologies), I just got a massive number of errors when my library update happened. Most of the titles failed due to "Source not found" when the sources are all happily available. I'm guessing the extensions list wasn't (fully) loaded?

MajorTanya:
> While I haven't gotten to test your download index fix yet (apologies), I just got a massive number of errors when my library update happened. Most of the titles failed due to "Source not found" when the sources are all happily available. I'm guessing the extensions list wasn't (fully) loaded? 

I can also confirm that after starting the app (and going to the extensions tab, though I don't know if that matters), running the same update again results in 0 errors.

Lolle2000la:
> While I haven't gotten to test your download index fix yet (apologies), I just got a massive number of errors when my library update happened. Most of the titles failed due to "Source not found" when the sources are all happily available. I'm guessing the extensions list wasn't (fully) loaded?

It probably is just that. I think the extensions stuff is probably racing the update currently. This isn't the issue once the app is started (all ordering dependencies are actually expressed), but here I don't see explicit waiting (or I'm blind).

Lolle2000la:
@MajorTanya Found the issue and opened PR #2972 for it! Basically, ordering before just happened to go right, but now it is explicit.

PR #2970, opened 2026-02-16
Fix cache invalidation isn't done at startup

Lolle2000la:
Due to `lastRenew` being initialized to the current time at startup, the logic decides no invalidation is necessary. Removing that invalidation will see it be initialized once invalidation is done.

This is a follow up upon the report in [this comment](https://github.com/mihonapp/mihon/pull/2955#issuecomment-3906561153) in #2955. I'm not sure why this is only happening now, but just removing the init fixes the issue.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
@@ -449,8 +449,9 @@ class DownloadCache(
 
     private var updateDiskCacheJob: Job? = null
     private fun updateDiskCache() {
-        updateDiskCacheJob?.cancel()
+        val previousJob = updateDiskCacheJob
         updateDiskCacheJob = scope.launchIO {
+            previousJob?.cancelAndJoin()
             delay(1000)
             ensureActive()
             val bytes = ProtoBuf.encodeToByteArray(rootDownloadsDir)


Lolle2000la:
Just to add to this: I tested it by downloading something, closing the app, deleting the downloaded chapter and then observing if the app upon startup notices the missing chapter automatically.

MajorTanya:
Only on mobile right now, but does not initialising the variable have downstream effects? Like, it's not lateinit var or something?

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
MajorTanya, line 91: Nevermind my comment, just found this in the expanded context.

MajorTanya:
I'll make a build for myself off of this branch and test drive it for a while. Will report back in a few hours once I've seen the restart reindexing a few times. 👍

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
Lolle2000la: I wonder if there originally was the intent to not refresh at startup for some reason, and the functionality just broke due to reindexing checking first before the variable was set and propagated?

Anyway, it seems that indexing at startup **is** the desired behavior, right?

MajorTanya: Not necessarily desired if it's not necessary but that's kind of an impossible problem (cache invalidation)

MajorTanya:
> I'll make a build for myself off of this branch and test drive it for a while. Will report back in a few hours once I've seen the restart reindexing a few times. 👍

Got bad news for ya. Built your branch earlier and after a bit, I started the app from the background and saw the "Checking downloads" banner, yet the downloads were not found. The manual route still worked as before.

This does not appear to be a fix.

Lolle2000la:
Aww dammit. It did on my machine (tm).

I assume that it has to do with threading behavior and ordering, could you tell me which device/emulator config, specifically the core configuration, you used? Maybe I can reproduce it then.

In the meantime I will take a look at what might be a correctness error.

Lolle2000la:
> Got bad news for ya. Built your branch earlier and after a bit, I started the app from the background and saw the "Checking downloads" banner, yet the downloads were not found. The manual route still worked as before.

Did you use the local source or is this about downloads?

Lolle2000la:
@MajorTanya Can you try again with the latest commit? I can't reproduce the bug on my machine, but maybe this concurrency guarantee would fix it?

MajorTanya:
> > Got bad news for ya. Built your branch earlier and after a bit, I started the app from the background and saw the "Checking downloads" banner, yet the downloads were not found. The manual route still worked as before.
> 
> Did you use the local source or is this about downloads?

Downloads from extensions.

> @MajorTanya Can you try again with the latest commit? I can't reproduce the bug on my machine, but maybe this concurrency guarantee would fix it?

Will do

> tell me which device/emulator config, specifically the core configuration, you used? Maybe I can reproduce it then.

My phone, which is a Xiaomi 11T. It is technically my fork of Mihon but the fork doesn't at all touch any code that could be related to this. 

```txt
Android version: 14 (SDK 34; build UP1A.231005.007)
Device brand: Xiaomi
Device manufacturer: Xiaomi
Device name: amber (amber_eea)
Device model: 21081111RG
WebView: Android System WebView 144.0.7559.132
```

I should probably also mention that I compile a proper build (`assemblePreview` is the specific Gradle task I use), not just a debug one.

MajorTanya:
Sadly still no dice. The banner appears but the downloads are not discovered.

I have applied your commits applied as patches to the main branch so it is actually building with your changes. 

I noticed that the banner appears only for ~1 second currently, while it stays on screen for about ~10 seconds when I manually reindex the downloads to fix the issue. Is it possible the task is not actually running and the banner just appears for the minimum amount of time? (we do something like that for the extensions list refresh because it's too fast for users to see thst anything happened, so it's possible we have that set for the reindexing banner as well?)

Lolle2000la:
Thank you for the info on the device, also for this note:

> I should probably also mention that I compile a proper build (assemblePreview is the specific Gradle task I use), not just a debug one.

I will try to configure more cores (for more real parallelism) and with optimized code to see if I can get it reproduced that way.

> Sadly still no dice. The banner appears but the downloads are not discovered.

In my emulator the downloads are discovered, so I don't know where this comes from :sweat_smile:, BUT:

> I noticed that the banner appears only for ~1 second currently, while it stays on screen for about ~10 seconds when I manually reindex the downloads to fix the issue. Is it possible the task is not actually running and the banner just appears for the minimum amount of time? (we do something like that for the extensions list refresh because it's too fast for users to see thst anything happened, so it's possible we have that set for the reindexing banner as well?)

Do you have a lot of downloads? Because my hypothesis based on this is that the indexing fails on the first run for the bigger amount of downloads.

That is the biggest difference I can come up with between the emulator. In my emulator I basically downloaded 5 chapters, while you might have a lot more downloads. If that is the case, then this would explain it, the first indexing stops prematurely. Still no game as to why though, so if you have any suspicious logs from immediately after starting that suggest an exception being thrown, that would help.

MajorTanya:
I have approximately 13 GB of downloaded chapters.

As for logs, I'll run it with logcat attached later and report back.

Edit: I should clarify that I've seen users with much, _much_ bigger download sizes in support cases.

MajorTanya:
> I have approximately 13 GB of downloaded chapters.
> 
> As for logs, I'll run it with logcat attached later and report back.
> 
> Edit: I should clarify that I've seen users with much, _much_ bigger download sizes in support cases.

Okay, I tried to replicate the behaviour on a debug install from AndroidStudio onto my device, with my current library restored into it and using the same underlying folder for downloads. Interestingly, the debug build does not have the problem of "forgetting" about the downloads.

Doing a preview build (which is effectively the same as a release build in terms of the process) does "forget" about downloads AND the banner doesn't appear at all?

With some excessive manual logging calls, I can see that the `init` code runs and it does find and use the `diskCacheFile` data. The `renewCache` function is also called and executed all the way through. Very strange.

MajorTanya:
Manually checking still works fine in both cases however 🤔

Lolle2000la:
I gave up, threw this at gemini, and it says that, most likely, with all of this context, the problem is improper code stripping. 

>   Here's my reasoning:
>   1. Debug vs. Release: This is the classic sign of an R8/ProGuard issue. Code that works in debug fails when optimizations and stripping are applied.
>   2. Downloads "Disappear": You've confirmed init loads the cache correctly, but after renewCache runs, the downloads are gone. This means renewCache is overwriting the valid, in-memory cache with an empty one.
>   3. `renewCache` "Completes": The function can execute from start to finish without crashing, but still produce an empty result. If R8 has stripped necessary code from the unifile library (for reading directories) or the protobuf library (for serialization), those operations will fail silently and return empty
      lists instead of throwing exceptions. The function "completes," but the result is wrong.
>   4. No Banner: The indexing process finishing almost instantly (due to the silent failures mentioned above) is why the banner doesn't appear. The isInitializing state changes from true to false so quickly that the UI's 1-second debounce doesn't register it.
>   5. Manual Indexing Works: This is likely due to a change in application state after startup that allows the stripped code to function by chance, or it could be that the underlying issue is still present but less likely to manifest.

The third reason in particular sound VERY reasonable, though it would be weird why this stipping didn't happen before in that case. I'm trying to test this right now, but also committed those changes just in case.

MajorTanya:
I think I may have found something but I can't say anything yet. Chasing phantoms right now...

I have applied your new commits as patches as well to see

Lolle2000la:
I tried with "benchmark" profile (basically equivalent to production) and it worked in both scenarios:

1. Moving out a file under the hood and restarting: Only the deleted chapter was not indexed
2. Moving the file back in and restarting: Also worked! The file is visible again.

I think it might actually have been a stripping problem, though could you please check to see if it works for you as well?

Lolle2000la:
Note that [724432d](https://github.com/mihonapp/mihon/pull/2970/commits/724432dca9aa35538c0906fa2a837f985b14f9d7) is still necessary, without it, it outright just doesn't work in any config. The rest is just nice to be correct.

MajorTanya:
Okay I've figured out something exciting. It's not a stripping issue (at least not necessarily).

https://github.com/mihonapp/mihon/blob/27749fc583527ab1360594a69ef586a4311c93f5/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt#L350-L361

The `sources` need to be initialised in order for the downloads to be found, and due to your previous changes, it is empty when the app does the startup reindexing.
You can confirm this for yourself if you switch very quickly to the Browse tab before/just as the reindexing banner/process begins. This loads the sources in time for the `renewalJob` to end up with a filled `sources` list and THEN download discovery works out fine.

For this setup, the only patch I applied was 724432d, none of the rest, except for some logging lines for my reference.

---

Edit: forgot to add my conclusion:

In conclusion, I think this would be helped much better by not having source be loaded this way since that would completely obviate most of your commits on this PR in exchange for a simplified data flow.

Lolle2000la:
@MajorTanya You mean like this? I reverted all changes unrelated to this and the first initial one.

Lolle2000la:
That would explain why I can't reproduce it though, maybe the emulator is too fast (at least on IO).

MajorTanya:
> That would explain why I can't reproduce it though, maybe the emulator is too fast (at least on IO).

I don't think that'd be the case. It's more likely a matter of having several extensions installed and a sizeable amount of downloads to traverse. Which is usually not the case for any normal emulator setup.

> @MajorTanya You mean like this? I reverted all changes unrelated to this and the first initial one.

I'll try this out and see. I'll show you the entire patch diff I have active right now:

```diff
Index: app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt	(revision 42fbe00e451845324623b6ae5d92eba44eae0e44)
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt	(date 1771357329287)
@@ -109,7 +109,6 @@
                             ProtoBuf.decodeFromByteArray<RootDirectory>(it.readBytes())
                         }
                         rootDownloadsDir = diskCache
-                        lastRenew = System.currentTimeMillis()
                     }
                 } catch (e: Throwable) {
                     logcat(LogPriority.ERROR, e) { "Failed to initialize from disk cache" }
@@ -345,6 +344,7 @@
         }
 
         renewalJob = scope.launchIO {
+            logcat { "TANYA: starting renewalJob" }
             if (lastRenew == 0L) {
                 _isInitializing.emit(true)
             }
@@ -358,6 +358,8 @@
                 sources = getSources()
             }
 
+            logcat { "TANYA: sources=$sources" }
+
             val sourceMap = sources.associate { provider.getSourceDirName(it).lowercase() to it.id }
 
             rootDownloadsDirMutex.withLock {
```

That logging for `sources` shows an empty list

Lolle2000la:
@MajorTanya Have you also applied [b32a070](https://github.com/mihonapp/mihon/pull/2970/commits/b32a070a5a278f73f8eb188030092fc62c9fc941)? Its changes don't appear in the changelog.

MajorTanya:
> @MajorTanya Have you also applied [b32a070](https://github.com/mihonapp/mihon/pull/2970/commits/b32a070a5a278f73f8eb188030092fc62c9fc941)? Its changes don't appear in the changelog.

That was meant to show you the setup I used to figure this out. I haven't tried your newest changes yet.

Lolle2000la:
Ah ok, I misunderstood!

MajorTanya:
Okay so I tried b32a070 but that does not seem to help at all. The banner does not appear and I can see from the `sources` log line that it's still an empty list. However, it appears that I can't get my workaround to work any more either (maybe due to the lack of the timeout or something? I don't know)

Lolle2000la:
I tried changing the condition to wait not just until init is reported, but until there are elements in the collection. Hopefully that fixes it :sweat:

MajorTanya:
> I tried changing the condition to wait not just until init is reported, but until there are elements in the collection. Hopefully that fixes it :sweat:

Wouldn't that make it hang indefinitely if the user has no sources installed/enabled?

Lolle2000la:
It would in essence be stuck doing nothing, that is right, but because it is suspend it would not cost anything. As soon as a user installs an extension, the task will continue then, though still not really doing anything. So the corner case is A-okay!

Just to make sure my assumption isn't wrong, I just tested the fix with a fresh install.

Lolle2000la:
Now, all of that is under the assumption that this works now, so could you test it again? On my machine(tm) it works.

MajorTanya:
> Now, all of that is under the assumption that this works now, so could you test it again? On my machine(tm) it works.

Still nothing... At this point, maybe we should revert #2951 for the time being so it doesn't end up in the next preview release or something

But that'd be up to Antsy

Lolle2000la:
That might make sense, though I would like to know where this is coming from. I can't reproduce this no matter what.

Like, even when I use my own library (which is not definitely not small), and basically download half of webtoon (literally thousands of chapters across many different series) and install a a bunch of extensions regardless of whether I actually plan to use it, I still just have no issues.

Is there anything else on your machine that separates it from my machine?

Lolle2000la:
I tried one more thing, here the extension loading should be triggered before renewal can happen by making it idempotent. If this doesn't solve it, then I will leave it and sleep a bit over it, hoping that I will come up with a solution later :joy:

MajorTanya:
I've cloned your branch directly as a "proper" Mihon install. I'm using the release build and installing that each time. All I have added are a few logcats around the place to trace the reindexing work, specifically `sources`.

Behaviourally, sometimes it can populate `sources`, sometimes it cannot. I don't know why. I'm assuming this is a pure data race condition at this point due to the parallelised/suspended changes of your thread starvation PR.

I have no clue what to do about this other than to make extension loading not-suspend again or undoing all of #2955.

MajorTanya:
@AntsyLich any better ideas?

Lolle2000la:
I think I fixed it, I was able to reproduce it by simply having a lot in the download queue regardless of whether it is downloaded.

[34be036](https://github.com/mihonapp/mihon/pull/2970/commits/34be036da20f4860da3312d6525b5993e230010e) should fix it! :crossed_fingers:

Lolle2000la:
The bug should now be both fixed and the whole job stuff should be thread safe!

MajorTanya:
For public context: I'm currently still daily-ing this PR to see if the other edge cases we discussed in private have been ironed out. So far it's looking alright but some things (like some background update / duplicate runs / other weirdnesses) haven't occurred yet (as in, I haven't had an automatic update run yet, for example), so this is not ready to merge.

app/src/main/java/eu/kanade/tachiyomi/data/download/DownloadCache.kt
```diff
@@ -349,15 +373,10 @@ class DownloadCache(
...
-            // Try to wait until extensions and sources have loaded
-            var sources = emptyList<Source>()
-            withTimeoutOrNull(30.seconds) {
-                extensionManager.isInitialized.first { it }
-                sourceManager.isInitialized.first { it }
-
-                sources = getSources()
-            }
+            // Wait until sources are loaded.
+            sourceManager.catalogueSources.first { it.isNotEmpty() }
```
AntsyLich, line 377: How is this better compared to the previous code

MajorTanya: I believe this is here to address the issue that `sources` would be empty at startup and would never populate (including with the 30 second timeout), thus leading the app to not reindex at startup even when it would be necessary (ie no downloaded chapters recognised)

Lolle2000la: More specifically, the querying can happen so close to launch now that the list of packages is always empty on some devices. In that case it will wait until some sources are loaded. This is why [this change](https://github.com/mihonapp/mihon/pull/2970/changes#diff-4880ffb9903e65194fd33307a58707dd2de827ac7cf9e1097d04a7cea4ca7a16R122-R127) is the case.

That informs this one. Basically, it makes little sense to wait for both of those `isInitialized` signals (because the collection can be empty) and it makes little sense to have a timeout (waiting is suspend so it can be kept around waiting forever in the worst case, though it won't). This means that no renewal is attempted with no sources, which implies `isInitialized`.

Going back [to the change further up](https://github.com/mihonapp/mihon/pull/2970/changes#diff-4880ffb9903e65194fd33307a58707dd2de827ac7cf9e1097d04a7cea4ca7a16R122-R127), there is still the chance to only have a partial list (due to Android weirdness), so the job needs to be interrupted and immediately replaced with the new one once Android informs the app of the rest of the packages.

Lolle2000la: TLDR. Without it and the retry watcher above, it can actually try to index nothing based on nothing. The bulk of what was done here is now above.

MajorTanya:
@AntsyLich Did you not read what I said

AntsyLich:
It's fine to test this in beta

PR #2971, opened 2026-02-17
Add feature to hide similar pages

LetianLi:
Adds the ability to hide images; all similar images will be hidden/minimized in the same way.

Features:
- Long press menu adds an in-sheet image preview, a second action button row, and buttons to: Hide image forever, Show image, Unhide forever.
- Hidden images are replaced with a placeholder UI element saying "Image hidden" and a show button. The UI elements are compressed side by side in webtoon reader, and vertical like ErrorLayout in pager reader.
- Matching is done by SHA-256, then a perceptual dHash for visual similarity. Depending on which page was triggered, matching can be scoped to Start/End/Any page for efficency.
- Hidden image rules are stored per manga
- The manga's overflow action dropdown (three dots) has a new "Manage hidden images" option that opens a full screen managing page
- Manage page is a gallery of compressed preview images, where rules can be cleared, selected, removed, re-scoped. Clicking the thumbnails renders a manga-cover-like preview, except we stored it compressed to save space.
- Settings/Advanced/Data has an additional option to erase hidden image rules.

Bug fixes:
- Long press menu items (Save, Set as Cover, Share) failing when stream is evicted from cache: Added a check when opening this menu which will refetch page if needed.
- Scrolling up sometimes doesn't fetch previous chapter if the preloaded 2 pages are short: Made WebtoonViewer.onScrolled direction aware.

### Images
| Device| Long Press Menu Hide | Long Press Menu Show | Paged Hidden | Long Strip Hidden | 
| -- | ------- | ------- | -- | -- |
| Phone | Phone_Hold_Hide | Phone_Hold_Show | Phone_PagedHidden | Phone_LongHidden_Multiple |
| Tablet | Tablet_Hold_Hide | Tablet_Hold_Show | Tablet_PagedHidden | Tablet_LongHidden |

| Overflow Action | Manager | Batch select | Full screen thumbnail |
| -- | -- | -- | -- |
| Tablet_Manga_Overflow | Tablet_Manager_and_Overflow | Tablet_Manager_BatchSelect | Tablet_Manager_FullScreen |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/manga/hiddenimages/HiddenImagesScreen.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/manga/hiddenimages/HiddenImagesScreen.kt
@@ -253,7 +253,7 @@ class HiddenImagesScreen(
                     onLongClick = onLongClick,
                 ),
         ) {
-            Column(verticalArrangement = Arrangement.spacedBy(8.dp)) {
+            Column(verticalArrangement = Arrangement.spacedBy(0.dp)) {
                 if (imageBitmap != null) {
                     Image(
                         bitmap = imageBitmap.asImageBitmap(),
@@ -278,7 +278,7 @@ class HiddenImagesScreen(
                 Row(
                     modifier = Modifier
                         .fillMaxWidth()
-                        .padding(horizontal = 4.dp, vertical = 4.dp),
+                        .padding(horizontal = 4.dp, vertical = 0.dp),
                     horizontalArrangement = Arrangement.SpaceBetween,
                     verticalAlignment = Alignment.CenterVertically,
                 ) {


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderViewModel.kt
```diff
@@ -831,6 +831,9 @@ class ReaderViewModel @JvmOverloads constructor(
 
     fun openPageDialog(page: ReaderPage) {
         mutableState.update { it.copy(dialog = Dialog.PageActions(page)) }
+        if (!canOpenPageStream(page)) {
+            requestPageReload(page)
```
LetianLi, line 835: Sometimes the underlying image was evicted (especially when going between chapters) even though it is still visible in the reader. This caused the page stream to stop working, and Share/Save/Set as cover/Preview all failed as a result. 

Here we refetch it, and reload it to fix this whenever the dialog is opened.

LetianLi: There's a small side effect which causes the layout to temporarily shift when it does the reload. But I don't know how to fix it. 

At the very least, it's better than before where Share/Save/Cover all failed silently.

cuong-tran:
why need to hide images and replace with an annoying placeholder?

LetianLi:
Why hide:
- Some sources have like 11 pages of ads at the start of each chapter (cough demonic scans cough). So I made this feature to try and hide them. Unfortunately it doesn't work for that source since they concat the ad pages with necessary pages, but it does in other simpler cases.

Why placeholder:
- easiest way to surface unhiding.
- dhash may have false positives.
- I don't expect this feature to be used super often. Just for intrusive pages.
- in long strip mode it's not annoying. In paged modes, I can see it being annoying.
- I tried completely hiding, and didn't like a page completely disappearing.

Why per-manga:
- easier to store and manage compared to per-source.
- discourage frequent use.

AntsyLich:
I'm not really interested in adding this feature

PR #2972, opened 2026-02-17
Fix first update not waiting for extension loading

Lolle2000la:
Fixes [this error](https://github.com/mihonapp/mihon/pull/2955#issuecomment-3912419537).

This ensures that the update starts once initialization is finished, making the dependency explicit and ensuring correct ordering.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
@@ -121,7 +121,7 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
 
         extensionManager.isInitialized.filter { it }.first()
 
-        // If the job starts within the first 10 seconds of launch, wait out the remainder
+        // If the job starts within the first 5 seconds of launch, wait out the remainder
         // to ensure Android has finished binding all extensions
         val elapsed = System.currentTimeMillis() - App.processStartTimeMillis
         val remaining = 5_000L - elapsed


Lolle2000la:
@MajorTanya I think this should fix that error!

MajorTanya:
I've been dailying #2970 to aid with debugging but without this specific change in it and haven't seen a failed update due to "Source not found" during that time. I'm guessing whatever was the actual cause was either a fluke or has been covered by that PR. I'll keep an eye on my update runs for the next few iterations but unless I see it again, I believe this isn't needed (anymore?).

Lolle2000la:
Thank you for testing! In that case I think I will convert this PR to draft status and eventually close it outright if that remains the case.

Lolle2000la:
@MajorTanya tested without these changes and it appears that this PR is still necessary

MajorTanya:
Okay, I've made a build of this and installed it, will report back in ~13 hours when my library has gone through some update cycles from the background.

app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
```diff
@@ -119,6 +121,12 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
 
         extensionManager.isInitialized.filter { it }.first()
 
+        // If the job starts within the first 10 seconds of launch, wait out the remainder
+        // to ensure Android has finished binding all extensions
+        val elapsed = System.currentTimeMillis() - App.processStartTimeMillis
+        val remaining = 5_000L - elapsed
```
MajorTanya: Is it 10 or 5 seconds? The `5_000L` makes me think 5

Lolle2000la: Sorry, I started with 10 seconds but 5 should suffice

Lolle2000la:
Probably better solved by #2984

PR #2974, opened 2026-02-18
Created a feature that displays the chapter's list with a long press …

ArjiAmin:
This adds a feature that will display the chapter's list when long pressing a navigation button.

Quick notes :
- Haptic feedback is included, couldn't try it on device though but i tried to compare debugs with haptic feedback elsewhere and it showed no issues with it.
- "FilledIconButton" was replaced in "ChapterNavigator.kt" because no onLongClick was present.
- Removing the download buttons was a bit of a necessity, because i for some reasons could not get the download icons to do anything at all, they would just exist and i would not be able to interact with them, so i've simply removed them altogether.
- I tried to add a scroll bar for better UX, but it would go out of bounds so i gave up.
- List uses user's sorting and naming configurations.

https://github.com/user-attachments/assets/cb1fde44-c15c-44db-bf45-b514142538aa

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/manga/components/MangaChapterListItem.kt
+++ b/app/src/main/java/eu/kanade/presentation/manga/components/MangaChapterListItem.kt
@@ -54,6 +54,7 @@ fun MangaChapterListItem(
     read: Boolean,
     bookmark: Boolean,
     selected: Boolean,
+    showDownloadIndicator: Boolean = true,
     downloadIndicatorEnabled: Boolean,
     downloadStateProvider: () -> Download.State,
     downloadProgressProvider: () -> Int,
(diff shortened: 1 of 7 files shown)


ArjiAmin:
Soooo... Any chance of this ever being reviewed and accepted/refused ?

Smol-Ame:
There is no ETA or guarantee for any request, issue or proposed change, to be worked on, reviewed, or approved.

Wait patiently for the dev's decision on if & when they want to implement your PR.

cuong-tran:
Long press is too hidden. Create a button like Komikku/SY.

ArjiAmin:
> There is no ETA or guarantee for any request, issue or proposed change, to be worked on, reviewed, or approved.
> 
> Wait patiently for the dev's decision on if & when they want to implement your PR.

True, i was just wondering if this had been silently ignored or if i needed to do some specific changes for it to be approved

ArjiAmin:
> Long press is too hidden. Create a button like Komikku/SY.

I just downloaded Komikku for the first time, i love the sidebar page scroll.
And i noticed whoever implemented the chapters list figured out the download button lol.

Anyways, i can't exactly place a button at the bottom due to space limits

But what i can do instead is keep the long press as it is, but add a "show chapters list" option in the top right select button.
I also thought of perhaps adding a sort of small notification badge right above the Next button that displays the remaining chapters, and also serves the same function as the long press on Next button

image

PR #2976, opened 2026-02-20
Remove userSelected from range-selection

cuong-tran:
Probably a remnant of the old days


MajorTanya:
What was this used for in the past?

cuong-tran:
> What was this used for in the past?

The first and only commit was 00519e3b933b66bb764f1f58f79fb781b0cb092a
The author probably intended to reuse the method for some other selection methods, but it never happened. All of the function's usages always have `userSelected` as `true`

MajorTanya:
> > What was this used for in the past?
> 
> The first and only commit was [00519e3](https://github.com/mihonapp/mihon/commit/00519e3b933b66bb764f1f58f79fb781b0cb092a) The author probably intended to reuse the method for some other selection methods, but it never happened. All of the function's usages always have `userSelected` as `true`

🤔 I see.

PR #1997, opened 2025-04-12
Duplicate managment functionality

NarwhalHorns:
Adds a 'Duplicate Management' page that searches for and displays any potential duplicates found in the user's library for management.

Possible Duplicates Tab - display potential dupes to remove/hide
- Currently gets all library manga and searches for duplicates on each one using a simple fuzzy search of the title {% title %}
- Displays any duplicates found against each manga
- Each entry can be unfavorited or 'hidden' as a duplicate
	- Key manga (manga checked against) can be hidden to hide all duplicates found against it at once.

Hidden Duplicates Tab - manage duplicates marked as hidden in case they need changing
- Lists all hidden duplicates against each entry
- Each can be 'unhidden' to put them back in normal duplicate search
	- KeyManga can be unhidden to unhide whole row of dupes against it

General
- Shows chapter count on entries
- Can backup and restore hidden duplicates

| Possible Duplicates Tab | Hidden Duplicates Tab |
| ----- | ----- |
| image | image |

| Change automatic duplicate search match level for when adding new manga | Change duplicate search match level on possible duplicates screen |
| ------ | ----- |
| image | image |

| Confirm removing manga and chose to delete downloads |
| ---- |
| image |

| Empty Possibles screen | Empty hidden dupes screen |
| ----- | ---- |
| image | image |

Problems
- [x] Fuzzy search just finds too much shit. (e.g. entries with single common word like 'ring' find ANYTHING containing substring 'ring' which is just insane.)
	- [x] Maybe multiple levels of search depth? i.e. exact match, word fuzzy, full substring fuzzy.
	- [x] ~~Probably needs more complex search algorithm~~ todo later if desired
	- [x] ~~Perhaps configurable percentage match threshold~~ todo later if desired

TODO:
- [x] Remove favorite popup check
	- [x] check delete downloads on remove
- [x] Reload tabs on change (not sure how to do this, scrapped together the tab functionality so might need re-working for that)
- [x] Backup and restore hidden duplicates
- [x] Probably wanna make them smaller/change the list design a bit, feels too packed atm?
	- [x] Make duplicate list item size configurable (by width?)
- [x] Some logic to not show duplicate (hah) duplicate lists i.e. when id1 = id2. it is listed one for each one
- [x] Sort duplicate list order
	- [x] alphabetically
	- [x] duplicate count

PR #2978, opened 2026-02-20
Duplicate manga search by tracks

NarwhalHorns:


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Reword download reindexing message to avoid confusion ([@MajorTanya](https://github.com/MajorTanya)) ([#2874](https://github.com/mihonapp/mihon/pull/2874))
 - Rework internals for better performance ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2955](https://github.com/mihonapp/mihon/pull/2955))
 - Optimize tracked library filter ([@NarwhalHorns](https://github.com/NarwhalHorns)) ([#2977](https://github.com/mihonapp/mihon/pull/2977))
+- Duplicates search includes tracks ([@NarwhalHorns](https://github.com/NarwhalHorns)) ([#2977](https://github.com/mihonapp/mihon/pull/2978))
```
```suggestion
- Utilize tracker for library duplicate detection ([@NarwhalHorns](https://github.com/NarwhalHorns)) ([#2977](https://github.com/mihonapp/mihon/pull/2978))
```

NarwhalHorns:
> Do people set up tracking before adding manga to library?

not usually I image, but it would help if that was to happen, or if attempting to add something that was previous tracked like if it was in the library before. OR more interestingly, if we go forward with my duplicates manager this will be very useful for that hehe

AntsyLich:
> [...] if we go forward with my duplicates manager this will be very useful for that hehe
>
So when is that happening

NarwhalHorns:
> > [...] if we go forward with my duplicates manager this will be very useful for that hehe
> 
> So when is that happening

Whenever you like :sip: [PR is ready](https://github.com/mihonapp/mihon/pull/1997)

Shirasho:
> Do people set up tracking before adding manga to library?

For what it is worth, I do more often than not. I tend to dislike reading stuff I can't find a tracker entry for due to my OCD. The only time I add without checking for a tracker entry is for series that have less than 5 chapters out since my tracker of choice only adds series that are somewhat well established.

Whatever the case may be, I am happy to have more features that can better associate one series with another, especially in the case of source migrations.

PR #2982, opened 2026-02-22
Prevent db lock contention

Lolle2000la:
This replaces the lock I introduced in #2951 and #2955 by instead using a single IO thread to move these interactions off to.

This means that there is no lock to content anymore, still yields (so no thread starvation) and fully prevents blocking of the main thread, so the app always remains responsive during DB interactions.

It should help users that saw regressions from the prior changes.

### DIFF
--- a/data/src/main/java/tachiyomi/data/TransactionContext.kt
+++ b/data/src/main/java/tachiyomi/data/TransactionContext.kt
@@ -1,30 +1,24 @@
 package tachiyomi.data
 
 import kotlinx.coroutines.CoroutineDispatcher
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.asContextElement
+import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.runBlocking
-import kotlinx.coroutines.suspendCancellableCoroutine
-import kotlinx.coroutines.sync.Mutex
-import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
-import java.util.concurrent.RejectedExecutionException
-import kotlin.concurrent.atomics.AtomicInt
-import kotlin.concurrent.atomics.ExperimentalAtomicApi
-import kotlin.concurrent.atomics.decrementAndFetch
-import kotlin.concurrent.atomics.incrementAndFetch
 import kotlin.coroutines.ContinuationInterceptor
 import kotlin.coroutines.CoroutineContext
-import kotlin.coroutines.EmptyCoroutineContext
 import kotlin.coroutines.coroutineContext
-import kotlin.coroutines.resume
-
-// Global mutex to serialize transaction entry and prevent thread pool exhaustion.
-// If you have multiple distinct database files/handlers, this should be a property of AndroidDatabaseHandler.
-private val transactionMutex = Mutex()
 
 /**
- * Returns the transaction dispatcher if we are on a transaction, or the database dispatchers.
+ * A single-threaded dispatcher backed by the IO thread pool.
+ * Allows yielding while still putting the SQLite interaction off-thread.
+ *
+ * This still keeps serialization, keeps the calling thread unblocked,
+ * but doesn't suffer from lock contention.
+ */
+private val transactionDispatcher = Dispatchers.IO.limitedParallelism(1)
+
+/**
+ * Returns the transaction dispatcher if we are on a transaction, or the database dispatcher.
  */
 internal suspend fun AndroidDatabaseHandler.getCurrentDatabaseContext(): CoroutineContext {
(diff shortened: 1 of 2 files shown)


Lolle2000la:
As can be seen in the diff, this has the nice side effect of being massively simpler.

data/src/main/java/tachiyomi/data/TransactionContext.kt
```diff
@@ -1,30 +1,24 @@
...
-// If you have multiple distinct database files/handlers, this should be a property of AndroidDatabaseHandler.
-private val transactionMutex = Mutex()
+/**
+ * A single-threaded dispatcher backed by the IO thread pool.
+ * Allows yielding while still putting the SQLite interaction off-thread.
+ *
+ * This still keeps serialization, keeps the calling thread unblocked,
+ * but doesn't suffer from lock contention.
+ */
+private val transactionDispatcher = Dispatchers.IO.limitedParallelism(1)
```
FlaminSarge, line 18: Why create this here instead of passing it into the original AndroidDatabaseHandler constructor (before you removed the transactionDispatcher param on it)?

Lolle2000la: The reason is that the sqlite transaction "eats" a thread till done, however, if one runs multiple threads they will be stuck waiting for the internal lock in sqlite, which serializes those write transactions. My previous approach was to reduce that by adding a outer mutex that can yield and thus minimizes the amount of time a thread is eaten, but unless the DB itself gets a lot quicker, threads (often not IO threads) will enter here and will then be unable to do anything else until sqlite is done.

This approach basically completely farms it out. Since it is only one thread it will also not run into write contention and serializes naturally. The calling thread can then yield and respond to other things (like button presses, opening a library entry etc.). 

Basically, this replaces the mutex and basically acts like a "channel of transactions", so its a different thing in purpose.

FlaminSarge: That's not what I'm asking, getting rid of the mutex and switching to this approach is the correct play.

I'm asking why the AndroidDatabaseHandler parameter for transactionDispatcher was removed instead of just changing its default value from queryDispatcher to `Dispatchers.IO.limitedParallelism(1)` (or setting it as the 3rd param passed to AndroidDatabaseHandler in AppModule.kt)? Does making it a private val inside TransactionContext do something on top of that (or allow for some extra code cleanup)?

Lolle2000la: > I'm asking why the AndroidDatabaseHandler parameter for transactionDispatcher was removed instead of just changing its default value from queryDispatcher to Dispatchers.IO.limitedParallelism(1) (or setting it as the 3rd param passed to AndroidDatabaseHandler in AppModule.kt)? Does making it a private val inside TransactionContext do something on top of that (or allow for some extra code cleanup)?

Basically, no other dispatcher is allowed to handle this here since we only ever want this one dedicated thread to do it. Making it the default param would also create a new limited parallelism dispatcher for each request (effectively defeating the point of replacing the mutex), though it could be bypassed by adding it as a private val and defaulting to that if nothing was passed.

But the point is that no other dispatcher should be allowed and we are offloading to this dedicated thread that is allowed to be blocked by sqlite.

FlaminSarge: > Making it the default param would also create a new limited parallelism dispatcher for each request

I'm don't believe this is correct, because it'd mean that we're creating a new AndroidDatabaseHandler for every single request rather than once at App start (looks like AppModule creates it as a Singleton, and I very much doubt it re-injects the two dispatchers every single time its methods are called).

However, I agree that in any case we need to limit it to this specialized dispatcher, and removing the param probably prevents (somewhere down the line) someone seeing that param, thinking they can use it, and shooting our collective feet. (That said, I'm relatively certain it could have been passed as the default param and still work fine, given the singleton thing).

PR LGTM in any case, good work.

Lolle2000la: Yeah, sorry! I wasn't completely here with my mind when I answered. Thinking about it, if making it a default parameter would be a problem then the current way would also be since it isn't static, so that was a bit of a brain fart 😅

AntsyLich:
The old TransactionContext was based on androidx's room's implementation. You might want to have a look what they do currently https://github.com/androidx/androidx/blob/4f1927c2c3b66d0c3a6b9118974d818d2dc5a06a/room3/room3-runtime/src/androidMain/kotlin/androidx/room3/coroutines/ThreadConfinedTransactionContext.android.kt as I would like to have a more tested implementation.

AntsyLich:
The offending PR was reverted and our own custom transaction management will be removed by #3270

PR #2983, opened 2026-02-22
Download queue isn't properly persisted and restored across runs

Lolle2000la:
Basically two bugs were happening:

1. The same thing we had with the library update and the index invalidation: The restoration happened too close to launch, the sources weren't populated fast enough.
2. Rarer and possible before: The queue wasn't immediately persisted, meaning that if Android kills the process immediately after an update, the list might not have been written yet.

This fixes both.


Lolle2000la:
Better solved by #2984

PR #2984, opened 2026-02-22
Fix extension loading reliably at its root

Lolle2000la:
Querying installed packages too close to app launch is impossible. By waiting for 5 seconds before it, we reliably get the full extension list. This also gives us a reliable initialization event, which we can use to easily prevent a number of prior bugs, like:

* (Automatic) Updates don't work (obsoleting #2972)
* Download queue isn't restored correctly (obsoleting #2983)
* Download indexing isn't performed correctly (actually better solved here than in #2970, reverted that change here for that reason).

As long as all code that needs extensions and sources properly expresses its dependency to the `ExtensionManager` (or another component that has such a dependency like `(Android)SourceManager`), there will no longer be any of those issues during startup.

Please let me know if I missed something, but this should be much more reliable than individually doing something similar everywhere.

Note that #2982 is still necessary regardless for other reasons.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/extension/ExtensionManager.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/extension/ExtensionManager.kt
@@ -135,7 +135,7 @@ class ExtensionManager(
                     stabilityCounter = 0
                 }
                 lastCount = currentCount
-                if (i < maxChecks - 1) delay(250)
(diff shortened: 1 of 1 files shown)


MajorTanya:
I've been once again dailying these changes since March 5th, excluding the changes of 2aec547f3b0676d78a8c025bdf8cda9aa46b7170 (which look like they should just make it slightly snappier), and they seem fine to me.

PR #2986, opened 2026-02-22
Implement parallel page loading and preloading in reader

AntsyLich:
Also:
- Now full chapter is loaded in the background
- Retry in reader redownloads the image


BrutuZ:
If the chapter is a high quality volume could pre-loading it entirely bust the cache?

MajorTanya:
> If the chapter is a high quality volume could pre-loading it entirely bust the cache?

Forget busting the cache, it could bust the available memory on the device. Recipe for OOM en masse.

PR #3001, opened 2026-02-26
Replace java.time APIs with kotlin(x).(date)time

### LINKED ISSUE #1947: Use 'kotlin.time.Instant' and 'kotlinx-datetime' in the app

Self explanatory

### Other details

_No response_

### Acknowledgements

- [x] I have updated the app to version **[0.18.0](https://github.com/mihonapp/mihon/releases/latest)**.

MajorTanya:
kotlinx-datetime [explicitly says in its README](https://github.com/Kotlin/kotlinx-datetime?tab=readme-ov-file#design-overview) that it does not cover i18n, so we have to stick with using their toJavaX converters or plain `java.time.X` wherever we localise dates and times.

Yes, some of these replacements are... questionable. But I wanted to try and maximise the replacement for now.

Closes #1947.

I'll add some comments for things I am not entire sure about.


app/proguard-rules.pro
```diff
@@ -80,3 +80,6 @@
 # Firebase
 -keep class com.google.firebase.installations.** { *; }
 -keep interface com.google.firebase.installations.** { *; }
+
+# KotlinX Datetime
+-keep,allowoptimization class kotlinx.datetime.** { public protected *; }
```
MajorTanya, line 91: Not sure if this is needed, but I saw it exists for `kotlinx.coroutines` and `kotlin.time`, so I copied it over

data/build.gradle.kts
```diff
@@ -34,5 +34,7 @@ dependencies {
     implementation(projects.domain)
     implementation(projects.core.common)
 
+    implementation(kotlinx.datetime)
```
MajorTanya: Note: Added `kotlinx.datetime` as a dependency for the `data` module.

domain/src/test/java/tachiyomi/domain/manga/interactor/FetchIntervalTest.kt
```diff
@@ -2,27 +2,27 @@ package tachiyomi.domain.manga.interactor
...
 import kotlin.time.DurationUnit
 import kotlin.time.toDuration
-import kotlin.time.toJavaDuration
 
 @Execution(ExecutionMode.CONCURRENT)
 class FetchIntervalTest {
 
-    private val testTime = ZonedDateTime.parse("2020-01-01T00:00:00Z")
-    private val testZoneId = ZoneOffset.UTC
+    private val testTime = LocalDateTime.parse("2020-01-01T00:00:00")
```
MajorTanya, line 21: The parser throws if zone information is present

domain/build.gradle.kts
```diff
@@ -27,6 +27,8 @@ dependencies {
     implementation(kotlinx.bundles.coroutines)
     implementation(kotlinx.bundles.serialization)
 
+    implementation(kotlinx.datetime)
```
MajorTanya: Note: Added `kotlinx.datetime` as a dependency for the `domain` module.

presentation-widget/build.gradle.kts
```diff
@@ -24,6 +24,8 @@ dependencies {
 
     implementation(kotlinx.immutables)
 
+    implementation(kotlinx.datetime)
```
MajorTanya: Note: Added `kotlinx.datetime` as a dependency for the `presentation-widget` module.

app/src/main/java/eu/kanade/domain/manga/interactor/UpdateManga.kt
```diff
@@ -91,25 +94,31 @@ class UpdateManga(
 
     suspend fun awaitUpdateFetchInterval(
         manga: Manga,
-        dateTime: ZonedDateTime = ZonedDateTime.now(),
-        window: Pair<Long, Long> = fetchInterval.getWindow(dateTime),
+        timeZone: TimeZone = TimeZone.currentSystemDefault(),
+        dateTime: LocalDateTime = Clock.System.now().toLocalDateTime(timeZone),
+        window: Pair<Long, Long> = fetchInterval.getWindow(dateTime.date, timeZone),
```
MajorTanya, line 29: Not a mistake that these are in a different order to literally everywhere else. To avoid querying the timezone twice and to be able to use the parameter for the default value of `dateTime`, I had to make it initialise before that. Probably the biggest wart of these changes and could be swapped around, but losing the default parameter elegance.

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/MangaRestorer.kt
```diff
@@ -34,12 +36,14 @@ class MangaRestorer(
...
+    private var timeZone = TimeZone.currentSystemDefault()
+    private var now = Clock.System.now().toLocalDateTime(timeZone)
+    private var currentFetchWindow = fetchInterval.getWindow(now.date, timeZone)
 
     init {
-        now = ZonedDateTime.now()
-        currentFetchWindow = fetchInterval.getWindow(now)
+        timeZone = TimeZone.currentSystemDefault()
+        now = Clock.System.now().toLocalDateTime(timeZone)
+        currentFetchWindow = fetchInterval.getWindow(now.date, timeZone)
```
MajorTanya: This code confused me. The init block seems to set the vars to the same values as the initial values -- is this redundant or am I missing some finer details of the restore process? I didn't see any reasoning in the commit history for this section.

app/src/main/java/eu/kanade/tachiyomi/util/lang/DateExtensions.kt
```diff
@@ -26,21 +32,21 @@ fun Date.toTimestampString(): String {
...
+    from: TimeZone,
+    to: TimeZone,
 ): Long {
-    return LocalDateTime.ofInstant(Instant.ofEpochMilli(this), from)
-        .atZone(to)
-        .toInstant()
-        .toEpochMilli()
+    return Instant.fromEpochMilliseconds(this)
+        .toLocalDateTime(from)
+        .toInstant(to)
```
MajorTanya, line 40: Yes this converts between zones. Internally, the `toInstant` method uses `atZone(to)` on the JVM implementation.

app/src/main/java/mihon/feature/upcoming/components/calendar/Calendar.kt
```diff
@@ -77,8 +80,8 @@ private fun CalendarGrid(
             .toImmutableList()
     }
 
-    val emptyFieldCount = weekDays.indexOf(selectedYearMonth.atDay(1).dayOfWeek)
-    val daysInMonth = selectedYearMonth.lengthOfMonth()
+    val emptyFieldCount = weekDays.indexOf(selectedYearMonth.firstDay.dayOfWeek.toJavaDayOfWeek())
```
MajorTanya, line 80: Days of the week are an i18n matter, so we have to fall back to Java's stuff here (thanks to people who wrongly insist that the week starts on Sunday)

app/build.gradle.kts
```diff
@@ -208,6 +208,8 @@ dependencies {
     implementation(platform(kotlinx.coroutines.bom))
     implementation(kotlinx.bundles.coroutines)
 
+    implementation(kotlinx.datetime)
```
MajorTanya: Note: Added `kotlinx.datetime` as a dependency for the `app` module.

MajorTanya (commented):
Please take a look at some of my notes.

MajorTanya:
The only replaceable occurrence of java.time that isn't i18n lives here:
https://github.com/mihonapp/mihon/blob/bc3230832704116b7eb0e5c3ad3edf4656d998fd/gradle/build-logic/src/main/kotlin/mihon/gradle/Commands.kt#L21-L35

Not sure what'd be the best way to replace this, since the format chosen excluded the fractional seconds that'd come from using the basic toString function on a Kotlin Instant.

Edit: this is what it'd look like:
```patch
Index: gradle/build-logic/src/main/kotlin/mihon/gradle/Commands.kt
IDEA additional info:
<+>UTF-8
diff --git a/gradle/build-logic/src/main/kotlin/mihon/gradle/Commands.kt b/gradle/build-logic/src/main/kotlin/mihon/gradle/Commands.kt
--- a/gradle/build-logic/src/main/kotlin/mihon/gradle/Commands.kt	(revision 54e9d3893acbc5a69e19ab6b0aef192217e564bc)
+++ b/gradle/build-logic/src/main/kotlin/mihon/gradle/Commands.kt	(date 1774127394545)
@@ -1,10 +1,9 @@
 package mihon.gradle
 
 import org.gradle.api.Project
-import java.time.Instant
-import java.time.LocalDateTime
-import java.time.ZoneOffset
-import java.time.format.DateTimeFormatter
+import kotlin.time.Clock
+import kotlin.time.Duration.Companion.nanoseconds
+import kotlin.time.Instant
 
 // Git is needed in your system PATH for these commands to work.
 // If it's not installed, you can return a random value as a workaround
@@ -18,19 +17,18 @@
     // return "1"
 }
 
-private val BUILD_TIME_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'")
-
 /**
  * @param useLatestCommitTime If `true`, the build time is based on the timestamp of the last Git commit;
  *                          otherwise, the current time is used. Both are in UTC.
- * @return A formatted string representing the build time. The format used is defined by [BUILD_TIME_FORMATTER].
+ * @return A formatted string representing the build time.
  */
 fun Project.getBuildTime(useLatestCommitTime: Boolean): String {
     return if (useLatestCommitTime) {
         val epoch = exec("git log -1 --format=%ct").toLong()
-        Instant.ofEpochSecond(epoch).atOffset(ZoneOffset.UTC).format(BUILD_TIME_FORMATTER)
+        Instant.fromEpochSeconds(epoch).toString()
     } else {
-        LocalDateTime.now(ZoneOffset.UTC).format(BUILD_TIME_FORMATTER)
+        val instant = Clock.System.now()
+        instant.minus(instant.nanosecondsOfSecond.nanoseconds).toString()
     }
 }
 
```

Not quite as horrendous as I expected but I'm unsure nevertheless

Kotlin Playground link to show that this would give the desired format:
https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMi4zLjIwIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiaW1wb3J0IGtvdGxpbi50aW1lLkNsb2NrXG5pbXBvcnQga290bGluLnRpbWUuRHVyYXRpb24uQ29tcGFuaW9uLm5hbm9zZWNvbmRzXG5pbXBvcnQga290bGluLnRpbWUuSW5zdGFudFxuXG5mdW4gbWFpbigpIHtcbiAgICB2YWwgaW5zdGFudCA9IENsb2NrLlN5c3RlbS5ub3coKVxuICAgIHByaW50bG4oaW5zdGFudC5taW51cyhpbnN0YW50Lm5hbm9zZWNvbmRzT2ZTZWNvbmQubmFub3NlY29uZHMpLnRvU3RyaW5nKCkpXG59In0=

MajorTanya:
> The only replaceable occurrence of java.time that isn't i18n lives here:
> 
> https://github.com/mihonapp/mihon/blob/bc3230832704116b7eb0e5c3ad3edf4656d998fd/gradle/build-logic/src/main/kotlin/mihon/gradle/Commands.kt#L21-L35
> 
> Not sure what'd be the best way to replace this, since the format chosen excluded the fractional seconds that'd come from using the basic toString function on a Kotlin Instant.
> 

Update: I have applied a change that's basically this. `kotlin.time.Instant.toString()` returns the ISO 8601 formatted string just like the old `BUILD_TIME_FORMATTER` did ([docs](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-instant/to-string.html))

MajorTanya:
- d176781 shows the additional changes
- I updated kotlinx.datetime to 0.8.0 which released in May

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/MangaRestorer.kt
AntsyLich: it's redundant

AntsyLich (commented):
> so we have to stick with using their toJavaX converters or plain java.time.X wherever we localise dates and times.
>
can we have a doc somewhere documenting the cases so we can come back to it later?

MajorTanya:
> > so we have to stick with using their toJavaX converters or plain java.time.X wherever we localise dates and times.
> 
> can we have a doc somewhere documenting the cases so we can come back to it later?

Would it be sufficient if I just list them as a comment here or do you want some document in the source somewhere?

AntsyLich:
> > > so we have to stick with using their toJavaX converters or plain java.time.X wherever we localise dates and times.
> > 
> > 
> > can we have a doc somewhere documenting the cases so we can come back to it later?
> 
> Would it be sufficient if I just list them as a comment here or do you want some document in the source somewhere?
>
comment is also fine

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/MangaRestorer.kt
MajorTanya: removed :+1:

app/src/main/java/mihon/feature/upcoming/components/calendar/Calendar.kt
MajorTanya: outside of scope, see #3649

PR #3010, opened 2026-02-28
.

ghost:


MajorTanya:
Excuse you but what the fuck is this? Downgrades an action, reverts string resources to a much older version, and DELETES THE LICENCE? Go away with this spam.

Smol-Ame:
Prime example of when you don't understand how things work, and don't know what you're doing/changing.

PR #3030, opened 2026-03-03
fix: enqueue duplicate chapters for deletion after marking as read

### LINKED ISSUE #2454: Duplicate chapters marked as read are not automatically deleted
### Steps to reproduce

1. Download both versions of a chapter from the same source
2. Fully read one of them (marking as read just mark one of them as read and deletes it properly as well)
3. Both chapters will be marked as read if "Mark duplicate read chapter as read" is properly set, but the one you didn't actually read will not be deleted, only the one you did.

### Expected behavior

Both chapters should be deleted at that point, since they are already marked as read anyway.

### Actual behavior

The one that was actually read is properly deleted, while the other one isn't.

Before:

- [x] Chapter 174 - Scan A
- [x] Chapter 174 - Scan B

After:

- [ ] ~~Chapter 174 - Scan A~~
- [x] ~~Chapter 174 - Scan B~~

### Crash logs

_No response_

### Mihon version

0.19.1 (029e36b, 13, 2025-08-07T13:59:01Z)

### Android version

15 (SDK 35; build V1UWS35H.63-14-6-4-2-3)

### Device

Motorola Edge 50

### Other details

Image

https://github.com/user-attachments/assets/3429c662-e046-4bcc-8fd8-189d03a67e9f

### Acknowledgements

leodyversemilla07:
## Summary

Fixes #2454

## Root Cause

In `updateChapterProgressOnComplete()`, when duplicate chapters are auto-marked as read (via the **'Mark duplicate chapters as read'** setting), `enqueueChaptersToDelete()` was never called for them.

The code used `mapNotNull { ChapterUpdate(...) }` which immediately converted full `Chapter` objects into lightweight `ChapterUpdate` objects (only `id` + `read` flag). This discarded the full chapter data required by `downloadManager.enqueueChaptersToDelete()`, and the delete call was simply missing from this code path.

## Change

**File:** `app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderViewModel.kt`

Replaced `mapNotNull { ChapterUpdate }` with `filter { Chapter }` to retain full `Chapter` objects, then:
- Mark duplicates read via inline `.map { ChapterUpdate(...) }` — identical behavior as before
- After marking read, enqueue downloaded duplicates for deletion via `downloadManager.enqueueChaptersToDelete()`, using the same `removeAfterReadSlots != -1` guard that protects the existing `deleteChapterIfNeeded()` call

```kotlin
// Before
val duplicateUnreadChapters = getUnfilteredChapterList()
    .mapNotNull { chapter ->
        if (!chapter.read && chapter.isRecognizedNumber &&
            chapter.chapterNumber.toFloat() == readerChapter.chapter.chapter_number
        ) {
            ChapterUpdate(id = chapter.id, read = true)
        } else { null }
    }
updateChapter.awaitAll(duplicateUnreadChapters)

// After
val duplicateChapters = getUnfilteredChapterList()
    .filter { chapter ->
        !chapter.read &&
            chapter.isRecognizedNumber &&
            chapter.chapterNumber.toFloat() == readerChapter.chapter.chapter_number
    }
updateChapter.awaitAll(duplicateChapters.map { ChapterUpdate(id = it.id, read = true) })

if (downloadPreferences.removeAfterReadSlots().get() != -1) {
    val manga = manga ?: return
    downloadManager.enqueueChaptersToDelete(duplicateChapters, manga)
}
```


app/src/main/java/eu/kanade/tachiyomi/ui/reader/ReaderViewModel.kt
```diff
@@ -577,19 +577,19 @@ class ReaderViewModel @JvmOverloads constructor(
...
-                    null
-                }
             }
-        updateChapter.awaitAll(duplicateUnreadChapters)
+        updateChapter.awaitAll(duplicateChapters.map { ChapterUpdate(id = it.id, read = true) })
+
+        // Also delete downloaded duplicate chapters if "delete after read" is enabled
+        if (downloadPreferences.removeAfterReadSlots().get() != -1) {
+            val manga = manga ?: return
+            downloadManager.enqueueChaptersToDelete(duplicateChapters, manga)
```
AntsyLich: This doesn't account for the value of `downloadPreferences.removeAfterReadSlots()`

Issue #3020, opened 2026-03-02
Add an option to hide chapters like 1.1 , 1.2 , they are basically same as chapter 1 mostly


imDusht:
### Describe your suggested feature

Add an option so that i can read next chapter after finishing previous one instead of middle ones

<img width="864" height="1920" alt="Image" src="https://github.com/user-attachments/assets/920e82e9-3755-40fa-9def-02a01ebe519f" />

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.19.4](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

SecretX33:
I went ahead and decided to give this a shot: [PR](https://github.com/mihonapp/mihon/pull/3034) :)

Muhammedahmed18:
> I went ahead and decided to give this a shot: [PR](https://github.com/mihonapp/mihon/pull/3034) :)

you did a really great job, now need to hope that they accept this

aarabdh:
Bump. Please include this feature if it is ready for release. Very essential feature for some mangas. Thanks!

PR #3034, opened 2026-03-03
Feat: add option to hide or show only sub-chapters

SecretX33:
I saw [this](https://github.com/mihonapp/mihon/issues/3020) issue today and realized that I also wanted this feature in Mihon, so I went ahead and gave it a shot.

## Motivation
Some sources split chapters into intermediary parts like `12.1`, `12.2`, etc., alongside a combined `Chapter 12`, effectively duplicating content. After finishing a chapter, users must either manually skip all intermediary parts or exit the reader and select the next main entry. As you can imagine, this can get extremely annoying very quickly.

This PR adds an option in Mihon to filter out these sub-chapters, allowing users to read only the main chapters and keep their library cleaner while still supporting sources that rely on split numbering.

## Notes
I also added some unit tests to validate that the manga chapter flags behave as expected and that the `Chapter.isSubChapter` value is being set correctly.

## Images
| No filters | Enabling sub-chapter only mode | List w/ sub-chapter only mode | Enabling sub-chapters removal mode | Listing w/ sub-chapter removal mode |
| ------- | ------- | ------- | ------- | ------- |
|  |  |  |   |  |

### DIFF
--- a/app/src/main/java/eu/kanade/domain/chapter/model/ChapterFilter.kt
+++ b/app/src/main/java/eu/kanade/domain/chapter/model/ChapterFilter.kt
@@ -8,7 +8,6 @@ import tachiyomi.domain.chapter.service.getChapterSort
 import tachiyomi.domain.manga.model.Manga
 import tachiyomi.domain.manga.model.applyFilter
 import tachiyomi.source.local.isLocal
-import kotlin.math.floor
 
 /**
  * Applies the view filters to the list of chapters obtained from the database.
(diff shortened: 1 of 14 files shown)


SecretX33:
Hi! I noticed there have been some reactions on this PR, but I haven’t seen any detailed feedback yet.

Since PRs usually reflect someone’s attempt to extend or adapt the project to cover a use case they need, I’d really appreciate hearing your thoughts on it. It also seems like a few other people might find this feature useful.

If you feel the feature isn’t a good fit for the project, or if there are concerns about the implementation, I’d be happy to discuss them and see whether adjustments can be made. Just let me know what you think.

Muhammedahmed18:
@SecretX33 When do you think your feature will be reviewed as I am really waiting for this feature to come?

SecretX33:
> @SecretX33 When do you think your feature will be reviewed as I am really waiting for this feature to come?

I don't know. I am not a maintainer of the project, I'm just another user that simply happened to know programming and decided to code a solution for your Issue.

aarabdh:
This is a good feature. And I would also like it if it made its way to official release. Thanks @SecretX33!

PR #3035, opened 2026-03-04
Refactor: "Pinned Only" toggle for global search

SecretX33:
## Motivation
The global search screen currently defaults to showing only pinned sources. This behavior is bad, especially for first-time users. When someone installs the app for the first time, they will not have any pinned sources, and as a result, their initial searches will always be an empty list unless they notice that there's a filter already applied that they need to remove (the "show results from pinned sources only" filter).

In general, we should assume that users don't want any filters applied unless they explicitly choose them. The current default assumes that users want to see results from pinned sources only, instead of showing results from all sources by default.

Additionally, since the state of this option is not persisted, users must manually switch to the "All" view every time they perform a search if they want results from all sources.

As a side note, the current UI uses two mutually exclusive filter chips ("Pinned Sources" and "All") for a binary choice. This is redundant and can be replaced with a more intuitive ON/OFF toggle.

## Changes
1. Replace the two filter chips with a single "Pinned Only" toggle chip that defaults to OFF (showing all sources) instead of ON (showing only pinned sources)
2. Persist the toggle state in `SourcePreferences` so the app remembers the user's choice across searched. This is aligned with the behavior of "Has results" chip
3. Automatically hides the "Pinned only" chip when there are no pinned sources. This prevents filtering the search result set into an empty result set and avoids confusing users who haven't pinned any sources

## Video

Check the video below to see a demonstration of my proposed changes.

https://github.com/user-attachments/assets/01a98c7b-6a2b-46ba-be5b-74d6b1ad87e6

_A short video showing the improvements on the search UI/UX_

## Notes
I considered not removing the `SourceFilter` enum, but I ended up not coming up with a good reason to keep this value as an enum, so I went ahead and modified it to be a simple boolean instead.

### DIFF
--- a/app/src/main/java/eu/kanade/domain/source/service/SourcePreferences.kt
+++ b/app/src/main/java/eu/kanade/domain/source/service/SourcePreferences.kt
@@ -64,6 +64,11 @@ class SourcePreferences(
         false,
     )
 
+    val globalSearchPinnedOnly: Preference<Boolean> = preferenceStore.getBoolean(
+        Preference.appStateKey("global_search_pinned_only"),
+        false,
+    )
+
     val migrationSources: Preference<List<Long>> = preferenceStore.getLongArray("migration_sources", emptyList())
 
     val migrationFlags: Preference<Set<MigrationFlag>> = preferenceStore.getObjectFromInt(
--- a/app/src/main/java/eu/kanade/presentation/browse/GlobalSearchScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/browse/GlobalSearchScreen.kt
@@ -39,6 +39,7 @@ fun GlobalSearchScreen(
                 navigateUp = navigateUp,
                 onChangeSearchQuery = onChangeSearchQuery,
                 onSearch = onSearch,
+                hasPinnedSources = state.hasPinnedSources,
                 pinnedOnly = state.pinnedOnly,
                 onTogglePinnedOnly = onTogglePinnedOnly,
                 onlyShowHasResults = state.onlyShowHasResults,
--- a/app/src/main/java/eu/kanade/presentation/browse/GlobalSearchScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/browse/GlobalSearchScreen.kt
@@ -13,7 +13,6 @@ import eu.kanade.presentation.browse.components.GlobalSearchToolbar
 import eu.kanade.tachiyomi.source.Source
 import eu.kanade.tachiyomi.ui.browse.source.globalsearch.SearchItemResult
 import eu.kanade.tachiyomi.ui.browse.source.globalsearch.SearchViewModel
-import eu.kanade.tachiyomi.ui.browse.source.globalsearch.SourceFilter
 import eu.kanade.tachiyomi.util.system.LocaleHelper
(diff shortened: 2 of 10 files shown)


BrutuZ:
So your "solution" is to have users always hammer every installed source on each search instead of guiding them to pin their most used sources?

SecretX33:
> So your "solution" is to have users always hammer every installed source on each search instead of guiding them to pin their most used sources?

Ooooh, so that's why the default option is to show pinned sources only? I didn't know that. 

To be fair, I downloaded the app yesterday, and when I did my first search I got confused on why I was seeing no results even though I did install some sources. I eventually figured out I had to switch to the "All" view to be able to see the results, and I thought that the "pinned" view being the default was a bug, especially because I had no pinned sources to begin with, which caused all my searches to yield 0 results on the default view.

From the three changes I implemented, the one I think is more important is:
> 2. Persist the toggle state in SourcePreferences so the app remembers the user's choice across searched. This is aligned with the behavior of "Has results" chip

Because the app should remember which option the user picked last.

**Given your feedback, here's my proposal:**

If the goal of making the pinned tab be the default tab on searches is to incentivize users to pin their preferred sources (to avoid hammering all sources, as you explained), the app is definitely missing an "empty state" screen on the "Pinned" tab nudging users to go and pin their favorite sources when they search - and that's something I'm up to give a shot and try implement myself.

I can make the default option still be the "Pinned tab" (now "Pinned only" filter), make the "Pinned only" button always visible (like it was before), and implement an empty state screen incentivizing the user to go and pin their favorite sources, so a new user would see this screen asking them to go and pin some sources as their first screen (and every time they enable the "Pinned only" filter if they haven't pinned anything yet)

If you liked my proposal, please let me know, otherwise let's iterate on it.

Thanks for the fast response :)

PennyJim:
The migrate page already prompts you to update your pinned sources before searching.

While I have problems with how clear that screen is about the fact that it updates your pins, I think showing that when searching with no pinned sources makes good sense.

SecretX33:
> The migrate page already prompts you to update your pinned sources before searching.

It does not?

I just checked that in the production build, and it asks you to select which sources you want the migration to search on. The behavior is consistent regardless of having pinned sources.

> While I have problems with how clear that screen is about the fact that it updates your pins, I think showing that when searching with no pinned sources makes good sense.

What screen updates your pins? The only one I saw that did this was the extension screen when clicking on the pin icon next to the extension name.

PennyJim:
> > The migrate page already prompts you to update your pinned sources before searching.
> 
> It does not?
> 
> I just checked that in the production build, and it asks you to select which sources you want the migration to search on. The behavior is consistent regardless of having pinned sources.

That's the page I was talking about.. a quick check that I should've done first shows it doesn't actually update your pins anymore. I could've sworn it used to though. I stopped maintaining my pin list because of it.

SecretX33:
I’ve rebased this branch onto the latest `main` and resolved the conflicts, so the PR should be mergeable again.

When you have a chance, I’d appreciate any feedback on whether the changes in this PR raise concerns. I’ve missed having these UI/UX improvements, and I’d like to understand whether there’s a path forward for the PR. I’d be happy to clarify the intent behind any of the three changes or discuss possible adjustments.

PR #3040, opened 2026-03-05
fix: tracker-induced duplicate key crash in duplicate detection

### LINKED ISSUE #3012: Trying to add manga that already exists in another source crashes the app if multiple trackers are set
### Steps to reproduce

1. Add more than one tracker to (or have them already in) manga that's not in library but exists in another source
2. Tap "Add to library"
3. Crash. 

### Expected behavior

It should show the migrate dialog. Or at least add it in library. 

### Actual behavior

Crash. 

### Crash logs

[mihon_crash_logs.txt](https://github.com/user-attachments/files/25620528/mihon_crash_logs.txt)

### Mihon version

0.19.4-7574 beta

### Android version

13

### Device

Lenovo Tab M9 (TB310FU) 

### Other details

Pretty sure this is related to the recent addition that looks for migrations based on trackers.
This only happens if two or more trackers are present. Removing all made it add the manga in the library directly, leaving one makes the migration dialog open as expected.
I think if there are multiple trackers pointing at the same title, migration finds that title multiple times but doesn't check that they are the *same* title. Then it tries to display all of them, but they have the same unique id, which throws the "the key is already used" error. 

### Acknowledgements

leodyversemilla07:
This addresses a crash in the 'DuplicateMangaDialog' where the Jetpack Compose LazyRow fails with an 'IllegalArgumentException: Key X was already used' error.

### Root Cause
The 'getDuplicateLibraryManga' SQL query joins 'manga_sync' (trackers) to find matching library entries. If a library manga has multiple trackers (e.g., both MAL and AniList) that match the new manga being added, the join returns one row per tracker for the SAME manga. This results in the UI receiving a list with duplicate manga IDs, which violates the requirement for unique keys in Compose's LazyRow.

### Fix
Added the 'DISTINCT' keyword to the 'track_dupes' CTE in 'mangas.sq' to ensure that each matching library manga ID is only returned once, regardless of how many trackers it shares with the target manga.

Fixes #3012

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 ### Improved
 - Rework internals for better performance ([@Lolle2000la](https://github.com/Lolle2000la)) ([#2955](https://github.com/mihonapp/mihon/pull/2955))
 
+### Fixed
+- Fix tracker-induced duplicate key crash in duplicate detection ([@leodyversemilla07](https://github.com/leodyversemilla07)) ([#3040](https://github.com/mihonapp/mihon/pull/3040))
+
 ## [v0.19.4] - 2026-02-25
 ### Added
 - Automatically remove downloads on Suwayomi after reading, configurable via extension settings ([@cpiber](https://github.com/cpiber)) ([#2673](https://github.com/mihonapp/mihon/pull/2673))


MajorTanya:
FYI because I've seen you do this several times now, but in general, you don't have to add comments to the original issue that you've fixed it with a given PR link. GitHub will automatically link the PR to the issue if you include the shorthand (`#12345`) of the issue, and it will even close it automatically if you use so-called "closing keywords" (e.g. "fix", "close", etc).

More info: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

MajorTanya:
That aside, this is a great find! I've seen this key crash several times myself and was very confused as to how it kept happening (especially since those keys are the internal manga IDs which you can find in your library with `id:123` in the search). But duplicated rows would have that effect, yeah.

leodyversemilla07:
Got it, I'll stop doing that going forward. Thanks for the heads up!

leodyversemilla07:
> That aside, this is a great find! I've seen this key crash several times myself and was very confused as to how it kept happening (especially since those keys are the internal manga IDs which you can find in your library with `id:123` in the search). But duplicated rows would have that effect, yeah.

Appreciate it! Honestly wasn't sure it was the right call at first, but the duplicated rows from the join made sense once I saw it.

AntsyLich:
add an entry to changelog file

PR #3041, opened 2026-03-06
fix: suppress JS dialogs from WebView after screen is closed

### LINKED ISSUE #2900: Webview window.alert being displayed when webview is closed
### Steps to reproduce

I have to mention the specific source due to the nature of the issue.

1. Select a title that uses the MangaPark extension
2. Open webview to a black screen
3. Wait 10-15 seconds, get redirected, and then close webview
4. Go back to the mihon library, wait ~20 seconds and then receive a box pop-up

Image

https://github.com/user-attachments/assets/fdd68470-154a-48bf-94ed-03f966f96775

The nature of the pop-up message makes me worry about some sort of malware being installed, so if you're able provide any solutions to that or assuage my worries in some way that would be much appreciated. I'm not technical, so apologies if any of the details are off.

### Expected behavior

It would be expected that when you close webview that instance and anything done in it also closes.

### Actual behavior

Instead, you receive a pop-up related to that webview instance that displays over Mihon.

### Crash logs

_No response_

### Mihon version

0.19.3

### Android version

Android 13

### Device

Google Pixel 4a

### Other details

_No response_

### Acknowledgements

leodyversemilla07:
## Summary

Fixes #2900

## Root Cause

The `webChromeClient` in `WebViewScreenContent.kt` never overrode `onJsAlert`, `onJsConfirm`, or `onJsPrompt`. Android's default `WebChromeClient` handles these by showing native system dialogs, with no awareness of whether the WebView is still open. Since JS executes asynchronously, `window.alert()` can fire after the user has already closed the WebView, causing a native dialog to appear over the main app UI.

## Fix

Added an `isActive` flag tracked via `DisposableEffect` that flips to `false` the moment the composable leaves the screen. Overrode `onJsAlert`, `onJsConfirm`, and `onJsPrompt` in the `webChromeClient` to silently suppress dialogs when the WebView is no longer active.

```kotlin
val isActive = remember { mutableStateOf(true) }

DisposableEffect(Unit) {
    onDispose { isActive.value = false }
}
```

```kotlin
override fun onJsAlert(view: WebView, url: String?, message: String?, result: JsResult): Boolean {
    if (!isActive.value) {
        result.confirm()
        return true
    }
    return super.onJsAlert(view, url, message, result)
}
```

- If the WebView is **still open** → dialogs show as normal
- If the WebView is **already closed** → dialogs are silently dismissed

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/webview/WebViewScreenContent.kt
+++ b/app/src/main/java/eu/kanade/presentation/webview/WebViewScreenContent.kt
@@ -188,7 +188,13 @@ fun WebViewScreenContent(
                 return super.onJsConfirm(view, url, message, result)
             }
 
-            override fun onJsPrompt(view: WebView, url: String?, message: String?, defaultValue: String?, result: JsPromptResult): Boolean {
+            override fun onJsPrompt(
+                view: WebView,
+                url: String?,
+                message: String?,
+                defaultValue: String?,
+                result: JsPromptResult,
+            ): Boolean {
                 if (!isActive) {
                     result.cancel()
                     return true


app/src/main/java/eu/kanade/presentation/webview/WebViewScreenContent.kt
```diff
@@ -95,6 +98,11 @@ fun WebViewScreenContent(
 
     var currentUrl by remember { mutableStateOf(url) }
     var showCloudflareHelp by remember { mutableStateOf(false) }
+    val isActive = remember { mutableStateOf(true) }
+
+    DisposableEffect(Unit) {
+        onDispose { isActive.value = false }
+    }
```
AntsyLich: This'll require some imports. Also update in other places
```suggestion
    var isActive by remember { mutableStateOf(true) }

    DisposableEffect(Unit) {
        onDispose { isActive = false }
    }
```

app/src/main/java/eu/kanade/presentation/webview/WebViewScreenContent.kt
```diff
@@ -335,3 +367,4 @@ fun WebViewScreenContent(
         }
     }
 }
+
```
AntsyLich:
```suggestion
```

CHANGELOG.md
```diff
@@ -501,3 +504,4 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 [v0.16.2]: https://github.com/mihonapp/mihon/compare/v0.16.1...v0.16.2
 [v0.16.1]: https://github.com/mihonapp/mihon/compare/v0.16.0...v0.16.1
 [v0.16.0]: https://github.com/mihonapp/mihon/compare/a9c7cbf...v0.16.0
+
```
```suggestion
```

PR #3042, opened 2026-03-07
add page read on history screen

CompileConnected:
add page read on history screen

Screenshot 2026-03-07 at 13 19 50


CompileConnected:
But i kinda need a feedback on the UI.

So when already completed read, the ui become dimmed.
Image

What i create rather simple, you already read until pages 12, which is the last page on that chapter.
Image

For me it wont make any sense to make the text dim like update screen or entry list.
is this enough?

PR #1386, opened 2024-10-26
Add Hikka Tracker Integration

Lorg0n:
Add Hikka tracker integration
Closes https://github.com/mihonapp/mihon/issues/1145

Add support for Hikka (https://hikka.io/), a Ukrainian online encyclopedia for manga and etc.

Changes implemented:
- Added new Hikka tracker with ID 10
- Implemented OAuth authentication flow with auto token refresh
- Added full sync support for:
  - Reading status (Reading, Completed, On Hold, etc.)
  - Chapter progress
  - Scores (0-10 scale)
  - Manga search and linking
- Created necessary data models and API integration
- Added UI elements for tracker settings

All changes have been tested:
- [x] Authentication flow
- [x] Manga search and linking
- [x] Status/progress syncing
- [x] Error handling

### Images
| Settings Screen | Manga Track Dialog |
| ------- | ------- |
| image | image |

PR #3047, opened 2026-03-08
Add support for MangaBaka

### LINKED ISSUE #2610: MangaBaka tracker

https://mangabaka.dev/
It has everything from other trackers and it also references those other trackers in the entries. Maybe it could also be used to "automatically" bind other logged in trackers into entries?
For example, the user tracks an entry using mangabaka and it will either ask if the user wants to also bind the other logged in trackers by using the urls provided by mangabaka automatically or maybe it could make the only result on the manga search be the url provided by mangabaka (sorry, I'm bad at explaining)
it also does have an API

### Other details

_No response_

### Acknowledgements

MajorTanya:
Closes #2610.

Added support for `id:` prefix searching for when the title may be too ambiguous to find in the 10 results we request. Works like all other prefix searches: 

`id:26575` will return 1 result (https://mangabaka.org/26575, Youjo Senki Restaurant)
If the ID is not recognised by Mangabaka, the results screen will display the "no results found" message as with a text search.

---

Current things to note:
- [x] Hikka (#1386) currently uses the same tracker ID (10). I could simply change this implementation to use the ID 11 and we'd avoid a conflict there. (~~other than me having to redo my local use of it but that's whatever 👍~~)
  Update: Changed ID to 11.
- [x] Note: missing a client ID which should be added by the project owner. I've tried this out with my own OAuth2 client and it _appears_ to be working fine (adding, deleting, updating, etc).
  Update: client ID added in e2044a3.
- [ ] MangaBaka does not currently (as of 2026-03-29) offer a data point of a given title's last chapter number (which Mihon would use to determine when to automatically Complete an entry). See https://github.com/mihonapp/mihon/pull/3047#issuecomment-4045438119 for further info.


app/src/main/java/eu/kanade/tachiyomi/data/track/mangabaka/MangaBakaApi.kt
```diff
@@ -0,0 +1,258 @@
...
+
+            with(json) {
+                client.newCall(POST("${OAUTH_URL}/token", body = formBody))
+                    .awaitSuccess().parseAs()
+            }
+        }
+    }
+
+    companion object {
+        private const val CLIENT_ID = ""
```
MajorTanya: @AntsyLich set here thanks

MajorTanya:
I discovered that MangaBaka uses the `total_chapters` field to represent an _ongoing tally of chapters_ instead of a definite and final count of chapters, meaning it is a number that both 1) changes regularly and 2) does not represent the point in time where a series can be finished by reading said numbered chapter.

I'm currently in discussion with their developers to see if there is an alternative way to solve this so Mihon can display x/y progress properly as well as auto-set Completed once users read the final chapter. But until we find such an alternative, this PR is **blocked** and **cannot be merged**.

If no such alternative is found, I'll have to remove all uses of/references to `total_chapters` and MangaBaka in Mihon would just show the current progress without a total.

MajorTanya:
Added support for different score step sizes as per MangaBaka's new feature. Currently has to be fetched from the OAuth userinfo endpoint but I have created a ticket to see if a new endpoint can be added that is under the normal API url.

MajorTanya:
> I discovered that MangaBaka uses the `total_chapters` field to represent an _ongoing tally of chapters_ instead of a definite and final count of chapters, meaning it is a number that both 1) changes regularly and 2) does not represent the point in time where a series can be finished by reading said numbered chapter.
> 
> I'm currently in discussion with their developers to see if there is an alternative way to solve this so Mihon can display x/y progress properly as well as auto-set Completed once users read the final chapter. But until we find such an alternative, this PR is **blocked** and **cannot be merged**.
> 
> If no such alternative is found, I'll have to remove all uses of/references to `total_chapters` and MangaBaka in Mihon would just show the current progress without a total.

For the record, I have removed the total_chapters data fetching in 0f96d02be033a2f61ab380e1a382989447232a47 (for now). We can think about adding it back if/when MangaBaka make their decision on this.

BrutuZ:
Can't you apply the total chapters conditionally on the publishing status?

MajorTanya:
> Can't you apply the total chapters conditionally on the publishing status?

I'm not going to do that for "total" chapters, no.

MajorTanya:
> Added support for different score step sizes as per MangaBaka's new feature. Currently has to be fetched from the OAuth userinfo endpoint but I have created a ticket to see if a new endpoint can be added that is under the normal API url.

MangaBaka developers have added the `/v1/my/profile` endpoint with this information, using this now since 9ba7d95.

MajorTanya:
I've decided that out of respect for the Hikka PR, I'll make MangaBaka tracker with the ID 11.

I'll have to do some finagling of my personal local use but that's just something for me to figure out. Hopefully nobody else was stupid enough to try and early bird this 😅

MajorTanya:
Mihon now identifies itself on all requests properly in the format
~~`Mihon/v[version name] (Android) (https://github.com/mihonapp/mihon)`~~
`[app name]/v[version name] ([app id] [commit sha]) (Android) (https://github.com/mihonapp/mihon)`

For example, the current stable would look like
~~`Mihon/v0.19.7 (Android) (https://github.com/mihonapp/mihon)`~~
`Mihon/v0.19.7 (app.mihon 54d2c7b9e) (Android) (https://github.com/mihonapp/mihon)`
and current preview would be
~~`Mihon/v0.19.7-7630 (Android) (https://github.com/mihonapp/mihon)`~~
`Mihon/v0.19.7-7630 (app.mihon.debug 85eb539c4) (Android) (https://github.com/mihonapp/mihon)`

MajorTanya:
Mihon now uses the freshly published `published.startDate` attribute on a series for search results.

MajorTanya:
I'll undraft this now but note that it does **not** have support for showing total chapters in the tracker sheet (as in, it can't show "13/200" or whatever) due to aforementioned issues with MB's handling of total chapter data.

MajorTanya:
~~Actually I can't undraft this because it's still missing Mihon's OAuth client ID 😅~~

Handled with e2044a3.

app/src/main/java/eu/kanade/tachiyomi/data/track/mangabaka/MangaBakaApi.kt
MajorTanya: Handled with e2044a3

MajorTanya:
Added a `state` parameter to the `/authorize` OAuth request. This is seemingly required by MangaBaka's auth library (patch version change btw) which we discovered due to reports of users in all sorts of clients failing to request an access token from the API even though the initial authorization code was successfully requested.

I'm very unsure about the error handling of a missing/wrong state response, please take a closer look and provide feedback. Currently, it logs a Warning to logcat and silently stops the MB login procedure, which seems suboptimal to me.

MajorTanya:
Drafted again because it uses the deprecated `title` field as opposed to the new `titles` array. I can't figure out how to decide what title to show from the array. Until such time that I can, draft.

Yes I know about https://mangabaka.org/pages/announcements/15-titles-v2#finding-the-title-you-want

MajorTanya:
Migrated to Titles v2 with the example algorithm mentioned on Mangabaka's announcement of Titles v2 (as custom ordering by users is not exposed in the API at this time), but including `zh-Latn` and `zh` as well.
Falls back to whatever the first title in the array is if none of the priority-ordered languages match.

Currently, I'm including an ultimate fallback in case `titles` ends up as an empty array due to some circumstance since I couldn't determine whether it's guaranteed to have at least one title (as opposed to the one-per-lang guarantee of `is_primary` MB mentions in the announcement).
I've contacted the developers about this and might remove the ultimate fallback again if it's not required.

BrutuZ:
Aww, didn't go the extra mile of prioritizing the media format language first (`jp` for manga, `ko` for manhwa, `zh` for manhua) :(

MajorTanya:
> Migrated to Titles v2 with the example algorithm mentioned on Mangabaka's announcement of Titles v2 (as custom ordering by users is not exposed in the API at this time), but including `zh-Latn` and `zh` as well. Falls back to whatever the first title in the array is if none of the priority-ordered languages match.
> 
> Currently, I'm including an ultimate fallback in case `titles` ends up as an empty array due to some circumstance since I couldn't determine whether it's guaranteed to have at least one title (as opposed to the one-per-lang guarantee of `is_primary` MB mentions in the announcement). I've contacted the developers about this and might remove the ultimate fallback again if it's not required.

After some talking, I've come up with the current format:
> ID: 438410 - Could not find name! (report on the Mangabaka Discord)

That ID happens to have a `titles: null` and as such will fall back on this message:
image

---

With that, the only remaining thing I can think of that _could_ prevent merging would be the lack of a definitive data point for what constitutes the last chapter which Mihon could use for the "12/34 chapters" display & for automatic "Completed" marking. In my eyes, the lack of such a data point is _not_ a blocking issue though.

MajorTanya:
Also rebased onto `main` for simplicity, though there weren't any conflicts.

MajorTanya:
> With that, the only remaining thing I can think of that could prevent merging would be the lack of a definitive data point for what constitutes the last chapter which Mihon could use for the "12/34 chapters" display & for automatic "Completed" marking. In my eyes, the lack of such a data point is not a blocking issue though.

@AntsyLich up to you to decide whether this is a blocking issue or not. I'll undraft based on your call.

app/src/main/java/eu/kanade/tachiyomi/data/track/mangabaka/dto/MangaBakaItem.kt
```diff
@@ -0,0 +1,76 @@
...
+                        "native" in it.traits -> 2
+                        else -> 3
+                    }
+                }
+        }
+
+        return TITLE_PRIORITIES
+            .firstNotNullOfOrNull { bestTitlePerLanguage[it]?.title }
+            ?: titles?.firstOrNull()?.title
+            ?: "ID: $id - Could not find name! (report on the MangaBaka Discord)"
```
BrutuZ, line 49: Should this be a i18n string?

MajorTanya: It _could_ be but at the same time, this is such a deep fallback value that it feels overkill to localise it.

_Technically_, this isn't even supposed to happen. It's just there because it _can_ happen sometimes when MB doesn't have a title for something in any language at all. When I asked in the Discord, I believe only 8 DB entries existed with this, in total.

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 ### Added
 - Add dedicated "Support Us" screen ([@AntsyLich](https://github.com/AntsyLich)) ([#3200](https://github.com/mihonapp/mihon/pull/3200))
 - Add a one time popup asking for donation from long time users ([@AntsyLich](https://github.com/AntsyLich)) ([#3203](https://github.com/mihonapp/mihon/pull/3203))
+- Add support for MangaBaka tracker ([@MajorTanya](https://github.com/MajorTanya)) ([#3047](https://github.com/mihonapp/mihon/pull/3047))
```
AntsyLich: wrong place. put it under hikka one

app/src/main/res/drawable-nodpi/brand_mangabaka.webp
AntsyLich, line 1: rename brand_mangabaka.webp

AntsyLich (changes requested):

MajorTanya:
- Display name support from my PR #3533 
- Fixed changelog
- Renamed icon file

AntsyLich (approved):

PR #3055, opened 2026-03-12
Fix manga details for non-library manga taking longer to load

### LINKED ISSUE #3027: The manga page accessed from the search within an extension became slower, and the overall handling of these items became slower.
### Steps to reproduce

Step 1: Open an extension. 
Step 2: Open any manga. 
Step 3: Open the filter options and click on the manga and check the time it takes for the action to occur. 
Step 4: Open any chapter so that the manga can be found in the history. 
Step 5: Return to the app's home screen and access the previously edited manga from your history.  
Step 6: Repeat steps 3 and 4 and verify that the execution is faster simply by accessing the manga in the history instead of accessing it from the extension in the browse tab. 

### Expected behavior

It was expected that when interacting with a manga accessed through an extension, this interaction would not be slower than in other parts of the system.  The difference between accessing the manga through the history or the item in the library, and accessing that same item in an entry in the navigation tab through an extension, ends up being transparent. 

### Actual behavior

The response time on the manga page accessed from an extension in the navigation tab is lower than the time achieved by version 0.19.3 of the application. This doesn't just apply to filter manipulation; as I mentioned in a previous issue, the reader experienced slower performance on manga accessed through extension searches. Accessing the same manga from the library or history ensures the reader experiences the same performance as the previous version, without any issues like... The page counter index does not track pagination until the page is loaded, and the loading process does not occur correctly. 

### Crash logs

_No response_

### Mihon version

0.19.4

### Android version

Android 15

### Device

Moto G 15

### Other details

This also serves as a complement to the first issue I sent yesterday; as I used the new version, the problem became clearer. The performance problem lies in how the application is handling extensions; the manipulation of manga files accessed by the manga search within extensions has become slower, en Accessing the same manga by opening its entry in the history or library works perfectly, with no issues of slow filtering or slow image loading in the reader. 

### Acknowledgements

FlaminSarge:
Fixes #3027 
Regression from #2955 (probably)

Moves setDefaultChapterFlags after the initial load so at least the Manga details page shows up while fetches are still happening. Also avoids setting default chapter flags if default is the same as the manga's current flags (for no defaults set or if this is reopening a non-favorite).

The problem seemed to be that the `mangas` DB write from setMangaDefaultChapterFlags had to wait for a long time (150+ms on my emulator with a 6000+ library). Not sure if that's due to the mutex changes for DB transactions or something else, but on top of that it _also_ increased the time that availableScanlators and excludedScanlators took; also unsure why that is.

Before these changes, timing was something like this:
```
  D  MangaScreen [<title>] awaitManga: 0ms
  D  MangaScreen [<title>] awaitChapters: 1ms
  D  MangaScreen [<title>] setDefaultFlags: 179ms
  D  MangaScreen [<title>] availableScanlators: 296ms
  D  MangaScreen [<title>] excludedScanlators: 379ms
  D  MangaScreen [<title>] State.Success: 381ms
```

After the changes:
```
  D  MangaScreen [<title>] awaitManga: 0ms
  D  MangaScreen [<title>] awaitChapters: 1ms
  D  MangaScreen [<title>] availableScanlators: 2ms
  D  MangaScreen [<title>] excludedScanlators: 2ms
  D  MangaScreen [<title>] State.Success: 2ms
  D  MangaScreen [<title>] setDefaultFlags: 422ms
```

And if setting the default flags gets to no-op because the default chapter flags are 0 or the same as before, that takes barely any time as well.

availableScanlators/excludedScanlators await's don't _have_ to be moved out of the state update block, but I figured it doesn't hurt and can prevent pain later if those also need to be debugged again (moved them around to get timings for those initially, left them there afterwards).


FlaminSarge:
Note that this should still probably go in despite #2982 being a good rework of the root cause, since currently any time a non-favorite is opened it triggers a `mangas` db write when it doesn't need to.

domain/src/main/java/tachiyomi/domain/manga/interactor/SetMangaChapterFlags.kt
```diff
@@ -71,22 +71,29 @@ class SetMangaChapterFlags(
...
+            .setFlag(downloadedFilter, Manga.CHAPTER_DOWNLOADED_MASK)
+            .setFlag(bookmarkedFilter, Manga.CHAPTER_BOOKMARKED_MASK)
+            .setFlag(sortingMode, Manga.CHAPTER_SORTING_MASK)
+            .setFlag(sortingDirection, Manga.CHAPTER_SORT_DIR_MASK)
+            .setFlag(displayMode, Manga.CHAPTER_DISPLAY_MASK)
+        if (currentFlags == newFlags) {
+            // avoid writing to the mangas DB if unnecessary
+            // debatable whether no-op should be counted as success
+            return true
+        }
```
AntsyLich, line 92: Move the comment above the statement

PR #3091, opened 2026-03-19
Replace preference getter functions with properties

AntsyLich:


MajorTanya:
Why?

AntsyLich:
> Why?

They return a stable value. No need to initialize them over and over.

MajorTanya:
But ones like the Updates filter prefs may change many times, depending on which filters the user is toggling etc. If they're vals, doesn't that mean they won't ever be checked again?

sudo-ezekiel:
@MajorTanya the val is just storing a reference to the preference, not the value itself. When you call .get() or .changes() it still reads the current value, so filters toggling/we should work the same as before. Plus like Antsy said, no point creating the same object over and over when it never changes.

MajorTanya:
> @MajorTanya the val is just storing a reference to the preference, not the value itself. When you call .get() or .changes() it still reads the current value, so filters toggling/we should work the same as before. Plus like Antsy said, no point creating the same object over and over when it never changes.

I see. I had never looked into how they're actually called and I was basically ignoring the get/changes calls on the preferences each time, thinking that it was just some quirk of what the system demands. If the Preference object itself is just an accessor object for the value stored within then yeah, I agree that these changes make sense.

PR #3096, opened 2026-03-20
fix: invalidate download cache after backup restore

### LINKED ISSUE #2574: Invalidate downloads cache when restoring backup
### Steps to reproduce

Take an existing library with some downloads. Create a backup. Then, uninstall and reinstall Mihon, select the same app folder, restore the backup.

### Expected behavior

Already-downloaded chapters should still be displayed as downloaded.

### Actual behavior

All chapters are displayed as not downloaded, even if they have been. Trying to download an already-downloaded chapter causes an error.

The user has to navigate to advanced settings and use the "invalidate download cache" option. After they do this, the correct state is displayed.

As far as I can tell, the correct behavior is straightforwardly that the download cache needs to be invalidated when the user restores a backup. It's clearly always going to be wrong to trust a newly-created empty downloads cache when there are in fact already downloads.

### Crash logs

_No response_

### Mihon version

58b25d697f7987e9888344e815d5646ec010a663

### Android version

Android 16

### Device

Google Pixel 6a

### Other details

This is the same problem reported in https://github.com/mihonapp/mihon/issues/13, which was closed without the underlying issue being fixed.

### Acknowledgements

leodyversemilla07:
## Summary

Fixes #2574

Automatically invalidate the download cache after backup restore to ensure the UI correctly reflects which chapters are actually downloaded.

## Root Cause

When a backup is restored, the download cache is not rebuilt. This causes:
- Downloaded chapters showing as 'not downloaded' in the UI
- Errors when attempting to download already-downloaded chapters
- Users having to manually invalidate the cache via Settings → Advanced

The download cache maintains an in-memory mapping of downloaded chapters. After backup restore, the cache remains empty even though downloaded files exist in storage, causing it to be out of sync with reality.

## Changes

- Added automatic download cache invalidation after backup restore completes
- Only invalidates when library entries are restored (no point if no manga data was restored)
- Wrapped in try-catch to prevent restore failure if cache invalidation encounters an error
- Follows the same pattern used for storage location changes

## Testing

- Self-reviewed the change
- Follows existing patterns: cache invalidation already used in \SettingsAdvancedScreen.kt\ and on storage changes
- Low risk: just calling existing, well-tested \invalidateCache()\ method

## Related

This is similar to how the cache is automatically invalidated when storage location changes (see \DownloadCache.kt:131\)


AntsyLich (approved):
Add a changelog entry

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 ### Fixed
 - Add missing `outlineVariant` color to Nord theme ([@CompileConnected](https://github.com/CompileConnected)) ([#3184](https://github.com/mihonapp/mihon/pull/3184))
 - Continue reading button missing when unread filter is off ([@AntsyLich](https://github.com/AntsyLich)) ([#3382](https://github.com/mihonapp/mihon/pull/3382))
+- Invalidate download cache after backup restore ([@leodyversemilla07](https://github.com/leodyversemilla07)) ([#3096](https://github.com/mihonapp/mihon/pull/3096))
```
AntsyLich: move under changed

leodyversemilla07: moved under changed. thanks!

PR #3118, opened 2026-03-22
Address Gradle warning about `Task.project`

MajorTanya:
Apparently, this is all that's needed to replace the forbidden `Task.project` accessor, which would be an error in Gradle v10 (it's been deprecated for a while)

This will also allow us to use the Configuration Cache if we wanted to (which we should since that'll be mandatory for v10 as well and it seems to be quite speedy from some cursory testing on my machine).

I checked, and the output XML contents generated by the previous code and this new code is identical.

Gradle docs about this:
https://docs.gradle.org/9.4.1/userguide/upgrading_version_7.html#task_project
https://docs.gradle.org/9.4.1/userguide/configuration_cache_requirements.html#config_cache:requirements:use_project_during_execution

### DIFF
--- a/gradle/build-logic/src/main/kotlin/mihon/gradle/tasks/GenerateLocalesConfigTask.kt
+++ b/gradle/build-logic/src/main/kotlin/mihon/gradle/tasks/GenerateLocalesConfigTask.kt
@@ -2,19 +2,24 @@ package mihon.gradle.tasks
 
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.DirectoryProperty
+import org.gradle.api.model.ObjectFactory
 import org.gradle.api.tasks.OutputDirectory
 import org.gradle.api.tasks.TaskAction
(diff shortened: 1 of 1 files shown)


MajorTanya:
Regarding the configuration cache thing:

Some basic comparisons running `gradle assembleDebug`:
| configuration-cache setting | run time (cold) | run time (warm) |
|-|-|-|
| false | ~1 minute [^1] | ~15 seconds |
| true | ~1 minute [^1] | ~ 1-3 seconds |

For `gradle clean` it gets funnier because it can go below 1 second which I've never seen before with any Gradle task.

[^1]: Can be (much) more, depending on other caches not looked at here. Also, dependency downloads will obviously slow it all down massively.

So the question is whether there are any reasons for us to _not_ enable it, really. I'll leave it to you, @AntsyLich, but I can make a PR for the setting when/if this one is merged 👍.

PR #3127, opened 2026-03-23
Enable Gradle Configuration Cache

MajorTanya:
Gradle did a magic: https://docs.gradle.org/9.4.1/userguide/configuration_cache.html

> Gradle’s Build Cache has long optimized the execution phase by reusing previously built outputs and parallelizing as much work as possible.
>
> The Configuration Cache builds on this idea of work avoidance and parallelization. When enabled, the Configuration Cache allows Gradle to skip the configuration phase entirely if nothing that affects the build configuration (such as build scripts) has changed. Additionally, Gradle applies performance optimizations to task execution.

In the docs, they also note the following:

> This feature is not enabled by default and has the following limitations:
>
> It does not yet support all [Core Gradle Plugins](https://docs.gradle.org/9.4.1/userguide/configuration_cache_status.html#config_cache:plugins:core) and [features](https://docs.gradle.org/9.4.1/userguide/configuration_cache_status.html#config_cache:not_yet_implemented). Full support is in progress.
>
> Your build and its plugins may require modifications to meet the [Configuration Cache requirements](https://docs.gradle.org/9.4.1/userguide/configuration_cache_requirements.html#config_cache:requirements).
>
> [IDE imports and syncs](https://docs.gradle.org/9.4.1/userguide/configuration_cache.html#config_cache:ide) do not yet use the Configuration Cache.
>
> Continuous Integration (CI) support is actively evolving.
>
> Since Gradle 9.0.0 the Configuration Cache is the [preferred mode of execution](https://blog.gradle.org/road-to-configuration-cache#preferred-mode-of-execution).

PR #3140, opened 2026-03-25
chore(gradle): Create Gradle task for git commit info that is gradle config cache friendly

ArthurKun21:
Since mihon is going to enable config cache in #3127, this is a workaround for gradle and git so that it won't caused trouble when creating new git commit and thus invalidating the gradle config cache.

There was a changed too with `versionNameSuffix` being currently removed for `debug` build due to still trying to find a workaround for that.


app/build.gradle.kts
```diff
@@ -41,16 +36,13 @@ android {
     buildTypes {
         val debug by getting {
             applicationIdSuffix = ".dev"
-            versionNameSuffix = "-${getLatestCommitCount()}"
```
ArthurKun21, line 44: This was removed, I still don't know the workaround

AntsyLich: This is an issue then 🤔

ArthurKun21: Going to check it again later

ArthurKun21: image

AntsyLich:
Just why gradle
image

app/build.gradle.kts
MajorTanya: That's not the format of debug releases

ArthurKun21: let me check again a bit later

MajorTanya: We use the commit _count_ as the suffix, not the last commit sha

ArthurKun21: oof, that was actually my previous config in my repo, was using sha my bad

ArthurKun21:
oh wait btw, the gradle config cache with git was only effective on a single branch. If you change branch then run a gradle command, it invalidates cache due to `.git/HEAD` was changed iirc but the succeeding gradle commands on the same branch would then be config cached

MajorTanya:
> oh wait btw, the gradle config cache with git was only effective on a single branch. If you change branch then run a gradle command, it invalidates cache due to `.git/HEAD` was changed iirc but the succeeding gradle commands on the same branch would then be config cached

I think that's a fair compromise. Usually when you dev locally, you stick to one branch for a while so you still get the benefits there. It should be reasonably rare for someone to do builds while repeatedly swapping branches.

AntsyLich:
Thanks for the PR but I ended up going with a much simpler approach in 12fbb13df5da7031575f60e151905da4a512bbd8

PR #3148, opened 2026-03-25
Reapply "Update android.gradle to v9 (major) (#3105)"

AntsyLich:
This reverts commit eb5b76251576559fca25e30f96ac87bdb3e9d6ad.


MajorTanya:
Can confirm this time it doesn't crash loop like it did when we tried this last time.

MajorTanya (approved):

PR #3159, opened 2026-03-27
chore(gradle): add gradle properties to build-logic

ArthurKun21:
https://github.com/skydoves/landscapist/blob/main/build-logic/gradle.properties

Copy the gradle properties to the included builds

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -4,7 +4,7 @@ android.useAndroidX=true
 kotlin.code.style=official
 kotlin.mpp.androidSourceSetLayoutVersion=2
 
-# Configuration should be synced with gradle/build-logic/gradle.properties
+# Configuration should be synced with [/gradle/build-logic/gradle.properties]
 org.gradle.caching=true
 org.gradle.configureondemand=true
 org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8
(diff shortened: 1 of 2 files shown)


AntsyLich:
Can you add a comment in the root `gradle.properties` to keep it synced with this file? Something like https://github.com/mihonapp/mihon/blob/85eb539c4d3774740790b5ab9e3b124cce00ce99/gradle/build-logic/build.gradle.kts#L8

gradle/build-logic/gradle.properties
```diff
@@ -0,0 +1,6 @@
+# Gradle properties are not passed to included builds
```
AntsyLich, line 3:
```suggestion
# Configuration should be synced with [/gradle.properties]
#
# Gradle properties are not passed to included builds 
```

gradle.properties
```diff
@@ -4,6 +4,7 @@ android.useAndroidX=true
 kotlin.code.style=official
 kotlin.mpp.androidSourceSetLayoutVersion=2
 
+# Configuration should be synced with gradle/build-logic/gradle.properties
```
AntsyLich:
```suggestion
# Configuration should be synced with [/gradle/build-logic/gradle.properties]
```

MajorTanya: I assume this makes it clickable or something?

AntsyLich: It doesn't. I just like the format

PR #3163, opened 2026-03-29
Include read history when migrating

### LINKED ISSUE #3005: Option to migrate (or duplicate) chapter date time read history

When adding a tracking it should auto fill the start date to when you've first started reading the manga. Although if you've only started tracking AFTER the migration. No meta data exists for the current manga in the current source, leading to a no autofill start date. 

### Other details

This is especially useful for nuked sources where you don't have time to track before a migrate or you've simply forgotten to track the manga in the past. 

### Acknowledgements

BrutuZ:
If `Chapters` is checked when migrating entries from library, also migrate the read history, matched by chapter number

Closes #3005

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 
 ### Improved
 - Show informative error when trying to add unapproved titles to list on MAL ([@MajorTanya](https://github.com/MajorTanya)) ([#3155](https://github.com/mihonapp/mihon/pull/3155))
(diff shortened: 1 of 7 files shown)


aaron01-dev:
would it be better to overwrite if the old/previous chapter has a read history that's older, that way it always keeps track of the oldest.

or a simple confirmation to overwrite, or don't
or an advanced confirmation to overwrite, keep oldest read date, don't overwrite.

Though I guess the workaround is to just mark the target chapter as unread, though majority won't understand since there's no read date preview

BrutuZ:
A wild rebase appears

PR #3167, opened 2026-03-29
Resumable image downloads

### LINKED ISSUE #3054: Add HTTP Range Header Support for Resumable Image Downloads

Currently, when downloading chapter pages, if the download of an image is interrupted (for example due to network issues or the app being closed), the downloader deletes the partial `.tmp` file and restarts the download from the beginning.

It would be beneficial if Mihon supported HTTP Range requests to resume partially downloaded images instead of restarting them.

### Other details

- When downloading a page image, if a `.tmp` file already exists:
  - Check the current file size.
  - Send a request with a Range header:
    ```Range: bytes=<existing_file_size>-```

- If the server responds with:
  - `206 Partial Content` → append the remaining bytes to the existing file and continue the download.
  - `200 OK` → assume the server does not support Range requests, delete the partial file, and restart the download normally.

This would allow the downloader to resume interrupted downloads, reducing wasted bandwidth and improving reliability for large images or unstable connections.

### Acknowledgements

xMohnad:
Closes #3054


app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -487,15 +480,25 @@ class Downloader(
...
+                .addQueryParameter("existingSize", file.length().toString())
+                .build().toString()
+
             val response = source.getImage(page)
-            val file = tmpDir.createFile("$filename.tmp")!!
+
             try {
-                response.body.source().saveTo(file.openOutputStream())
+                // append if partial (HTTP 206), else overwrite
+                response.body.source().saveTo(file.openOutputStream(response.code == 206))
```
MajorTanya: What? What is that Boolean doing? Also, does this properly close the resource?

xMohnad: see

https://github.com/seven332/UniFile/blob/master/library/src/main/java/com/hippo/unifile/UniFile.java#L440-L448

https://github.com/mihonapp/mihon/blob/main/app/src/main/java/eu/kanade/tachiyomi/util/storage/OkioExtensions.kt#L33-L40

MajorTanya: For context, our used UniFile is here: https://github.com/tachiyomiorg/UniFile/blob/a9de196cc7ec623d5641fbe0f768c082736ac84c/library/src/main/java/com/hippo/unifile/UniFile.java#L434-L442

Could you put the argument name in the call so it's intuitive for future readers? And maybe it's worth spacing the line a bit since it's so dense:

```suggestion
                // append if partial (HTTP 206), else overwrite
                response.body
                    .source()
                    .saveTo(
                        file.openOutputStream(append = response.code == 206)
                    )
```

I don't remember ad hoc whether it's 2 or 4 space indents though but spotless should figure that out on its own either way.

MajorTanya:
Why the repeated force pushes? Just commit on top, it's getting squashed if Antsy merges it.

MajorTanya:
Re: 1d16205: Interesting, I never realised you couldn't use named arguments with non-Kotlin functions 🤔 apologies for the extra work there

xMohnad:
Got it, no problem

core/common/src/main/kotlin/eu/kanade/tachiyomi/network/OkHttpExtensions.kt
```diff
@@ -124,9 +125,17 @@ fun OkHttpClient.newCachelessCallWithProgress(request: Request, listener: Progre
     val progressClient = newBuilder()
         .cache(null)
         .addNetworkInterceptor { chain ->
-            val originalResponse = chain.proceed(chain.request())
+            var req = chain.request()
+            val existingSize = req.url.queryParameter("existingSize")?.toLong() ?: 0L
```
AntsyLich: Utilize tags (see https://square.github.io/okhttp/5.x/okhttp/okhttp3/-request/-builder/tag.html and https://square.github.io/okhttp/5.x/okhttp/okhttp3/-request/tag.html)

AntsyLich (changes requested):

core/common/src/main/kotlin/eu/kanade/tachiyomi/network/OkHttpExtensions.kt
```diff
@@ -118,13 +118,19 @@ suspend fun Call.awaitSuccess(): Response {
...
         .cache(null)
         .addNetworkInterceptor { chain ->
-            val originalResponse = chain.proceed(chain.request())
+            var req = chain.request()
+
+            req = req.newBuilder()
+                .apply { if (existingSize > 0 && req.header("Range") == null) header("Range", "bytes=$existingSize-") }
+                .build()
+
+            val originalResponse = chain.proceed(req)
```
AntsyLich:
```suggestion
            val request = chain.request()
                .newBuilder()
                .apply {
                    if (existingSize > 0 && req.header("Range") == null) {
                        header("Range", "bytes=$existingSize-")
                    }
                }
                .build()

            val originalResponse = chain.proceed(request)
```

CHANGELOG.md
```diff
@@ -54,6 +54,9 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 ### Removed
 - Remove the empty "Updated to X.Y.Z" popup after app update ([@AntsyLich](https://github.com/AntsyLich)) ([#3203](https://github.com/mihonapp/mihon/pull/3203))
 
+### Added
+- Resumable image downloads support ([@xMohnad](https://github.com/xMohnad)) ([#3167](https://github.com/mihonapp/mihon/pull/3167))
```
AntsyLich: Move this to the appropriate place at the top
```suggestion
- Resumable image downloads if supported by server ([@xMohnad](https://github.com/xMohnad)) ([#3167](https://github.com/mihonapp/mihon/pull/3167))
```

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
```diff
@@ -441,14 +436,11 @@ class Downloader(
...
-        val tmpFile = tmpDir.findFile("$filename.tmp")
-
-        // Delete temp file if it exists
-        tmpFile?.delete()
 
         // Try to find the image file
         val imageFile = tmpDir.listFiles()?.firstOrNull {
-            it.name!!.startsWith("$filename.") || it.name!!.startsWith("${filename}__001")
+            (it.name!!.startsWith("$filename.") || it.name!!.startsWith("${filename}__001")) &&
+                !it.name!!.endsWith(".tmp")
```
AntsyLich: Block is only relevant for completed downloads
```suggestion
            (it.name!!.startsWith("$filename.") || it.name!!.startsWith("${filename}__001"))
```

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
xMohnad: We still need to ignore `*.tmp` files as completed; otherwise they won’t be resumed or re-downloaded


Screenshot_2026-06-25-05-06-24-163_com mixplorer

AntsyLich: Couple of lines later
```kt
// If the image is already downloaded, do nothing. Otherwise download from network
```

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 ### Added
 - Add dedicated "Support Us" screen ([@AntsyLich](https://github.com/AntsyLich)) ([#3200](https://github.com/mihonapp/mihon/pull/3200))
 - Add a one time popup asking for donation from long time users ([@AntsyLich](https://github.com/AntsyLich)) ([#3203](https://github.com/mihonapp/mihon/pull/3203))
+- Resumable image downloads if supported by server ([@xMohnad](https://github.com/xMohnad)) ([#3167](https://github.com/mihonapp/mihon/pull/3167))
```
AntsyLich: Move it again. under unreleased

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
xMohnad: Is this appropriate?

```kotlin
        // Try to find the image file
        val candidateFile = tmpDir.listFiles()?.firstOrNull {
            it.name!!.startsWith("$filename.") || it.name!!.startsWith("${filename}__001")
        }

        // Only treat as completed if it's not a tmp file
        val imageFile = candidateFile?.takeIf { !it.name!!.endsWith(".tmp") }
```

AntsyLich: ah sorry. the amount of `!` here made me confused. instead of the original do. this is much cleaner imo
```kt
tmpDir.listFiles()?.firstOrNull {
    val filename = it.name
    if (filename == null || filename.endsWith(".tmp")) return@firstOrNull null
    filename.startsWith("$filename.") || filename.startsWith("${filename}__001")
}
```

CHANGELOG.md
```diff
@@ -11,6 +11,8 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - `Other` - for technical stuff.
 
 ## [Unreleased]
+### Added
+- Resumable image downloads if supported by server ([@xMohnad](https://github.com/xMohnad)) ([#3167](https://github.com/mihonapp/mihon/pull/3167))
```
```suggestion
- Support resumable image downloads if supported by source ([@xMohnad](https://github.com/xMohnad)) ([#3167](https://github.com/mihonapp/mihon/pull/3167))
```

AntsyLich (approved):

PR #3168, opened 2026-03-29
Update strings.xml

ShaLimar-wep:


MajorTanya:
Edit: Checked your translation and it's not a real contribution. Please go away with useless PR spam.

PR #3170, opened 2026-03-30
Fix SQLDelight migration verification task complaining

MajorTanya:
Edit: This PR now also added the discussed migration verification task run in CI.

While figuring out how to fix my dailying of #3047 with the ID change I made, I was looking into SQLDelight a bit more and found out about the `verifySqlDelightMigration` task. Out of curiosity, I ran it but it return an error:

```txt
Execution failed for task ':data:verifyDebugDatabaseMigration'.
> A failure occurred while executing app.cash.sqldelight.gradle.VerifyMigrationTask$VerifyMigrationAction
   > Error Reading 9.sqm:

    C:/Users/User/AndroidStudioProjects/mihon/data/src/main/sqldelight/tachiyomi/migrations/9.sqm:21:37 ',', <compound operator real>, FROM, GROUP, HAVING, LIMIT, ORDER, WHERE or WINDOW expected, got 'excludedScanlator'
    04    CREATE VIEW updatesView AS
    05    SELECT
    06        mangas._id AS mangaId,
    07        mangas.title AS mangaTitle,
    08        chapters._id AS chapterId,
    09        chapters.name AS chapterName,
    10        chapters.scanlator,
    11        chapters.url AS chapterUrl,
    12        chapters.read,
    13        chapters.bookmark,
    14        chapters.last_page_read,
    15        mangas.source,
    16        mangas.favorite,
    17        mangas.thumbnail_url AS thumbnailUrl,
    18        mangas.cover_last_modified AS coverLastModified,
    19        chapters.date_upload AS dateUpload,
    20        chapters.date_fetch AS datefetch,
    21        excluded_scanlators.scanlator as excludedScanlator
                                                ^^^^^^^^^^^^^^^^^
    22    FROM mangas JOIN chapters
    23    ON mangas._id = chapters.manga_id
    24    LEFT JOIN excluded_scanlators
    25    ON mangas._id = excluded_scanlators.manga_id
    26    AND chapters.scanlator = excluded_scanlators.scanlator
    27    WHERE favorite = 1
    28    AND date_fetch > date_added
    29    ORDER BY date_fetch DESC
```

This was because the `as` keyword was lowercased, which SQLDelight takes umbrage with. I checked and SQLite itself does not care at all, and extracting a DB from my install shows the migration was applied (the `updatesView` has the `excludedScanlator` column).

```console
>sqlite3
SQLite version 3.52.0 2026-03-06 16:01:44
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> CREATE TABLE t (x INTEGER);
sqlite> INSERT INTO t VALUES (1);
sqlite> CREATE VIEW tview AS SELECT x as eeeeee FROM t;
sqlite> SELECT * FROM tview;
╭────────╮
│ eeeeee │
╞════════╡
│      1 │
╰────────╯
```

So this is a purely SQLDelight skill issue.

---

We should strongly consider running the `verifySqlDelightMigration` task in CI in my opinion.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,6 +51,9 @@ jobs:
         id: unit_tests
         run: ./gradlew testReleaseUnitTest
 
+      - name: Verify SQLDelight migrations
+        run: ./gradlew verifySqlDelightMigration
+
       - name: Upload test report
         if: steps.unit_tests.outcome == 'failure'
         uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0


MajorTanya:
> We should strongly consider running the verifySqlDelightMigration task in CI in my opinion.

@AntsyLich what do you say?

Hypothetically, this could be done as a filtered workflow step that only runs if the migration directory is modified I believe. I could look into it if you'd like.

AntsyLich:
Sure, just do it in this PR

MajorTanya:
> Sure, just do it in this PR

Sadly I found out that you can't simply skip one step of a workflow, and it seemed stupid to me to make this a whole new workflow just to run the pretty fast migration verification.

PR #3171, opened 2026-03-30
Migrate Manga Screen Long Click Range Selection

NarwhalHorns:
long click on manga migrate screen to select the range of items between previous selection and long held one, just used the same logic as the library screen.


NarwhalHorns:
added a limit of 50 to batch migration selection as @MajorTanya didn't want people batching too much at once. Dunno if that appeases you or not but. 

If you also want to limit batch migrations from the library screen we could add a limit there too?

PR #3179, opened 2026-04-01
Always add manga to update list if not initialized

NarwhalHorns:


cuong-tran:
Won't this flood the Updates screen with lots of chapters?

NarwhalHorns:
> Won't this flood the Updates screen with lots of chapters?

uhh this change shouldn't affect that since this is just making uninitialised manga get metadata which occurs before the chapter fetch. pretty sure. I have a few PRs in this area on the go so not 100% sure.

NarwhalHorns:
yeah, I checked on both main and this branch and both updates still get flooded 👍 no change 🤣

and actually has made me notice something of a flaw in the updates? the updates button in the bar gets labelled with a shit load of updates to view but i open it and there is dramatically less... something is fucked. but it's in main and this branch so a different issue.

PR #3184, opened 2026-04-04
Fix: missing nord outlinevariant base color

CompileConnected:
Execution failed for task ':presentation-core:lintDebug'.
> Lint found errors in the project; aborting build.
  
  Fix the issues identified by lint, or create a baseline to see only new errors.
  To create a baseline, run `gradlew updateLintBaseline` after adding the following to the module's build.gradle file:
  ```
  android {
      lint {
          baseline = file("lint-baseline.xml")
      }
  }
  ```
  For more details, see https://developer.android.com/studio/write/lint#snapshot
  
  Lint found 1 error. First failure:
  
  /Users/yohanes/Documents/GitHub/mihon/presentation-core/src/main/res/values-night/colors_nord.xml:27: Error: The color "nord_outlineVariant" in values-night has no declaration in the base values folder; this can lead to crashes when the resource is queried in a configuration that does not match this qualifier [MissingDefaultResource]
      <color name="nord_outlineVariant">#90939a</color>
             ~~~~~~~~~~~~~~~~~~~~~~~~~~
  
     Explanation for issues of type "MissingDefaultResource":
     If a resource is only defined in folders with qualifiers like -land or -en,
     and there is no default declaration in the base folder (layout or values
     etc), then the app will crash if that resource is accessed on a device
     where the device is in a configuration missing the given qualifier.
  
     As a special case, drawables do not have to be specified in the base
     folder; if there is a match in a density folder (such as drawable-mdpi)
     that image will be used and scaled. Note however that if you  only specify
     a drawable in a folder like drawable-en-hdpi, the app will crash in
     non-English locales.
  
     There may be scenarios where you have a resource, such as a -fr drawable,
     which is only referenced from some other resource with the same qualifiers
     (such as a -fr style), which itself has safe fallbacks. However, this still
     makes it possible for somebody to accidentally reference the drawable and
     crash, so it is safer to create a default fallback in the base folder.
     Alternatively, you can suppress the issue by adding
     tools:ignore="MissingDefaultResource" on the element.
  
     (This scenario frequently happens with string translations, where you might
     delete code and the corresponding resources, but forget to delete a
     translation. There is a dedicated issue id for that scenario, with the id
     ExtraTranslation.)
  
  
  The full lint text report is located at:
    /Users/yohanes/Documents/GitHub/mihon/presentation-core/build/intermediates/lint_intermediate_text_report/debug/lintReportDebug/lint-results-debug.txt

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
> Get more help at https://help.gradle.org.

BUILD FAILED in 20s


MajorTanya:
I'd be curious to know how you found this issue and how you went about finding the fix

CompileConnected:
Usually i just run ./gradlew build assemblePreview and it works, but it just happen today.
I can't run everything that previously works before.
it shocking that the theme is never been updated for 2 years, but the app works and build just fine.

MajorTanya:
Weird, I also run assemblePreview all the time (I use a custom nightly build off of `main` directly) and have never seen this issue pop up

MajorTanya:
Were you editing something in those files before you ran the tasks by chance?

CompileConnected:
nope, re check everything im just editing the kotlin code, nothing change the build.gradle or settings.
clean it, and re build from main resulting the same
this is my current android studio

Android Studio Panda 1 | 2025.3.1 Patch 1
Build #AI-253.29346.138.2531.14876573, built on February 12, 2026
Runtime version: 21.0.9+-14649483-b1163.86 aarch64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.

MajorTanya:
Can reproduce this, but weirdly only if I run `presentation-core:lintDebug` specifically as opposed to the general `lintDebug`... Also, I get 1000+ lint errors when I run the general one, but most are about string translation issues 😕

> nope, re check everything im just editing the kotlin code, nothing change the build.gradle or settings.
> clean it, and re build from main resulting the same
> this is my current android studio

I believe that gradle (sometimes?) only runs tasks for changed or at least touched files. Did you work in the presentation-core module with the Kotlin you were editing?

CompileConnected:
yes, i want to try the storage viewer from another contributor so i copy his commit

MajorTanya:
> yes, i want to try the storage viewer from another contributor so i copy his commit

I see, then Gradle probably woke up and checked the that module and found this. 

Where did you get this colour for outlineVariant in light mode from?

CompileConnected:
i just ask gemini which color appropriate for the nord theme

MajorTanya:
I see. Seems Gemini pulled that from the [Nord theme website](https://www.nordtheme.com/docs/colors-and-palettes) since #D8DEE9 is one of the light theme version colours. It'd be the same as inverseSurface but that should probably be fine I think.

Please add the colour to `app/src/main/java/eu/kanade/presentation/theme/colorscheme/NordColorScheme.kt` in the lightTheme value as well.

(We have to define each theme twice because we use both Compose and legacy XML views)

CompileConnected:
Done

AntsyLich (approved):

PR #3203, opened 2026-04-08
Add a one time popup asking for donation from long time users

AntsyLich:
Also remove the empty "Updated to X.Y.Z" popup after app update


KraXen72:
ach ja mihon sellout, i see ze developer is commercializing this product, i will blog about zis

PR #3246, opened 2026-04-20
Always update metadata on library update and remove related options

AntsyLich:


MajorTanya:
Doesn't this lead to more requests to source websites?

AntsyLich:
tachiyomix 1.6 will reduce the request since it has a single method to get manga update and chapter list

MajorTanya:
> tachiyomix 1.6 will reduce the request since it has a single method to get manga update and chapter list

Does that mean this will be merged after #3243 or am I missing something here

AntsyLich:
Both will be part of the same release

PR #3254, opened 2026-04-22
Add Yamtrack as a new tracker service

Breezyslasher:
Adds Yamtrack as a new tracker service, letting users sync manga reading progress to a self-hosted Yamtrack instance alongside the existing trackers (MyAnimeList, AniList, Kitsu, etc.).
Features

    Self-hosted login — user provides their Yamtrack host URL + API token; credentials are verified against /api/v1/statistics/ before being saved.
    Search — queries /api/v1/search/manga/ and returns matches from Yamtrack's upstream providers (MangaUpdates, MAL, AniList, etc.), plus a "manual entry" fallback for items those providers don't index (mirrors Yamtrack's own source=manual flow).
    Bind / Track — first-time tracking POSTs to /media/manga/; the response's tracked flag is checked so already-tracked items skip the add and go straight to update.
    Update — PATCHes status, progress, score, and reading dates to /media/manga/{source}/{media_id}/.
    Refresh — pulls the latest consumption record back into the local track.
    Delete — removes the entry from the user's Yamtrack library.
    Reading dates — supportsReadingDates = true; start/end dates are serialized as ISO yyyy-MM-dd and round-tripped through the consumption record.
    Statuses — Planning, Reading, Paused, Completed, Dropped, mapped to Yamtrack's integer status codes (0–4) per api/helpers.py.
    Scores — 0.0–10.0 in 0.1 increments, submitted only when non-zero.

Implementation notes

    Stable remote IDs are derived by hashing (source, media_id) since Yamtrack identifies entries by a pair rather than a single integer.
    Tracking URLs use the form {base}/media/manga/{source}/{media_id} and are parsed back to (source, media_id) for API calls.
    max_progress = 1 is treated as "unknown" (Yamtrack's placeholder for ongoing manga with no chapter count) and mapped to total_chapters = 0, so the progress picker isn't capped at chapter 1.
    Auth is applied via a dedicated YamtrackInterceptor using Bearer token headers.
    Uses the official Yamtrack logo (android-chrome-512x512.png) as the tracker brand image.

## Summary by Sourcery

Add Yamtrack as a new self‑hosted tracker integration alongside existing tracking services.

New Features:
- Introduce a Yamtrack tracker implementation that supports login via host URL and API token, search, bind, update, refresh, and delete operations for manga tracking.
- Expose Yamtrack as a configurable tracker option in the tracking settings screen.
-

Enhancements:
- Map Yamtrack-specific statuses, scores, dates, and remote identifiers into the app’s generic tracking model, including a manual-entry flow and handling of unknown chapter counts.


app/src/main/java/eu/kanade/tachiyomi/data/track/TrackerManager.kt
```diff
@@ -28,8 +29,10 @@ class TrackerManager {
     val mangaUpdates = MangaUpdates(7L)
     val kavita = Kavita(KAVITA)
     val suwayomi = Suwayomi(9L)
+    val yamtrack = Yamtrack(10L)
```
MajorTanya: before I even give this the light of day, absolutely not ID 10

At least pretend to pay attention to the repository you're trying to contribute to

Breezyslasher: Updated ID to 16, I could change it yamtrack = Yamtrack(YAMTRACK) if you don't want me to use a number?

MajorTanya: I said no to 10L because #1386 is pending with ID 10L and #3047 is pending with ID 11L.

The only reason you'd need to make Yamtrack's ID a constant like AniList's would be if you needed to reference it somewhere else, like AniList does for the score type preference.

Breezyslasher: OK I will keep it as 16.

Breezyslasher:
All features are done.

dj5kvjj4jn-alt:
+1

PR #3267, opened 2026-04-25
perf(backup): batch database operations during restore

Lolle2000la:
Reduce overhead by chunking manga and repo restoration into transactions of 100 entries. Categories now restored in a single transaction. Optimized DatabaseHandler to reuse existing transaction contexts instead of creating redundant nested ones.

Changed TransactionElement to internal to allow AndroidDatabaseHandler to check transaction state from the same package.

|Device|Before|After|
|-------|-------|-----|
|Pixel 9 XL emulator image¹|image|image|
|S24 Ultra²|Screenshot_20260425_132402_One UI Home|Screenshot_20260425_131722_One UI Home|

Much simpler and more efficient than #2475. A higher blocking factor yields higher performance on my emulator and lower on my phone, a lower one (i.e. 50) yields less on both. High blocking factor helps with libraries with many small entries (with few chapters), but I don't have statistics on how the users libraries are distributed.

¹ PC Specs: Ryzen 9 9950X, Samsung 990 Pro 4TB
² Snapdragon variant


AntsyLich:
I'm actually working on redoing database 🫠

Lolle2000la:
Wow, very quick response! Thank you!

---

I mean this changes nothing on the database, just batches the write requests. Its a fairly standard thing. Just wraps the existing logic with a context so that it makes bigger updates instead of many small ones. The degree devices benefit from that is basically how high their storage latency is as opposed to bandwidth. That's why my Emulator on a fast SSD only has a ~2x speedup and the smartphone has a ~4x one. The lower end the devices, the higher the expected speedup (until bandwidth limit is hit of course).

I compared it to the previous PR because this is very very simple and gives a good performance bump with much less changes.

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/BackupRestorer.kt
```diff
@@ -125,18 +129,23 @@ class BackupRestorer(
...
+                        ensureActive()
+
+                        try {
+                            mangaRestorer.restore(it, backupCategories)
+                        } catch (e: Exception) {
+                            val sourceName = sourceMapping[it.source] ?: it.source.toString()
+                            errors.add(Date() to "${it.title} [$sourceName]: ${e.message}")
+                        }
+
+                        restoreProgress += 1
```
AntsyLich: `restoreProgress` needs to be atomic.

AntsyLich (changes requested):

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/BackupRestorer.kt
Lolle2000la: Fixed that and synchronized the error collection as well.

Lolle2000la:
Regarding [efa4d20](https://github.com/mihonapp/mihon/pull/3267/commits/efa4d203d19e9314e06f5ebe1a149b99753a072f), I got another good speedup by not updating the notification on literally every element. It turns out that this is actually itself quite an expensive operation.

It now happens on every chunk, leading to another speedup from 46 seconds on my emulator to 27 seconds.

Lolle2000la:
Merged with main with the changes from #3270 and adjusted to the different API. Got this speedup against my emulator:

Main:
image

This branch:
image

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/BackupRestorer.kt
```diff
@@ -19,26 +19,32 @@ import kotlinx.coroutines.coroutineScope
...
+import tachiyomi.data.Database
 import tachiyomi.i18n.MR
+import uy.kohesive.injekt.Injekt
+import uy.kohesive.injekt.api.get
 import java.io.File
 import java.text.SimpleDateFormat
+import java.util.Collections
 import java.util.Date
 import java.util.Locale
+import java.util.concurrent.atomic.AtomicInteger
```
AntsyLich: Use the kotlin version

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/BackupRestorer.kt
```diff
@@ -19,26 +19,32 @@ import kotlinx.coroutines.coroutineScope
...
     private val preferenceRestorer: PreferenceRestorer = PreferenceRestorer(context),
     private val extensionStoreRestorer: ExtensionStoreRestorer = ExtensionStoreRestorer(),
     private val mangaRestorer: MangaRestorer = MangaRestorer(),
 ) {
 
     private var restoreAmount = 0
-    private var restoreProgress = 0
-    private val errors = mutableListOf<Pair<Date, String>>()
+    private val restoreProgress = AtomicInteger(0)
+    private val errors = Collections.synchronizedList(mutableListOf<Pair<Date, String>>())
```
AntsyLich: Use CopyOnWriteArrayList

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/BackupRestorer.kt
```diff
@@ -125,18 +131,23 @@ class BackupRestorer(
...
         mangaRestorer.sortByNew(backupMangas)
-            .forEach {
-                ensureActive()
-
-                try {
-                    mangaRestorer.restore(it, backupCategories)
-                } catch (e: Exception) {
-                    val sourceName = sourceMapping[it.source] ?: it.source.toString()
-                    errors.add(Date() to "${it.title} [$sourceName]: ${e.message}")
+            .chunked(100)
```
AntsyLich, line 137: Any particular reason for this value? also why chunked?

AntsyLich (changes requested):
Can't say I particularly like the notification not being synced for a chunk. and add an entry to changelog

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/BackupRestorer.kt
Lolle2000la: I wrote down the reasoning on that in the PR description

>A higher blocking factor yields higher performance on my emulator and lower on my phone, a lower one (i.e. 50) yields less on both.

Basically, do too much in one transaction and you get diminishing returns while preventing other writes. Past a certain point you also just wait for storage to flush the cached writes (like on my phone, where it got slower on a higher factor).

> High blocking factor helps with libraries with many small entries (with few chapters), but I don't have statistics on how the users libraries are distributed.

If a library is only oneshots, a higher factor might help more (though not that much), conversly, a library where every entry has hundreds of chapters might somewhat run into the performance issues, though still better than not chunked.

Honestly, I didn't want to microtest to see if some value between 80-120 or something is more ideal since it is probably well within device-to-device variance and most phone storage is almost equally shitty, so I just left it at a nice and round 100.

AntsyLich: alright

Lolle2000la:
> Can't say I particularly like the notification not being synced for a chunk. and add an entry to changelog

[It is though?](https://github.com/mihonapp/mihon/pull/3267/changes#diff-a21058fc99536673e8b47a24e21aadb4322564e6e4b5f47f6ef5c3ae08ced768R150) Or do I misunderstand what you mean? 

The notification is updated on every successful chunk transaction.

AntsyLich:
> > Can't say I particularly like the notification not being synced for a chunk. and add an entry to changelog
> 
> [It is though?](https://github.com/mihonapp/mihon/pull/3267/changes#diff-a21058fc99536673e8b47a24e21aadb4322564e6e4b5f47f6ef5c3ae08ced768R150) Or do I misunderstand what you mean?
> 
> The notification is updated on every successful chunk transaction.

That's what I meant. It might appear to the user nothing is happening.

Lolle2000la:
It should still happen quickly enough

> > > Can't say I particularly like the notification not being synced for a chunk. and add an entry to changelog
> > 
> > 
> > [It is though?](https://github.com/mihonapp/mihon/pull/3267/changes#diff-a21058fc99536673e8b47a24e21aadb4322564e6e4b5f47f6ef5c3ae08ced768R150) Or do I misunderstand what you mean?
> > The notification is updated on every successful chunk transaction.
> 
> That's what I meant. It might appear to the user nothing is happening.

It should happen quickly enough, though I definitely see what you mean. The problem is that in essence, doing the update in the chunks `forEach` would go by almost instantly since nothing is done until the transaction is submitted, so it wouldn't make much of difference. But the updating of the notification is quite expensive IPC, so it would actually slow that down.

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/BackupRestorer.kt
Lolle2000la: The kotlin `AtomicInt` is still experimental and requires opt-in. Is that ok?

AntsyLich: it's used everywhere else in the app

AntsyLich:
> It should happen quickly enough, though I definitely see what you mean. The problem is that in essence, doing the update in the chunks `forEach` would go by almost instantly since nothing is done until the transaction is submitted, so it wouldn't make much of difference. But the updating of the notification is quite expensive IPC, so it would actually slow that down.

I suppose we can leave it as is

app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/BackupRestorer.kt
Lolle2000la: Ok, I just applied it, and if you decide not to, you can tell me and I will roll back that part of the change.

Lolle2000la:
The latest CI failure doesn't appear to come from me. I merged the latest changes from main and that broke it.

AntsyLich (approved):

Issue #1152, opened 2024-08-24
Sub-Categories feature


maxshovel:
### Describe your suggested feature

Description:
Add support for sub-categories to improve manga organization and browsing experience.

Proposed functionality:
1. Allow creation of sub-categories within main categories
2. Enable assigning manga to both main and sub-categories
3. Update UI to display and navigate sub-categories
4. Implement filtering and sorting options for sub-categories

Would you like me to explain or elaborate on any part of this issue request?

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.16.5](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

wasu-code:
Maybe UI for subcategories can be something similar to the **first screenshot** from TachiyomiJ2K:  
![TachiyomiJ2K](https://raw.githubusercontent.com/Jays2Kings/tachiyomiJ2K/refs/heads/master/.github/readme-images/screens.gif) 
Namely: subcategories listed in a vertical view, that can be collasped or expanded with a tap.

So grouping **categories** will work as it does now (tabs in top part of the UI) and **subcategories** will be verical collapsible sections.

NarwhalHorns:
So i was playing around with this concept and thought there was a few different ways it could be approached. Wondering what people are thinking themselves as to how this would look/shape up in concept.

Personally I had a few ideas being:
- **Super Categories** Where we just have 2 levels of categories. Supers would sit above normal ones and basically just add a second layer of categorisation but independent of regular categories. (This is my preferred idea, think it's simpler and easier to grasp and probably to code)
	- So as example, you could have super categories for reading, planning, dropped etc. and then regular categories for genres or something. And in execution it would look as if all the regular categories were splits under the super splits. If that makes any sense? Hard to explain
- **Sub Categories** This would be far more complex and intricate but would allow for further customisation. What I expect most people are thinking of when we think of this concept. Categories that are explicitly tied to a parent category (or perhaps multiple) that split the categories as desired.


Super UI looking something like this i guess?
<img width="1411" height="395" alt="Image" src="https://github.com/user-attachments/assets/c0303929-d9df-4a2f-a49e-66d08b89385f" />

No idea what a UI for specific sub-categories could be like unless we change how the library categories work a lot? idk

AntsyLich:
[...]

PR #3284, opened 2026-05-02
Pinned Categories

NarwhalHorns:
Something _like_ subcategories but _not_. As (briefly) discussed on #1152, this is something of an expanstion on categories as we have now as that is significantly easier to put into place on top of the current system and can serve as subcategories if so desired.

Basically implemented as being able to pin categories to a second set of categories on top of the regular set becoming 'Pinned Categories'. These pinned categories are then grouped outside the regular category groups leading to a double layer of grouping in the library. In effect the user basically has a distinct set of categories within each pinned category, though that is only when used as such. Since in the background they are just another category, if unpinned they go back to being a regular category with no issues.

**if y'all are interested or have any input GIMME**

## TODOs
- Category stuff in settings
	- Default category
		1. Split into default pin and default regular?
		2. Assume no pinned cat for defaults?
	- Global updates
		1. Seperate settings per pinned category?
		2. Pinned category selection over rules underlying categories?
		3. regular category settings overule pinned category settings?
		4. Only settings for regular categories?
- Remember category across different pinned categories in library
- Probably improve choose category dialog visuals (I don't like much but idk what's better)
- 'Show category tabs' off things
	- Tab title - <Pinned cat - cat> maybe?
	- Can't get to other pinned since not in pager
		- Library pager with pinned categorys one after the other?
- Library 'update category' update by pinned too

## Category select bars
<table>
<tr>
 <td> Default (when given no pinned category as normal)
 <td> As part of a pinned category, effectively giving a second library of categories.
<tr>
 <td> image
 <td> image
</table>

## Edit Categories Screen
Categories can be pinned on this screen to bring them to a second level of categorization.
<table>
<tr>
 <td> No pinned cats
 <td> With pinned cat
<tr>
 <td> image
 <td> image
</table>

## Choose Categories Dialog
Pinned Categories seperated on this screen
<table>
<tr>
 <td> No pinned cats
 <td> With pinned cat
<tr>
 <td> image
 <td> image
</table>

### DIFF
--- a/app/src/main/java/eu/kanade/domain/DomainModule.kt
+++ b/app/src/main/java/eu/kanade/domain/DomainModule.kt
@@ -49,12 +49,14 @@ import tachiyomi.data.updates.UpdatesRepositoryImpl
 import tachiyomi.domain.category.interactor.CreateCategoryWithName
 import tachiyomi.domain.category.interactor.DeleteCategory
 import tachiyomi.domain.category.interactor.GetCategories
+import tachiyomi.domain.category.interactor.PinCategory
 import tachiyomi.domain.category.interactor.RenameCategory
 import tachiyomi.domain.category.interactor.ReorderCategory
 import tachiyomi.domain.category.interactor.ResetCategoryFlags
 import tachiyomi.domain.category.interactor.SetDisplayMode
 import tachiyomi.domain.category.interactor.SetMangaCategories
 import tachiyomi.domain.category.interactor.SetSortModeForCategory
+import tachiyomi.domain.category.interactor.UnPinCategory
 import tachiyomi.domain.category.interactor.UpdateCategory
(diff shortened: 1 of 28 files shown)


cuong-tran:
similar to pinned sources?

NarwhalHorns:
no actually just got a tickbox when you make a new category atm but that's probably a better idea

MajorTanya:
So basically, subcategories.

NarwhalHorns:
whatever you wanna call em sure lmao

MajorTanya:
> whatever you wanna call em sure lmao

I called them that because that's what most of the previous feature request in the issue tracker have called them.

NarwhalHorns:
well to me subcategories would be category dependent, which these are not

PR #3300, opened 2026-05-07
Always update titles from source

luigidotmoe:
I would expect that to be the default behavior
Changing the title preference in an extension’s settings and having it have no impact on my library entries unless this option is manually enabled feels really unintuitive to me

Removes the option from settings -> advanced 
image


BrutuZ:
Intuitiveness is a matter of perspective. Yours is you deliberately ticked a custom extension option that changes titles, for most users it would be something like: Source suddenly changed the title (either language or reworded) and now user can't find it in the library anymore

PR #3328, opened 2026-05-17
Add margins setting for paged view

pialony:
DISCLAIMER: This was done with AI assistance. However I did review and polish the code before submitting it.

There are situations when a user may desire to add some margin to the full screen paged view, but still retain the Scale type setting (such as Fit to screen and so on).

For example, (my own issue that I wanted to fix) there are some e-ink devices that cut the display a tiny bit, so when viewing on full display, part of the image is being cut.
Additionally, some viewers may want to avoid round corners or the camera notch of their display.

This change adds 4 sliders for each side and color selection of the margin in the app settings and in the viewer mode.
It retains how things are done, unless you add a margin, and will also revert back if margin is 0. So it should not affect anyone unless they actually use this option.
I've decided to add a section name of it's own in the settings for better visibility.
I put the max value of the sliders to 75px, as more than this makes the slider buggy, but it's good enough to avoid the camera notch on a few devices I've tested.

I would appreciate any feedback for improvement, or any other approach to take here.


pialony:
Related to https://github.com/Jays2Kings/tachiyomiJ2K/issues/1540

PR #3333, opened 2026-05-19
Refactor NotificationReceiver to use lifecycleScope and goAsync

karmatheHacker:


AntsyLich:
Hi, thanks for the PR. I took a bit different approach in #3403 alongside more changes.

PR #3340, opened 2026-05-22
Fix 25 bugs across the app

corecommit:
Found and fixed 25 bugs while poking around the codebase. Most were crashes or
NPEs I hit during normal use, plus a few tracker and data integrity issues.

Crash fixes:
- Version comparison crashed when segment counts didn't match
- Several NPEs from using !! on nullable values (migration, stats, icons, locale)
- NumberFormatException from parsing invalid version strings
- Track date picker crashes on null selection
- DummyTracker was throwing on client access

Tracker fixes:
- Komga/Kavita/Suwayomi search() threw NotImplementedError instead of returning empty results
- Chapter numbers with decimals (e.g. 12.5) got truncated by toLong(), causing false COMPLETED status
- Suwayomi wasn't filtering chapters on the server side
- MyAnimeList had a disabled User-Agent header
- A few trackers could crash on malformed manga URLs

Data fixes:
- DB exceptions were silently swallowed in chapter operations
- lastUpdate timestamp got updated even when nothing changed
- InMemoryPreferenceStore.changes() never emitted on set()
- getStringSet() threw instead of returning a valid preference
- Custom library interval was hidden behind a release build check

Other:
- mapIndexed used for side effects instead of forEachIndexed
- Deprecated trim(*CharArray) idiom replaced
- Double equality comparison without epsilon
- Removed stale TODOs and an unused OsTooOld variant
- Added a warning for legacy local source JSON format

### DIFF
--- a/app/src/main/java/eu/kanade/domain/chapter/interactor/SyncChaptersWithSource.kt
+++ b/app/src/main/java/eu/kanade/domain/chapter/interactor/SyncChaptersWithSource.kt
@@ -207,8 +207,13 @@ class SyncChaptersWithSource(
             chapterRepository.removeChaptersWithIds(toDeleteIds)
         }
 
+        var addAllSucceeded = false
         if (updatedToAdd.isNotEmpty()) {
-            updatedToAdd = chapterRepository.addAll(updatedToAdd)
+            val added = chapterRepository.addAll(updatedToAdd)
+            if (added.isNotEmpty()) {
+                updatedToAdd = added
+                addAllSucceeded = true
+            }
         }
 
         if (updatedChapters.isNotEmpty()) {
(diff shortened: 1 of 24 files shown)


AntsyLich:
take your ai slop somewhere else

corecommit:
> take your ai slop somewhere else

You scared that ai will take ur job?

Smol-Ame:
Mihon is an open-source project built & maintained by people contributing their own time & efforts, there is no job. AI usage is only allowed if you have thoroughly reviewed & tested your changes.

Plus, mass "fixing" of different parts of the app makes it hard to review/separate if there's a regression. If you're going to contribute, contribute with effort.

corecommit:
> Mihon is an open-source project built & maintained by people contributing their own time & efforts, there is no job. AI usage is only allowed if you have thoroughly reviewed & tested your changes.

Hmm ok

AntsyLich:
> > take your ai slop somewhere else
> 
> You scared that ai will take ur job?

what's there to be scared of when it does a shittier job than an intern

corecommit:
> > > take your ai slop somewhere else
> > 
> > You scared that ai will take ur job?
> 
> what's there to be scared of when it does a shittier job than an intern

😂

PR #3349, opened 2026-05-25
 Change extension repo to extension store and add support for newer extension index format

AntsyLich:


ZerOri:
A major change!😮

PR #3388, opened 2026-06-06
Add high quality webgpu renderer

wwww-wwww:
New renderer that uses WebGPU and libvips for decoding images.
Completed features: 
- Area filter for downsampling (No more moire)
- Catmull-rom for upsampling
- Crop zooms instead
- Avoid display cutout
- Page transition animations
  - Basic
  - Stack
  - Fade
  - Fade to white
  - Cube
  - Cube Outer
  - Flip
  - Sphere
- Animated images
- Continuous vertical view
- Background color
  - Auto detect color
- Download progress
- Transition page
- Double page mode
  - Can use Exif tag PageNumber "Left" "Right" "Center"
- Landscape images zoom on start

What's not complete:
- Color filters
- Double page detection without a tag
  - How to detect dropped pages (commonly after a double spread at the beginning of a chapter)
  - Gesture to toggle page offset?

Can be enabled under Settings > Advanced > Render > Use high quality renderer
Currently off by default

Tested on:
Z Fold 8 - Qualcomm Snapdragon 8 Elite Gen 5 for Galaxy (Adreno 840) - >120fps (~900fps)
Z Fold 6 - Qualcomm Snapdragon 8 Gen 3 for Galaxy (Adreno 750) - >120fps (~400fps)
S23 - Qualcomm Snapdragon 8 Gen 2 for Galaxy (Adreno 740) - >120fps
S10e - Qualcomm Snapdragon 855 (Adreno 640) - >60fps (~300fps)
Pixel 10 - Google Tensor G5 (PowerVR DXT-48-1536) - >120fps (~300fps)
Pixel 8a - Google Tensor G3 (Mali-G715 MP7) - >120fps
moto g pure - Mediatek MT6762G Helio G25 (PowerVR GE8320) - ~30 fps

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -99,6 +99,7 @@
     public <init>();
     public void destroy();
 }
+
 # webgpu
 -keep interface androidx.webgpu.** { *; }
 -keep class androidx.webgpu.** { *; }
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -99,8 +99,3 @@
     public <init>();
     public void destroy();
 }
-
-# webgpu
--keep interface androidx.webgpu.** { *; }
--keep class androidx.webgpu.** { *; }
--keep class ca.mpreg.** { *; }
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -99,3 +99,6 @@
     public <init>();
     public void destroy();
 }
+# webgpu
+-keep interface androidx.webgpu.** { *; }
+-keep class androidx.webgpu.** { *; }
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -102,3 +102,4 @@
 # webgpu
 -keep interface androidx.webgpu.** { *; }
 -keep class androidx.webgpu.** { *; }
+-keep class ca.mpreg.** { *; }
--- a/app/src/main/java/eu/kanade/presentation/reader/settings/ReadingModePage.kt
+++ b/app/src/main/java/eu/kanade/presentation/reader/settings/ReadingModePage.kt
@@ -252,4 +252,15 @@ private fun ColumnScope.WebGpuViewerSettings(screenModel: ReaderSettingsScreenMo
             )
         }
     }
+
+    val cutoutMode by screenModel.preferences.cutoutMode.collectAsState()
+    SettingsChipRow(MR.strings.pref_cutout_mode) {
+        ReaderPreferences.CutoutMode.entries.map {
+            FilterChip(
+                selected = it == cutoutMode,
+                onClick = { screenModel.preferences.cutoutMode.set(it) },
+                label = { Text(stringResource(it.titleRes)) },
+            )
+        }
+    }
 }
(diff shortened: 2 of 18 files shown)


AntsyLich:
> Currently off by default but I want to get this tested in more hands
>
Maybe I'll finally have a reason to complete the incomplete feature flag stuff

wwww-wwww:
Maybe it can be a-b tested as well

app/proguard-rules.pro
```diff
@@ -80,3 +80,7 @@
 # Firebase
 -keep class com.google.firebase.installations.** { *; }
 -keep interface com.google.firebase.installations.** { *; }
+
+# webgpu
+-keep interface androidx.webgpu.** { *; }
+-keep class androidx.webgpu.** { *; }
```
AntsyLich, line 105: This should be part of the library imo.

AntsyLich (changes requested):

Palloxin:
Question, does it use less battery? less cpu etc

app/src/main/java/eu/kanade/presentation/reader/settings/ReadingModePage.kt
```diff
@@ -56,6 +79,9 @@ internal fun ColumnScope.ReadingModePage(viewModel: ReaderSettingsViewModel) {
         WebtoonViewerSettings(viewModel)
     } else {
         PagerViewerSettings(viewModel)
+        if (viewer is WebGpuViewer) {
+            WebGpuViewerSettings(viewModel)
+        }
```
AntsyLich, line 84: isn't this and the other 2 mutually exclusive?

wwww-wwww: it's only re-using PagerViewerSettings for WebGpuViewer, or I am misunderstanding

AntsyLich: Nvm confused with something else

Drazzy9295:
A few issues I noticed while testing the new viewer:

1. When moving through the chapter navigator quickly, the viewer flickers/jitters between two pages. It’s pretty visually distracting and makes the navigator difficult to use at speed.

2. Compared to the legacy renderer, scrolling in the continuous viewer loses momentum much more quickly after you let go. Page swipes feel noticeably heavier instead of gliding smoothly like they did before.

3. Double-tapping to zoom has some strange movement and doesn’t seem to zoom into the expected position correctly.

4. If you double-tap to zoom in and then double-tap again to zoom back out, you don’t return to your original position on the page. Repeatedly zooming in and out effectively causes you to scroll through the page.

AntsyLich:
@Drazzy9295 please make separate proper issues to better help track it

wwww-wwww:
and specify that 3 and 4 are for continuous view

Drazzy9295:
@AntsyLich @wwww-wwww Done 👍

https://github.com/mihonapp/mihon/issues/3779
https://github.com/mihonapp/mihon/issues/3780
https://github.com/mihonapp/mihon/issues/3781

Issue #3249, opened 2026-04-20
Make double tap to zoom an option in manga as well as webtoon and when disabled remove tap delay in reader


muffinl0rd:
### Describe your suggested feature

Currently double tap to zoom is the only usage of double tapping in the mihon reader. This is not a requirement as pinch to zoom exists and can only be disabled in the webtoon version.

When double tap is not used there is no reason to use onSingleTapConfirmed as it adds a 300ms delay by default before a single tap can have any feedback.

This value is set here:
https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/view/ViewConfiguration.java;l=148

Used here:
https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/view/GestureDetector.java

And mentioned in the documentation here:
https://developer.android.com/reference/android/view/GestureDetector.SimpleOnGestureListener#onSingleTapConfirmed(android.view.MotionEvent)

1. Disabling double tap to zoom should be added to the manga version of the reader as well.
2. mihon should only use onSingleTapConfirmed when double tap to zoom is enabled
3. When double tap to zoom is disabled onSingleTapUp should be used instead as it has no extra delay

I don't have any experience with programming android apps but was able to get a demo working using onSingleTapUp
It is quite limited as I haven't done anything about disabling double tap to zoom or adding the preference.
I would love it if someone experienced with this codebase was willing to properly implement this toggle.

Attached is a video that shows me moving pages 5 times then opening and closing the menu first in the current build of mihon, then in the modified build. I am tapping as soon as the UI responds in the current build (any faster would be interpreted as zooming) then as fast I can for the modified, the speed difference is huge.

https://github.com/user-attachments/assets/30a95bf4-5926-4cfd-8ae0-1f9f7b99c9e0

It goes from an app which feels sluggish in the reader to responding in real time.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.19.9](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

Smol-Ame:
Duplicate of #815 & (most likely) part of #3011 (which is intended via #2680)

muffinl0rd:
It's an extension on #815 but #2680 did not introduce the slower tap detection.
[...]

PR #3402, opened 2026-06-11
Add "disable doubletap" option to paged reader

### LINKED ISSUE #815: Can not disable "double tap to zoom" in "paged" reading mode
### Steps to reproduce

1. open a comic book
2. setting the reading mode for this series
3. set the reading mode as "paged"
4. there is no option called "double tap to zoom"

### Expected behavior

Allow user disable "double tap to zoom" in "paged" reading mode.

### Actual behavior

Can not disable "double tap to zoom" in "paged" reading mode in the latest version.

### Crash logs

_No response_

### Mihon version

Beta r6697 (2f243fae1, 5/19/24 05:16)

### Android version

Android 14

### Device

Honor Magic V2

### Other details

_No response_

### Acknowledgements

rszyma:
Closes https://github.com/mihonapp/mihon/issues/815

This PR introduces a new option for paged reader, allowing to disable double tap to zoom. `WebtoonViewer` has pretty much identical option, but it doesn't apply to paged reader.

I decided go other impl route than trying to re-use the one from `WebtoonViewer`, as at first I thought I through that I found a simpler (though a bit hacky) way to effectively disable double tap. And it works okayish, only a bit weird behavior (mentioned in the "Future work" at the end).

One notable discovery I made during writing the code, is that when the new zoom-on-double-tap option is disabled page flips can be made much faster, now that we don't need to wait have a delay for tap confirmation (`onSingleTapConfirmed`). So I took a bit longer route to implement it to enable this lower delay path.

Future work: 
- double-tap-then-drag should not trigger scroll on 2nd tap release - this feels a bit buggy. 
	- `WebtoonViewer` in fact already contains gesture handler with `onDoubleTapConfirmed`, which we would something like this, and it could probably solve this issue. Looks like an angle for refactor, but feels like it would be a lot more changes than those I made.

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/manga/components/MangaCoverDialog.kt
+++ b/app/src/main/java/eu/kanade/presentation/manga/components/MangaCoverDialog.kt
@@ -178,7 +178,7 @@ fun MangaCoverDialog(
                                     ?.copy(Bitmap.Config.HARDWARE, false)
                                     ?.toDrawable(view.context.resources)
                                     ?: drawable
-                                view.setImage(copy, ReaderPageImageView.Config(zoomDuration = 500))
+                                view.setImage(copy, ReaderPageImageView.Config(zoomDuration = 500, doubleTapZoom = true))
                             }
                             .build()
                         view.context.imageLoader.enqueue(request)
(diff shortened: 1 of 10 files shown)


muffinl0rd:
I'm glad you noticed the opportunity to remove the delay in confirming a single tap like I had here https://github.com/mihonapp/mihon/issues/3249
Thanks for the help even if by coincidence.

PR #3417, opened 2026-06-16
fix(installer): report error on Shizuku installation failure

### LINKED ISSUE #3295: Shizuku: Unable to install extensions
### Steps to reproduce

1.set install programme to shizuku
2.try to install extersions 
3.it stopped

Image

4,set install programme to private or legacy
5.install successfully

### Expected behavior

it can install extersions by shizuku

### Actual behavior

it stopped and nothing happened
it doesnt crash

### Crash logs

[mihon_crash_logs.txt](https://github.com/user-attachments/files/27408254/mihon_crash_logs.txt)

### Mihon version

r7670

### Android version

android 14

### Device

mipad 5 pro and oneplus 12R

### Other details

btw, what is private in install programme

### Acknowledgements

EnderAvaritia:
Closes #3295

**Problem**: When installing extensions via Shizuku, any failure in `ShellInterface.install()` (running in Shizuku's privileged process) would crash the Shizuku service silently. The app-side error paths also failed to report errors back to the UI:

1. **ShellInterface.kt**: `install()` had no try/catch. A reflection failure on Android 13/14 (e.g. `createSession` API change) would kill the Shizuku process with no feedback to the app.
2. **ShizukuInstaller.kt**: `shizukuDeadListener` only called `service.stopSelf()` without calling `continueQueue(Error)`, leaving the install step stuck at "Installing".
3. **Installer.kt**: `onDestroy()` processed queue entries but silently dropped the active `waitingInstall` entry without reporting an error.

**Fix**:
- `ShellInterface.install()`: wrapped entire body in try/catch. On exception, sends a `STATUS_FAILURE` broadcast via the same `PendingIntent` used for commit results, so the existing `BroadcastReceiver` in `ShizukuInstaller` handles it.
- `ShizukuInstaller.shizukuDeadListener`: calls `continueQueue(InstallStep.Error)` before `service.stopSelf()`.
- `Installer.onDestroy()`: uses `waitingInstall.exchange(null)` to atomically retrieve and report the active entry's error, instead of silently nulling it.

**Testing**: Verified on OnePlus 12R (Android 14) , Mi Pad 5 Pro (Android 14) and MuMu emulator. Switching from Shizuku to Private/Legacy previously worked; after this fix Shizuku installation completes successfully with proper error feedback on failures.

**Relevant issue**: https://github.com/mihonapp/mihon/issues/3295

**Commits**:cb3aa33769c0b0e5d0cc5ab8c923827a13ce72ce

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/extension/installer/Installer.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/extension/installer/Installer.kt
@@ -117,7 +117,10 @@ abstract class Installer(private val service: Service) {
         LocalBroadcastManager.getInstance(service).unregisterReceiver(cancelReceiver)
         queue.forEach { extensionManager.updateInstallStep(it.downloadId, InstallStep.Error) }
         queue.clear()
-        waitingInstall.store(null)
+        val activeEntry = waitingInstall.exchange(null)
+        if (activeEntry != null) {
+            extensionManager.updateInstallStep(activeEntry.downloadId, InstallStep.Error)
+        }
     }
 
     protected fun getActiveEntry(): Entry? = waitingInstall.load()
--- a/app/src/main/java/eu/kanade/tachiyomi/extension/installer/Installer.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/extension/installer/Installer.kt
@@ -117,9 +117,8 @@ abstract class Installer(private val service: Service) {
         LocalBroadcastManager.getInstance(service).unregisterReceiver(cancelReceiver)
         queue.forEach { extensionManager.updateInstallStep(it.downloadId, InstallStep.Error) }
         queue.clear()
-        val activeEntry = waitingInstall.exchange(null)
-        if (activeEntry != null) {
-            extensionManager.updateInstallStep(activeEntry.downloadId, InstallStep.Error)
+        waitingInstall.exchange(null)?.let { entry ->
+            extensionManager.updateInstallStep(entry.downloadId, InstallStep.Error)
         }
     }
 
--- a/app/src/main/java/eu/kanade/tachiyomi/extension/installer/ShizukuInstaller.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/extension/installer/ShizukuInstaller.kt
@@ -70,6 +70,7 @@ class ShizukuInstaller(private val service: Service) : Installer(service) {
 
     private val shizukuDeadListener = Shizuku.OnBinderDeadListener {
         logcat { "Shizuku was killed prematurely" }
+        continueQueue(InstallStep.Error)
         service.stopSelf()
     }
 
--- a/app/src/main/java/mihon/app/shizuku/ShellInterface.kt
+++ b/app/src/main/java/mihon/app/shizuku/ShellInterface.kt
@@ -66,89 +66,103 @@ class ShellInterface : IShellInterface.Stub() {
 
     @SuppressLint("PrivateApi")
     override fun install(apk: AssetFileDescriptor) {
-        val pmInterface = Class.forName($$"android.content.pm.IPackageManager$Stub")
-            .getMethod("asInterface", IBinder::class.java)
-            .invoke(null, SystemServiceHelper.getSystemService("package"))
-
-        val packageInstaller = Class.forName("android.content.pm.IPackageManager")
-            .getMethod("getPackageInstaller")
-            .invoke(pmInterface)
-
-        val params = PackageInstaller.SessionParams(PackageInstaller.SessionParams.MODE_FULL_INSTALL).apply {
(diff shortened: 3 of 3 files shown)


infyProductions:
Does it work with other Shizuku forks like the thedjchi one?

EnderAvaritia:
> Does it work with other Shizuku forks like the thedjchi one?

thedjchi works 
but i dont know other shizuku forks

MajorTanya:
> if anyone want to have a try the link can be use for 30 days

Don't post random links please.

EnderAvaritia:
> > if anyone want to have a try the link can be use for 30 days
> 
> Don't post random links please.

does there any way to post the app i built 
what i build is about 50mb and i cant upload that
we need that app to test if it can work for other shizuku forks

EnderAvaritia:
i post the app to the release of my fork

Smol-Ame:
People can test it once the dev gets around to reviewing & approving your PR. In the meantime, make sure this PR branch is up-to-date with main.

app/src/main/java/eu/kanade/tachiyomi/extension/installer/Installer.kt
```diff
@@ -117,7 +117,10 @@ abstract class Installer(private val service: Service) {
         LocalBroadcastManager.getInstance(service).unregisterReceiver(cancelReceiver)
         queue.forEach { extensionManager.updateInstallStep(it.downloadId, InstallStep.Error) }
         queue.clear()
-        waitingInstall.store(null)
+        val activeEntry = waitingInstall.exchange(null)
+        if (activeEntry != null) {
+            extensionManager.updateInstallStep(activeEntry.downloadId, InstallStep.Error)
+        }
```
AntsyLich:
```suggestion
        waitingInstall.exchange(null)?.let { entry ->
            extensionManager.updateInstallStep(entry.downloadId, InstallStep.Error)
        }
```

AntsyLich (changes requested):
Does shizuku continue to work when just one installation fails?

EnderAvaritia:
> Does shizuku continue to work when just one installation fails?

i am not sure 
but i hope i have some methods to make the install fail manually 

the ai said:
> Yes. If a single APK installation fails (e.g. corrupted file or commit rejection), continueQueue(InstallStep.Error) is called and the next item in the queue proceeds normally. The only case where remaining installs are cancelled is when the Shizuku process itself crashes — the shizukuDeadListener then stops the service, which marks all queued entries as Error in onDestroy().

NGB-Was-Taken:
> 1. **ShellInterface.kt**: `install()` had no try/catch. A reflection failure on Android 13/14 (e.g. `createSession` API change) would kill the Shizuku process with no feedback to the app.

I'm not opposed to adding a `try`/`catch` there, but there haven't been any changes to the `createSession` API in Android 13/14 or later. The last change was in Android 12 (version code S), which is already accounted for.
<br>
> * `ShizukuInstaller.shizukuDeadListener`: calls `continueQueue(InstallStep.Error)` before `service.stopSelf()`.

`continueQueue()` immediately marks the next queued item as installing unless `ready` is set to false beforehand.
<br>
> * `Installer.onDestroy()`: uses `waitingInstall.exchange(null)` to atomically retrieve and report the active entry's error, instead of silently nulling it.

This makes the earlier change redundant anyways since `service.stopSelf()` eventually leads to `Installer.onDestroy()`.
<br>

p.s. This still doesn't address the issue of failure not being shown to the user and just being silently logged. A couple of toasts should be sufficient in this case.

EnderAvaritia:
> > 1. **ShellInterface.kt**: `install()` had no try/catch. A reflection failure on Android 13/14 (e.g. `createSession` API change) would kill the Shizuku process with no feedback to the app.
> 
> I'm not opposed to adding a `try`/`catch` there, but there haven't been any changes to the `createSession` API in Android 13/14 or later. The last change was in Android 12 (version code S), which is already accounted for.
> 
> > * `ShizukuInstaller.shizukuDeadListener`: calls `continueQueue(InstallStep.Error)` before `service.stopSelf()`.
> 
> `continueQueue()` immediately marks the next queued item as installing unless `ready` is set to false beforehand.
> 
> > * `Installer.onDestroy()`: uses `waitingInstall.exchange(null)` to atomically retrieve and report the active entry's error, instead of silently nulling it.
> 
> This makes the earlier change redundant anyways since `service.stopSelf()` eventually leads to `Installer.onDestroy()`.
> 
> p.s. This still doesn't address the issue of failure not being shown to the user and just being silently logged. A couple of toasts should be sufficient in this case.

i think it caused by continueQueue() immediately marks the next queued item as installing it block at installing ,so that nothing can be installed
i am not sure what had happened in shizuku
but after this modify , i can install plugins by shizuku
i really need someone who konws what happened to help me

PR #3426, opened 2026-06-17
Add setting for manual updates to respect search and filters

### LINKED ISSUE #3425: Have library refresh actions respect search and filters
### Steps to reproduce

1. Go to library tab
2. Search / Filter a category
3. Try swiping down to refresh, or from the dropdown menu select either "Update library" or "Update category".
4. Notice that none of these respect the search and filter.

### Expected behavior

I expect swiping down to refresh to at the minimum respect the search and filter.

I have slightly less confidence on whether "Update category" and "Update library" should respect the filter. I feel that they probably should, as there is no other way to do a custom subset.

I also later plan to file a feature request and PR to enable updating a library multi-selection.

### Actual behavior

The entire category or library refreshes instead of the subset, so search-hidden items refresh.

### Crash logs

_No response_

### Mihon version

0.19.9

### Android version

Android 9

### Device

Samsung Galaxy S8

### Other details

Image

Random entry seems to already respect this.

### Acknowledgements

LetianLi:
Closes #3425.

### Summary

Add a switch setting (Settings -> Library -> Global update -> Filter Manual Updates) defaulting to off.

When switched on, update actions (Drag down refresh, selecting "Update category", and selecting "Update library") will respect any search and filters in the library. This doesn't impact automatic updates, just the ones triggered manually.

Selecting "Open random entry" already respects search and filters, so this PR allows users to opt-in to having the other actions respect search/filters.

Two new strings have been added: "Update filtered category" and "Update filtered library". Translations left for Weblate(?).

### Technical changes

The three actions are `onRefresh()`, `onClickRefresh()`, and `onClickGlobalRefresh()`.

Some of `LibraryUpdateJob.kt`'s functions now support specific manga ids.
- Before: `addMangaToQueue()` and `startNow()` takes in only a `categoryId`, where a -1 refers to a global update, and when specified is a single category update.
- After: They now also take in an additional `mangaIds`, specified only when helper `shouldScopeUpdateToVisibleItems()` says that filter or search was enabled. If for whatever reason it's non-null but empty I fail with message "No entries found in this category". `categoryId` is still used to determine whether the update is global or category-only.

### Images
| Image | 
| ------- | 
| Mihon filter manual updates |

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
@@ -111,7 +111,8 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
         libraryPreferences.lastUpdatedTimestamp.set(Instant.now().toEpochMilli())
 
         val categoryId = inputData.getLong(KEY_CATEGORY, -1L)
-        addMangaToQueue(categoryId)
+        val mangaIds = inputData.getLongArray(KEY_MANGA_IDS)?.toSet()
+        addMangaToQueue(categoryId, mangaIds)
 
         return withIOContext {
             try {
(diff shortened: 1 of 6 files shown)


app/src/main/java/eu/kanade/tachiyomi/data/library/LibraryUpdateJob.kt
```diff
@@ -148,20 +149,24 @@ class LibraryUpdateJob(private val context: Context, workerParams: WorkerParamet
...
-        } else {
-            val includedCategories = libraryPreferences.updateCategories.get().map { it.toLong() }
-            val excludedCategories = libraryPreferences.updateCategoriesExclude.get().map { it.toLong() }
-
-            libraryManga.filter {
-                val included = includedCategories.isEmpty() || it.categories.intersect(includedCategories).isNotEmpty()
-                val excluded = it.categories.intersect(excludedCategories).isNotEmpty()
-                included && !excluded
+        val listToUpdate = when {
+            mangaIds != null -> libraryManga.filter { it.id in mangaIds }
```
LetianLi, line 158: It looks weird in the diff, but all that changed was if -> when, and the extra case for mangaIds != null

BrutuZ:
That behavior will get really confusing if the ever popular Unread filter is enabled.
Anticipating lots of "My library is no longer updating" support messages

LetianLi:
Hmm, I can see that happening.
One way is to just respect search, but I also see people trying to update hiatus only. 
Maybe a setting? I usually try to avoid that though

LetianLi:
There's another side effect rn, which is that filtering to Completed and updating doesn't actually update, because Smart update skips completed entries.

LetianLi:
One way is maybe to disambiguate is to change the toast to something like "Updating filtered library/category", but that requires new translation strings.

Let me know what the PR needs.

LetianLi:
Added strings "Updating filtered library" and "Updating filtered category". Not sure how translations would work.

Added Settings -> Library -> Global update -> Filter Manual Updates, setting. Defaulting to off, so this is opt-in right now

LetianLi:
Draft until upstream completes ScreenModel to ViewModel migration and it stabilizes.

LetianLi:
Closing since i dont like the hacky way ids were plumbed in.

PR #3436, opened 2026-06-18
color change in catppuccin color scheme

xenitane:
changed secondary and tertiary colors in catppuccin color scheme for better discernment of buttons and badges.

Now The badge for downloaded chapters and unread chapters are of different colors plus other places where secondary and tertiary colors are used when using catppuccin theme.

| Old | New |
| --- | --- |
|  |  |


AntsyLich:
update changelog file

PR #3451, opened 2026-06-21
Add source groups to browse and global search

### LINKED ISSUE #3385: Multiple Search Groups

This would probably be a stretch - but would be good to have . 
Can we have a option where we can have different extensions grouped, and search can be implemented for each one separately as required.  Essentially similar to the Pinned , but ability to have multiple 'groups' of pinned extensions. 
Sharing a sample snap, on what I think it could be - a very basic thought/view

### Other details

Image

Image

### Acknowledgements

MarcosLorCar:
Add the ability to organize sources into named groups (issue #3385), alongside the existing single pinned state.

Group membership is persisted as a `Set<String>` preference of `"group|sourceId"` entries — mirroring how pinned sources are already stored — and is kept separate from the existing `Pins` bitflags (which can't represent arbitrary named groups).

Long-press a source's pin icon to open a dialog to pin/unpin and check/uncheck its groups, create a new group, or delete a group. This dialog can also be accessed by a new entry in the source options dialog.
Grouped sources appear as their own sections (between Pinned and the language sections). Tapping the group icon within a group's section removes the source from that group; tapping it elsewhere opens the dialog.
A filter chip per group scopes a global search to that group's sources, next to the existing Pinned/All chips.

Only `base/strings.xml` is touched for new strings; translations are left to Weblate.

Closes #3385

### Images
| Sources list | New source options dialog | Groups dialog | Global search |
| - | - | - | - |
|  |  |  |  |

### Testing
- `./gradlew :app:compileDebugKotlin` — builds successfully
- `./gradlew spotlessCheck` — ktlint/formatting passes
- `./gradlew test` — unit test suite passes (no regressions)
- Manually tested the functionality on-device, as shown in the images above

### DIFF
--- a/app/src/main/java/eu/kanade/domain/DomainModule.kt
+++ b/app/src/main/java/eu/kanade/domain/DomainModule.kt
@@ -12,11 +12,14 @@ import eu.kanade.domain.manga.interactor.GetExcludedScanlators
 import eu.kanade.domain.manga.interactor.SetExcludedScanlators
 import eu.kanade.domain.manga.interactor.SetMangaViewerFlags
 import eu.kanade.domain.manga.interactor.UpdateManga
+import eu.kanade.domain.source.interactor.DeleteSourcePinGroup
 import eu.kanade.domain.source.interactor.GetEnabledSources
 import eu.kanade.domain.source.interactor.GetIncognitoState
 import eu.kanade.domain.source.interactor.GetLanguagesWithSources
+import eu.kanade.domain.source.interactor.GetSourcePinGroups
 import eu.kanade.domain.source.interactor.GetSourcesWithFavoriteCount
 import eu.kanade.domain.source.interactor.SetMigrateSorting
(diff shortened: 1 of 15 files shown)


mrissaoussama:
is this still being worked on?

MarcosLorCar:
> is this still being worked on?

I mean, I just stopped caring about it being mergeable as I saw this would just be ignored, but seeing as a new issue was open about basicallty the same thing, I could try to resolve conflicts to have the pr available for review again. I also read that you objected on something about my implementation, something about a modal, but I didn't really get what you meant, so if you have anything to add, just say so.
Edit: maybe I should just re-do the feature on top of the newly implemented groups you talk about in the other issue, instead of just fixing my current implementation.

wasu-code:
I’m curious how `Has results` behaves in that PR - if it ends up pushed outside the viewport, that would be a problem. It probably should stay fixed on the right side, with only the other pills scrollable.

Issue #1559, opened 2024-12-09
Sudden closing


Ammaralkaaki:
### Steps to reproduce

When I want to install or update any extension the app sudden close

### Expected behavior

Install or update

### Actual behavior

Sudden close

### Crash logs

![8](https://github.com/user-attachments/assets/73c407c8-7a6e-414d-b48d-eb4c52ae8578)
![7](https://github.com/user-attachments/assets/85b215cc-ed3e-4dec-ac85-c5a70a160e6a)
![6](https://github.com/user-attachments/assets/1102cd6d-80b2-4508-9104-d98605cab7a3)
![5](https://github.com/user-attachments/assets/0b7eafd6-aab5-4aa9-bd5a-3cf0c646048a)
![4](https://github.com/user-attachments/assets/cdd9ca87-88d7-4096-b461-705fb9bdf66a)
![3](https://github.com/user-attachments/assets/5c45239a-be6c-414f-8858-c2dd180f880c)
![2](https://github.com/user-attachments/assets/34c75f88-a581-41e3-8a06-67c425d2c189)
![1](https://github.com/user-attachments/assets/64e302b4-e173-4d11-b0dd-cf150df03f1e)
[mihon_crash_logs.pdf](https://github.com/user-attachments/files/18068019/mihon_crash_logs.pdf)

### Mihon version

0.17.1

### Android version

13

### Device

Mi 12s ultra

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [X] I have updated the app to version **[0.17.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

MajorTanya:
We appreciate you providing the screenshots and the PDF, but if at all possible, please provide crash logs in their natural .txt format in the future. 👍

MajorTanya:
Could you go to the App Info for Mihon (the screen where you see all the info like storage use, force stop, uninstall, etc.) > Permissions > Other Permissions, and check that "Access list of installed apps" is _allowed_ for Mihon? If it is not, please grant it, then restart Mihon to see if the problem is solved.

If it is already granted OR you don't see that option, please report back.

[...]

Issue #1558, opened 2024-12-09
Having trouble reading manhwa/hua in Longstrip.


0MikeySano:
### Steps to reproduce

I updated my mihon to 0.17.1 but I can't still see any images.

### Expected behavior

This should've been fix after the patch.

### Actual behavior

It still not loading images after the patch.

### Mihon version

0.17.1

### Android version

Android 8

### Device

Vivo

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [X] I have updated the app to version **[0.17.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

Smol-Ame:
It's not fully fixed for all the old devices having the blank page issue.

For now, try adjusting the bitmap threshold setting in Mihon Settings > Advanced, lower it until it shows (if it shows).

Otherwise:
>>> ### Devices experiencing blank/black pages:
** **
For some older devices, they are unable to handle the really large/long pages sources use:
1. Go to Mihon Settings > Downloads, enable "Split tall images"
    This setting is not retroactive, redownload the chapter if it's already downloaded
2. Download the chapter & try reading it again
    Otherwise, change the reading mode to Paged (if the above doesn't work/you don't want to download)
### Easy download management
* Go to Mihon Settings > Downloads, and enable "Download ahead" paired with "Delete automatically after read"
* This makes it so you only download as you read, reducing unnecessary requests & to easily manage your storage usage

MajorTanya:
> I can confirm this bug is back on my Samsung S7 Android 8.
> Tested on Beta r7031.

Have you tried the steps lined out above? As you have a Samsung Galaxy device running Android 8, the setting mentioned may not be available for you.

doudoud82:
> 
> Have you tried the steps lined out above? As you have a Samsung Galaxy device running Android 8, the setting mentioned may not be available for you.

Yeah, the settings for adjusting the Bitmap threshold are indeed hidden from the advanced settings.

arozaqf:
Have the same probelm, the mentioned step already done but still not displayed properly

AntsyLich:
[...]

Issue #1155, opened 2024-08-25
When 


demon456blade:
### Steps to reproduce

Download the app
Restore backup
![Screenshot_20240824_155604](https://github.com/user-attachments/assets/1cf83438-2184-455c-9dfc-13d636067da2)

### Expected behavior

Read chapters should show up like this 
![Screenshot_2024-08-24-19-22-53-402](https://github.com/user-attachments/assets/b78a1989-0b51-4eb6-a0eb-a21942bcd693)
Transparent

### Actual behavior

This is what actually happens 
![Screenshot_20240824_155604](https://github.com/user-attachments/assets/4bc26625-be72-4c89-a111-3816ca9970a9)

### Mihon version

0.16.5

### Android version

Android 14

### Device

Red magic 9s pro

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [X] I have updated the app to version **[0.16.5](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

Smol-Ame:
This isn't Mihon, open whatever issue you're having in the correct fork's GitHub page.

PR #3460, opened 2026-06-22
feat: Add SyncYomi cross-device sync (port from TachiyomiSY)

VictorVow:
## Summary

This adds **SyncYomi** cross-device library sync. It is essentially a **direct port from TachiyomiSY**, where the feature has lived for a while, consolidated from jobobby04/TachiyomiSY#1005 (initial cross-device sync), #1155 (ETag syncing + improvements), #1558 (sync events), and #1559 (improved category merging). The core (`SyncManager`, the `SyncService` merge/conflict-resolution engine, `SyncYomiSyncService`, `SyncDataJob`, `SyncNotifier`, `SyncPreferences`, settings UI, and triggers) is copied as-is from TachiyomiSY's current state with only the adaptations required to fit Mihon's codebase.

## What it does

- Self-hosted [SyncYomi](https://github.com/SyncYomi/SyncYomi) backend over HTTP with ETag-based optimistic concurrency.
- Settings → Data & storage → **Sync**: service selector, host/API-key fields, a "choose what to sync" selector, and sync-trigger options.
- Triggers: sync on chapter read/open, app start/resume, manual "Sync library" from the library overflow menu, and a periodic background job.
- Category versioning (uid/version/last_modified_at) for UID-based category merging.

## Differences from the TachiyomiSY original

It's a faithful port, but a few things differ:

- **SyncYomi only** — the Google Drive backend from the original PRs is intentionally omitted, so this adds no `google-api-client`/Play Services dependencies, no bundled OAuth `client_secrets.json`, and no `GoogleDriveLoginActivity` or manifest entries.
- **Extension repos → extension stores** — adapted to Mihon's `backupExtensionStores` (TachiyomiSY uses `backupExtensionRepo`), and extension stores are now carried through the merge (keyed by `indexUrl`).
- **SY-only sync targets stripped** — saved searches, custom manga info, and EH/read-entry specifics that don't exist in Mihon were removed.
- **Backup models made `data class`** — `BackupManga`/`BackupChapter`/`BackupCategory`/`BackupExtensionStore` are now data classes, which the merger relies on for structural equality and `copy()` (TachiyomiSY's equivalents already are).
- **No new dependency for HTTP status codes** — `org.apache.http.HttpStatus` constants replaced with plain `Int` literals (304/404/412).
- **Category versioning migration** is `14.sqm` here (vs `39.sqm` upstream) and omits SY's `manga_order` column.
- **No library-update sync trigger** — current TachiyomiSY `master` already dropped it, so it's not included.

## Testing

- Compiles (`:app:compileDebugKotlin`) and passes `spotlessCheck`.
- Verified on a physical device (Pixel 8): a manual sync pushes the library to a fresh SyncYomi key and returns an ETag; a fresh reinstall pulls and merges it back. No errors in logs.

### DIFF
--- a/app/src/main/java/eu/kanade/domain/sync/SyncPreferences.kt
+++ b/app/src/main/java/eu/kanade/domain/sync/SyncPreferences.kt
@@ -1,0 +1,85 @@
+package eu.kanade.domain.sync
+
+import eu.kanade.domain.sync.models.SyncSettings
+import eu.kanade.tachiyomi.data.sync.models.SyncTriggerOptions
+import tachiyomi.core.common.preference.Preference
+import tachiyomi.core.common.preference.PreferenceStore
+import java.util.UUID
+
+class SyncPreferences(
+    private val preferenceStore: PreferenceStore,
+) {
+    val clientHost: Preference<String> = preferenceStore.getString("sync_client_host", "")
+    val clientAPIKey: Preference<String> = preferenceStore.getString("sync_client_api_key", "")
+    val lastSyncTimestamp: Preference<Long> = preferenceStore.getLong(Preference.appStateKey("last_sync_timestamp"), 0L)
+
+    val lastSyncEtag: Preference<String> = preferenceStore.getString("sync_etag", "")
+
+    val syncInterval: Preference<Int> = preferenceStore.getInt("sync_interval", 0)
+    val syncService: Preference<Int> = preferenceStore.getInt("sync_service", 0)
+
+    fun uniqueDeviceID(): String {
+        val uniqueIDPreference = preferenceStore.getString(Preference.appStateKey("unique_device_id"), "")
+
+        // Retrieve the current value of the preference
+        var uniqueID = uniqueIDPreference.get()
+        if (uniqueID.isBlank()) {
+            uniqueID = UUID.randomUUID().toString()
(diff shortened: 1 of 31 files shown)


MajorTanya:
The same sync that is still experimental and has several known problems? Not sure if we should apply that to Mihon yet.

VictorVow:
It's been around for more than 2 years now, I wouldn't call it experimental. I do agree that there may still be bugs, I was planning to address those post-port but I can look into solving any major issues in this PR also. I was just a bit concerned about making sure this PR stays reviewable.

lord-ne:
Sync is a very useful feature. It's the only thing keeping me on forks right now, and I doubt I'm alone in that. There are a few rough edges still (the main one I'm aware of is entries sometimes coming back after being deleted), but I think for an opt-in feature that can be acceptable. The sync can be improved gradually by contributors once the basic code is in this repo, having the code in the main repo instead of on forks will definitely increase the pace of development.

FFFold:
期待mihon主分支早日支持Syncyomi，这对多端用户来说非常实用，也是我一直保持使用[TachiyomiSY](https://github.com/jobobby04/TachiyomiSY/releases/latest)这个分支的主要动力

AntsyLich:
I'll later roll out my own implementation of sync

Issue #1919, opened 2025-03-25
Unable to select a folder in the Welcome page


Luminaree7:
### Steps to reproduce

1. "Select a folder"
![Image](https://github.com/user-attachments/assets/06bae652-46e6-40aa-87da-ba20efa87b8f)

2. Impossible to use nor see any folder (I've created one called Mihon)
![Image](https://github.com/user-attachments/assets/ce91b10b-7406-498d-8f5e-b487d7f434e3)

### Expected behavior

Usually I should able to see and select any folder to be used.

### Actual behavior

I am able to create a new folder but not select it.

### Mihon version

0.18.0

### Android version

Android 15

### Device

Google Pixel 6 Pro

### Other details

Actually this is an issue I also have with some other apps (Aniyomi, Poweramp). 2 days ago I completely lost access to my folders in those apps.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [x] I have updated the app to version **[0.18.0](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I have filled out all of the requested information in this form, including specific version numbers.
- [x] I understand that **Mihon does not have or fix any extensions**, and I **will not receive help** for any issues related to sources or extensions.

Luminaree7:
Hi ! 👋
Yes I thought it had to be an issue from my OS too.
I've joined the Discord, it should be easier this way. 👀

Luminaree7:
Additional info:

After restarting, I got the access during a short time before it got restricted again.

It was enough to get past the Welcome page.
Downloading is still not working though.


Also, when selecting Mihon in the Settings, I don't have the 3 dots on the top right.
Only the shortcut to open the app.

PR #3461, opened 2026-06-22
fix(onboarding): offer all-files access when the folder picker is unavailable

VictorVow:
## Problem

On some devices and Android versions the system folder picker (`ACTION_OPEN_DOCUMENT_TREE`) refuses to grant access to any directory — it shows an empty list or "Can't use this folder, to protect your privacy, choose another folder", even for a freshly created folder. Because the onboarding storage step only completes once a folder tree has been granted, and Mihon does not request all-files access, affected users are stuck on onboarding with no way to finish setup. See #1919 and the same report against TachiyomiSY (jobobby04/TachiyomiSY#1514).

## Fix

Add an opt-in **"Grant all-files access"** button to the onboarding storage step (Android 11+). Tapping it sends the user to the system all-files-access screen (`ACTION_MANAGE_APP_ALL_FILES_ACCESS_PERMISSION`, falling back to the global list on OEMs that lack the per-app screen). Once `MANAGE_EXTERNAL_STORAGE` is granted, the app points its storage at its **default folder in shared storage** (`AndroidStorageFolderProvider`, e.g. `/storage/emulated/0/Mihon`) directly via a `file://` path — no SAF grant required. `StorageManager` then creates its `downloads`/`autobackup`/`local` subdirectories there as usual, so the app is fully functional.

The system folder picker remains the recommended primary path and is unaffected; this is only an escape hatch for users the picker leaves stuck. Unlike the app-specific external directory, the shared-storage folder persists across uninstalls.

On Android 10 and below the legacy storage permission already covers the default shared-storage folder, so the button is only shown on API 30+.

## Notes

This declares `MANAGE_EXTERNAL_STORAGE` in the manifest. It is never requested silently — only via explicit user action on this onboarding step — but it is a [policy-restricted permission on Google Play](https://support.google.com/googleplay/android-developer/answer/10467955) and may require a declaration/justification for Play distribution. Flagging for maintainer awareness.

## Testing

Verified on a physical device (Pixel 8, Android 16, where the picker rejected every folder): tapping "Grant all-files access", enabling it in system settings, and returning to the app sets the storage location, enables Next, and lets onboarding complete into the library. Existing folder-picker selection still works as before.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -11,6 +11,13 @@
     <uses-permission
         android:name="android.permission.WRITE_EXTERNAL_STORAGE"
         tools:ignore="ScopedStorage" />
+    <!-- Optional escape hatch for devices where the SAF folder picker refuses to grant access
+         to any folder ("Can't use this folder, to protect your privacy"). Lets the user opt in to
+         writing the storage folder directly to shared storage. Only requested explicitly from the
+         onboarding storage step. -->
+    <uses-permission
+        android:name="android.permission.MANAGE_EXTERNAL_STORAGE"
+        tools:ignore="ScopedStorage" />
 
     <!-- For background jobs -->
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
--- a/app/src/main/java/eu/kanade/presentation/more/onboarding/StorageStep.kt
+++ b/app/src/main/java/eu/kanade/presentation/more/onboarding/StorageStep.kt
@@ -1,6 +1,12 @@
 package eu.kanade.presentation.more.onboarding
 
 import android.content.ActivityNotFoundException
+import android.content.Context
+import android.content.Intent
+import android.os.Build
+import android.os.Environment
+import android.provider.Settings
+import androidx.annotation.RequiresApi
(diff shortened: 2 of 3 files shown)


MajorTanya:
This folder isn't accessible to users though, is it?

VictorVow:
No, but I imagine not everyone cares about having their manga files accessible, especially if they were to use something like [Syncyomi](https://github.com/mihonapp/mihon/pull/3460). An alternative solution is a permission prompt for "All files access", I could switch to that.

MajorTanya:
You don't consider the autobackup and local folders important for users to be able to access? Those seem definitionally important to the app's core functioning. Local source is impossible to use with your approach and autobackups as a safety mechanism becomes obsolete if the user can't access the files. 

Plus, there are plenty of situations where users need access to their downloaded files, especially in relation to unstable source urls causing the chapter hashes to not match anymore and Mihon considering the files as gone. Or users wanting to move downloaded but source-removed chapters to the local source.

VictorVow:
I use Seedvault for backups and don't use local source so I hadn't run into these limitations personally. However I've now swapped the fix to instead offer the "All files access" permission to cover these usecases.

PR #3474, opened 2026-06-26
Add chapter count delta badge to migration search

### LINKED ISSUE #3471: Feature Proposal: Chapter Count Delta Badge in Migration Screen
### Overview

This proposes adding an optional visual badge in the migration screen that shows the difference in chapter counts between the currently selected manga and each migration candidate source.

The badge format:
- `+N` → candidate has more chapters
- `-N` → candidate has fewer chapters
- `=` → equal chapter count

This is intended purely as a UX improvement to help users compare sources faster during migration.

---

### Motivation

Currently, users must manually open each migration candidate to compare chapter counts.

This feature provides:
- Faster visual comparison between sources
- Better migration decision UX
- No changes to core migration logic or data models

---

### Proposed Approach (Performance-Safe)

To avoid performance issues, the implementation should follow these constraints:

- Chapter counts must be loaded lazily per item
- Results must be cached per `(sourceId, mangaId)` pair
- No preloading in `ViewModel` init block
- No blocking calls during UI composition
- UI only renders already computed state

---

### Data Source

Chapter counts are already available via existing repository logic:

- `getChaptersByMangaId.await(mangaId)`

No new API or database schema changes are required.

---

### Expected Behavior

- Migration list loads normally
- Chapter delta is computed asynchronously per item
- UI updates when data becomes available
- Cached results prevent duplicate computation

---

### Performance Considerations

- No additional network requests beyond existing data access patterns
- Computation is bounded to visible/listed items
- In-memory caching avoids repeated work
- Async loading is non-blocking and incremental

---

### Risks

- Slight increase in asynchronous work per migration result item
- Requires careful handling to avoid recomposition or duplicate loading
- Must ensure UI does not trigger repeated loads

---

### Scope

This is strictly a UI enhancement and does not modify:
- migration logic
- source resolution
- database schema
- extension API behavior

---

### Status

This is a **proposal / proof-of-concept idea** intended for discussion before implementation.

### Other details

_No response_

### Acknowledgements

Cavinee:
## Summary

Adds an opt-in badge on each migration search result showing how its chapter count compares to the entry being migrated from  `+N` (more chapters), `-N` (fewer), or `=` (equal) so you can tell at a glance which migration target is the most complete before migrating.

## Motivation / context

Closes #3471

When migrating an entry, the search results from other sources give no indication of how many chapters each candidate has relative to your current entry, you have to open each one to find out. This surfaces that information directly on the result covers.

The issue floated a DB-only comparison, but migration candidates are search results from sources you don't follow and have **0 chapters in the local DB until opened**, so a DB count would be meaningless for them. This fetches both sides live instead (see Implementation notes).

## Areas touched

- [x] `app/`  migration search UI + screen model, Browse settings, source preference
- [x] `presentation-core/`  new `positive` / `onPositive` theme colors
- [x] `i18n/`  two new strings (toggle title + summary)
- [ ] `.github/` (workflows, actions)

## Scope & constraints

**In scope**
- New **Settings → Browse → "Show chapter count difference during migration"** toggle (default **off**); subtitle warns it loads chapters for each result, using extra network data.
- Badge on each migration search result cover showing its chapter-count delta vs. the entry being migrated from.
- Counts for both the from-entry and each candidate are fetched live from their respective sources.
- Unit tests for the pure delta / label / guard helpers.

**Explicitly out of scope / not done here**
- In-flight fetches are not cancelled when a card scrolls off screen (the migration result set is small and bounded).
- `requestedIds` is not cleared on failure, so a failed fetch is not retried within the same session.
- No badge in regular global search, the feature is migration-only; the shared composable defaults the delta to `null` everywhere else.

## Implementation notes

- Both the from-entry and each candidate have their chapter counts fetched live from their own source (`source.getMangaUpdate(..., fetchChapters = true)`), so the comparison is apples-to-apples.
- The from-count is fetched once and memoized behind a mutex; per-item candidate fetches are bounded by a `Semaphore(5)` and triggered lazily as cards appear (not preloaded in `init`). The UI renders only already-computed values.
- A new `positive` / `onPositive` `ColorScheme` extension backs the "more chapters" badge (green), reusing `error` for fewer and `secondary` for equal.
- No regression test for the `loadChapterCountDelta` orchestration itself. It depends on the Injekt service locator, Voyager's `screenModelScope`, and `Dispatchers.IO`, with no clean seam. Only the pure helpers (`chapterCountDelta`, `chapterCountDeltaLabel`, `shouldRequestChapterCountDelta`) are unit-tested.

## Testing & verification

- [x] `./gradlew spotlessCheck`
- [x] `./gradlew :app:compileDebugKotlin`
- [x] `./gradlew :app:testDebugUnitTest --tests "*MigrateSearchScreenModelTest" --tests "*GlobalSearchCardRowTest"` (13 tests pass)
- [x] Manual on-device test (debug build): toggled the setting in Settings → Browse and confirmed badges render with the correct deltas in migration search

## Risk & rollout

- Opt-in and **default off**, no behavior change unless the user enables it.
- When enabled, adds one extra (memoized) network fetch for the from-entry plus one per visible candidate, bounded by a semaphore. No DB migration, no breaking changes.

## Screenshots

| Settings toggle | Badges in migration search |
| --------------- | -------------------------- |
|  |  |


BrutuZ:
Now instead of a single call for the search sources will get spammed with that AND 1+ for every entry they return to populate the chapter list, even the obviously unrelated ones 🤔

Cavinee:
Fair point. When it's on, each source does get hammered with extra calls. What if the badge only fetched/revealed when you tap the destination you're migrating to? That way the unrelated results never get touched

PR #1032, opened 2024-07-22
feat: Komga page-based sync

RandomNamer:
As mentioned in issue https://github.com/mihonapp/mihon/issues/236
Use book APIs to achieve chapter-level read progress sync for Komga with the Komga tracker. Every page turn would be reported to Komga just like the Komga web app; to sync page progress  from Komga, just trigger refresh tracker manually and all read progress, with page-level granularity would be merged with local. Other trackers should also be able to reuse this feature by implementing `PageTracker` interface. Additionally, there are 3 sync strategies implemented but currently no UI for configuring them.

Screenshot_20240721_220342_Mihon

<details>
<summary>In case the CI is not running</summary>
[app-dev-arm64-v8a-preview.apk.zip](https://github.com/user-attachments/files/16327306/app-dev-arm64-v8a-preview.apk.zip)
</details>

Issue #236, opened 2024-01-25
Sync read page progress with trackers


kanjieater:
### Describe your suggested feature

Currently the behavior of only syncing when a chapter is read is limiting. If mihon could sync per page, you could read something on your phone, then pick up where you left off on a desktop or other device.

There has been interest voiced for this with Komga, suwayomi, to name a few.

Original issue was here https://github.com/tachiyomiorg/tachiyomi/issues/5978

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.16.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

barrelltitor:
This would be a great addition. Happens quite often I might be reading a comic on one device, then I want to finish it on another device

RandomNamer:
I don't think trackers are meant for syncing read progress across devices at such granularity. That said, I tried to implement a workable version of [page-based Komga tracker](https://github.com/RandomNamer/mihon/commit/0a841188be939d4fddb6c3d8b713bba80f3f680c) for myself. It's not a elegant solution though: to sync you have to click tracker icon, wait for a few seconds before close the dialog for the request to be done and page progress writen to the DB.

RandomNamer:
[app-dev-arm64-v8a-preview.apk.zip](https://github.com/user-attachments/files/16279777/app-dev-arm64-v8a-preview.apk.zip)
> making a reverse proxy for it to high jack on syncs

For Komga it would be easier to just send read-progress requests in app to achieve sync. In this implementation, you can see the page progress get updated in Komga web app each time you turned page in Mihon, cuz it uses a event stream to get updates from server in real time. However to sync server progress to Mihon you need to manually trigger tracker refreshing.

kanjieater:
It's working great so far. Thank you so much for sharing.

garret:
I just installed your Mihon apk but to me the tracking does not work. Should I do anything in particular? I am reading a manga from Komga. when I try to enable "Tracking" I have the "Update progress after reading" enabled as well as the Komga 'enhanced tracker'. However, it seems no page progress is synced to Komga :/

garret:
Update: I managed to get the syncing but it seems only one way: only mihon sends page reading progress to komga but if I advance reading on komga the progress is not synced back to mihon. Is this the supposed behavior?

[...]

PR #3481, opened 2026-06-27
feat: add Komga page-based sync

MathiasDPX:
This PR is an updated version of RandomNamer's PR (cf #1032) which itself mention issue #236

However, this isn't a 1:1 copy, I updated the code because it is 988 behind main, I removed the "Enable detailed tracking" option to enable it by default. And added an option to switch between sync strategies

> Use book APIs to achieve chapter-level read progress sync for Komga with the Komga tracker. Every page turn would be reported to Komga just like the Komga web app; to sync page progress from Komga, just trigger refresh tracker manually and all read progress, with page-level granularity would be merged with local. Other trackers should also be able to reuse this feature by implementing `PageTracker` interface. Additionally, there are 3 sync strategies implemented but currently no UI to configure them

### Keep furthest progress (DEFAULT)

- Local ahead -> updates the tracker
- Tracker ahead -> updates local progress
- Same progress -> no change

###  Use tracker progress (ACCEPT_ALL)

- If both are completed, or the page number is the same -> no change
- Otherwise -> updates local progress from the tracker

### Allow rereading progress (ALLOW_REREAD)

This is the default behavior
- If the chapter is already marked read locally, but the tracker shows an in-progress reread at page > 1 -> updates local progress from the tracker
- Otherwise -> behaves like Keep furthest progress

## Screenshots

| Settings screen | Sync strategy dialog |
| ------- | ------- |
| settings | dialog
 |


tlneondo:
Very Very interested in this.

kanjieater:
Would there be any way to get this to be bidirectional in sync? As in if Komga has progress mihon clients update via polls or pushes to automatically sync across devices?

MathiasDPX:
It depend on the sync strategy you choose. With the default one (ALLOW_REREAD), if you start reading on a device A, it will get push to Komga and if you open Mihon on device B, the progress on Komga will be use because it is ahead

kanjieater:
> It depend on the sync strategy you choose. With the default one (ALLOW_REREAD), if you start reading on a device A, it will get push to Komga and if you open Mihon on device B, the progress on Komga will be use because it is ahead


Thank you - I gave it a try with the default settings you recommended!

I'm it's not automatically syncing. I have to manually open the tracker option for the series before reading it to "manually" trigger the work around to pull from komga for that series.

I was hoping you could read on komga, then open mihon and it would have the page progress updated without having to open the tracker for the series.

MathiasDPX:
Thanks for reporting this issue! This has been fix (https://github.com/mihonapp/mihon/pull/3481/commits/0c5eb66377706df67eff8c04f27a427302f5a7b6)

Before, SyncChapterProgressWithTrack was only called when:
- a tracker was added (AddTracks)
- a chapter was manually marked as read (markChaptersRead)
- the dialog was opened (TrackInfoDialog)
	- 	that's why you had to open the tracker options

Now the tracking is refreshed when you pull-to-refresh

kanjieater:
Nice! I do think that's a better experience! I'm still wondering:
- Can we have it update automatically without the user having to do the pull motion manually? does that get triggered by mihon automatically on some cadence? If we had some safe poll amount or event/maintenance hook we could use to update our series that would be perfect.

- Would it be possible to do it for all series when you pull on a library page instead of manually doing it for each series before you read?

My concern here is primarily if a user is reading something on Device A, but then moves to Device B, and they forget to pull, they start reading on Device B at the old offset, then that gets pushed to the server and they lose their progress. Ideally we'd have some reliable way to keep our libraries up to date eagerly in advance of the user have to interact.

MathiasDPX:
> Can we have it update automatically without the user having to do the pull motion manually?

Since Mihon doesn't automatically pull from trackers, I think I'll implement that as an option to refresh tracking when opening the library/serie

> Would it be possible to do it for all series when you pull on a library page instead of manually doing it for each series before you read?

I implemented this one as it isn't automatic (https://github.com/MathiasDPX/mihon/commit/a968d15c65add7b4b12a9b561782ae0439aa2684)

> My concern here is primarily if a user is reading something on Device A, but then moves to Device B, and they forget to pull, they start reading on Device B at the old offset, then that gets pushed to the server and they lose their progress.

I just tested and progress gets overwritten if Mihon doesn't pull the tracking before. Do you know how Mihon behave with other trackers (the non-enhanced one like MAL) ? as it might be a deliberate behavior

kanjieater:
I would welcome @AntsyLich to chime in and add direction if need be but from what I know, historically, trackers like MAL have been treated more as destinations to update than as a synchronization source, and Mihon has generally assumed a single-device workflow.

That said, I think enhanced trackers change the expectations a bit. They're used specifically for cross-device sync, so pushing local progress without first pulling the remote state can cause accidental regressions. I think if we know were using an enhanced tracker we can use the more advanced techniques.

It might be best to hook the automatic enhanced pull on app load or on series load if the data is over a minute old if we want to get this properly automated.

pgosar:
Thanks for making this PR! I tried it myself, and I believe there's a bug. Page level sync does work on local Mihon progress and pushes it to Komga, but if on another device I update what page we're on, Mihon will still overwrite it. I have the sync strategy set to use tracker progress because I'd like to keep Komga as the sole source of truth. 

1. Read pages on phone
2. see that page progress is synced to komga
3. try closing and opening Mihon, full library refresh, going to the manga and refreshing
4. see that Mihon still hasn't updated the page>
5. See on Komga that Mihon has actually overwritten the page progress back to what it was on Mihon


This is fixed if I manually remove and re-add tracking. 

I was able to fix this locally for myself with the following patch file: 

[mihon-pr-3481-tracker-refresh-fix.patch](https://github.com/user-attachments/files/30169941/mihon-pr-3481-tracker-refresh-fix.patch)

My findings were:
1. Tracker sync is skipped based on cached state or whether the library update job was successfully started
2. Individual manga refreshes invoke `RefreshTracks` and caused Komga’s bulk book response to throw `NetworkOnMainThreadException`. This was never surfaced so it made the refresh silently seem like it did nothing. Ideally we can do some better error handling here.

So this patch makes the following changes
1. Run library tracker sync independently of the library update job result
2. make RefreshTracks do its stuff on the IO dispatcher

I'm not particularly familiar with best practices with android dev, but hoping that this is useful

pgosar:
Doing an automatic sync on app load is probably not a great idea since it would hammer the network. I was testing this with Blue Lock, and it triggered 4 GET requests which yielded 840kb over the network in total. So you can imagine it may become a lot for people that have lots of manga in their library. 

I did a quick implementation of a series based sync. It syncs on the following cases
1. Open a chapter list for a manga will sync at most every minute
2. If already on the chapter list for the manga, it syncs at most every minute. 
3. manual refreshes override the 1 minute period

[mihon-pr-3481-auto-refresh-on-manga-resume.patch](https://github.com/user-attachments/files/30170451/mihon-pr-3481-auto-refresh-on-manga-resume.patch)

Feel free to grab this and use in your PR/modify yourself. I haven't tested this beyond my immediate use case of wanting to sync specifically with Komga.

kanjieater:
Thanks, I'll give yours a try as well.

If we're only refreshing advanced trackers of self-hosted servers like komga or kavita, which is the intent of the advanced trackers it seems, I think the network request for them probably is less of a concern. 

Though I think if you can refresh fast enough from the series page like you are thinking that might be the better solution. My concern is the race of the user clicking read and getting the wrong page while its still syncing with your method though, where it could overwrite progress if your series page doesnt pull from komga fast enough.

pgosar:
Yeah the race condition is real, it takes nearly a second to pull from komga even on my home network sometimes. Locally, I just forced it to wait for the pull to complete before it would let you resume the manga (clicking too fast will just do nothing, before it then opens slightly afterwards). But the patch file I sent doesn't do that.

Issue #3367, opened 2026-05-31
Incognito function for categories


jo13IT1:
### Describe your suggested feature

Add an incognito feature for categories in settings 

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.19.9](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

sanjay-rengarajs:
Hey! I just built this feature and opened a Pull Request for it here: https://github.com/mihonapp/mihon/pull/3496
It adds an eye icon to the category management screen so you can toggle incognito mode on a per-folder basis. Hopefully, the maintainers will review and merge it soon!

PR #3496, opened 2026-07-01
feat: add incognito mode toggle for specific library categories

sanjay-rengarajs:
## Summary
This PR introduces the ability to flag specific library categories as "Incognito" as requested by https://github.com/mihonapp/mihon/issues/3367

Currently, users who want to read sensitive content without polluting their public trackers (AniList/MAL) or local history must remember to toggle the global Incognito mode on and off manually. This feature automates the privacy layer at the folder level.

### 1. Domain & Data Layer (`Category.kt`)

- Added an `INCOGNITO_MASK` (`0x2L`) to the `Category` model.
- Reuses the existing `flags` field in the database, avoiding the need for a schema migration.
- The incognito state is stored and retrieved alongside existing category flags (such as update restrictions).

### 2. Presentation & UI Layer (`CategoryListItem.kt`, `CategoryScreen.kt`)

- Updated `CategoryListItem` to include a trailing icon button next to the existing edit controls.
- The button toggles the category's incognito state, switching between the Material `visibility` and `visibility_off` icons to indicate whether the category is private.
- Updated `CategoryScreenModel` to handle the toggle action and persist the updated flag to the database.

### 3. Business Logic & Reader Layer (`ReaderViewModel.kt`)

- Updated reader initialization and lifecycle handling to evaluate both the global incognito preference and category-specific flags.
- When a manga belongs to one or more categories, the reader checks whether any parent category has `INCOGNITO_MASK` enabled.
- If any category is marked as incognito, the reader temporarily enters incognito mode for that reading session by:
  - Skipping local reading history updates.
  - Preventing read count increments.
  - Disabling synchronization with third-party tracking services (e.g., MyAnimeList and AniList).
- This behavior is applied only for the active reading session and does not modify the global incognito preference.

## Testing & Review
- [x] I have performed a self-review of the code.
- [x] Tested locally on emulator/device.
- [x] Confirmed that reading a manga in a flagged category does not update local history, increment read counts, or trigger third-party tracking updates.
- [x] Confirmed UI toggle correctly saves to the database.
- [x] Checked across base light/dark themes (icons respect theme tints).

### Images
| Screenshot_20260701_142809 |


app/src/main/java/eu/kanade/presentation/category/components/CategoryListItem.kt
```diff
@@ -52,6 +55,13 @@ fun ReorderableCollectionItemScope.CategoryListItem(
                 text = category.name,
                 modifier = Modifier.weight(1f),
             )
+            IconButton(onClick = onToggleIncognito) {
+                Icon(
+                    imageVector = if (category.isIncognito) Icons.Outlined.VisibilityOff else Icons.Outlined.Visibility,
+                    contentDescription = "Toggle Incognito Mode",
```
H1ghSyst3m: Strings/texts should use i18n instead of hardcoded

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryScreenModel.kt
```diff
@@ -1,5 +1,4 @@
 package eu.kanade.tachiyomi.ui.category
-
```
H1ghSyst3m, line 2: Also why removed the separator line?

H1ghSyst3m:
I'm not a maintainer and have no right to tell what's good and what not, but I personally like this feature.

One thing I would like to ask: I might be wrong since I haven’t tested it myself, but could the incognito flag get lost when category flags are updated elsewhere, like sorting/resetting settings or backup restore?

domain/src/main/java/tachiyomi/domain/category/model/Category.kt
```diff
@@ -10,8 +10,10 @@ data class Category(
 ) : Serializable {
 
     val isSystemCategory: Boolean = id == UNCATEGORIZED_ID
+    val isIncognito: Boolean = (flags and INCOGNITO_MASK) !=0L
```
H1ghSyst3m:
```suggestion
    val isIncognito: Boolean = (flags and INCOGNITO_MASK) != 0L
```

Just some missing space

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryScreenModel.kt
```diff
@@ -44,6 +46,18 @@ class CategoryScreenModel(
...
 
+    fun toggleIncognito(category: Category) {
+        screenModelScope.launch {
+            val newFlags = if (category.isIncognito) {
+                category.flags and Category.INCOGNITO_MASK.inv()
+            } else {
+                category.flags or Category.INCOGNITO_MASK
+            }
+
+            updateCategory.await(CategoryUpdate(id = category.id, flags = newFlags))
```
H1ghSyst3m: Small thing: should `toggleIncognito()` also handle errors from `UpdateCategory.await(...)` like the other category actions do?

BrutuZ:
> Currently, users who want to read sensitive content without polluting their public trackers (AniList/MAL)

Can't see how that would ever be an issue since tracking has to be manually assigned, so just don't assign tracking to your questionable lewds. And for AL in particular you can also track it privately.

Entries can be assigned to multiple categories, what if an entry is on both a regular and incognito category?

H1ghSyst3m:
> Entries can be assigned to multiple categories, what if an entry is on both a regular and incognito category?

From what I can see, incognito would win here, since the code checks whether any assigned category is marked as incognito. So if a manga is in both a regular and an incognito category, it would be treated as incognito. 

But I can't tell for sure since I didn't test this PR personally

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryScreenModel.kt
sanjay-rengarajs: Thanks for pointing it out! I looked into matching it with the `when(...)` block like `renameCategory`, but it looks like the `UpdateCategory` interactor doesn't currently return a sealed `Result` class like the others do. 

To keep the UI behavior identical without rewriting the interactor, I wrapped it in a standard `try/catch` that emits `_events.send(CategoryEvent.InternalError)` if it fails, so the screen still handles it consistently.

sanjay-rengarajs:
> > Currently, users who want to read sensitive content without polluting their public trackers (AniList/MAL)
> 
> Can't see how that would ever be an issue since tracking has to be manually assigned, so just don't assign tracking to your questionable lewds. And for AL in particular you can also track it privately.
> 
> Entries can be assigned to multiple categories, what if an entry is on both a regular and incognito category?

You make a fair point about tracking being manual! While it acts as a nice fail-safe there. 

Regarding multiple categories, @H1ghSyst3m is exactly right! The logic in `ReaderViewModel` is "greedy." If an entry belongs to *any* parent category flagged as incognito, the entire reading session defaults to incognito. It acts as the safest privacy fallback.

app/src/main/java/eu/kanade/presentation/category/components/CategoryListItem.kt
sanjay-rengarajs: Fixed in the latest commit. Thanks

app/src/main/java/eu/kanade/tachiyomi/ui/category/CategoryScreenModel.kt
sanjay-rengarajs: Fixed in the latest commit.

domain/src/main/java/tachiyomi/domain/category/model/Category.kt
sanjay-rengarajs: Fixed in the latest commit.

sanjay-rengarajs:
> I'm not a maintainer and have no right to tell what's good and what not, but I personally like this feature.
> 
> One thing I would like to ask: I might be wrong since I haven’t tested it myself, but could the incognito flag get lost when category flags are updated elsewhere, like sorting/resetting settings or backup restore?

Since the category flags are stored natively as an integer in the database, the bitmask (including this incognito bit) is preserved seamlessly during standard backups and restores. I also audited the other category update functions, and since they use bitwise operations (like `oldFlags` | `newFlag`), the incognito state won't get accidentally wiped by other setting changes.

Issue #1, opened 2024-01-16
Availability on F-Droid


sitiom:
### Describe your suggested feature

Would be nice to add this on F-droid like [Tachiyomi](https://staging.f-droid.org/en/packages/eu.kanade.tachiyomi/) or self-hosting an F-droid repo like [Aniyomi](https://fdroid.aniyomi.org/)

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an official extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/extensions/issues/new/choose).
- [X] I have updated the app to version **[0.15.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

halshar:
i don't mind the updates being available late on fdroid but please add it and thank you :)

JFronny:
FYI: F-Droid now encourages reproducible builds, which mean that the developer's **official** binary is hosted by them, with them providing additional assurance that something didn't sneak in on the dev's side (by comparing the binary to a build they do).
IIRC, not hosting official binaries was the reason F-Droid was not supported by Tachiyomi, but with this that is now essentially fixed.

papjul:
F-Droid does [some patching](https://gitlab.com/fdroid/fdroiddata/-/blob/master/metadata/eu.kanade.tachiyomi.yml) (remove analytics and proprietary bits such as JUnrar or Google) which make binaries hosted on F-Droid differ from official ones. I'm pretty sure this would mean adding this patching to the [push workflow](https://github.com/mihonapp/mihon/blob/main/.github/workflows/build_push.yml). Basically, making 10 APK, 5 for the standard release and 5 other for a F-Droid flavor. Am I right? Would the project accept a pull request doing that?

JFronny:
F-Droid does not, and can not, patch apps using reproducible builds, that's the whole point of them.
It is, however, correct, that F-Droid only allows FOSS software, and that using proprietary dependencies would prevent adding it to the repository in the first place. I was not aware of Mihon using play services, but those would have to be removed from the binary here or replaced with FOSS alternatives to make it available on F-Droid.
At least that is what I gather from other cases of apps getting on F-Droid.

papjul:
[...]

PR #3498, opened 2026-07-01
Add fastlane metadata for F-Droid

wh201906:
This PR provides the metadata for F-Droid. See https://github.com/mihonapp/mihon/issues/1#issuecomment-4818573326 for reference.

This PR is based on the following guides:
https://f-droid.org/docs/Submitting_to_F-Droid_Quick_Start_Guide/
https://f-droid.org/en/docs/All_About_Descriptions_Graphics_and_Screenshots/

I added a translated short description to match the `Latest tab criteria`
> In F-Droid client, since release 1.6, only applications that have all of the following are shown on the Latest tab:
> Name
> Icon
> Summary
> Description
> License
> a What’s New entry for at least one release
> at least one graphic (screenshot or a feature graphic)
> at least one of the above translated

The screenshots are from https://github.com/mihonapp/website/tree/main/website/src/public/home/extra, and the icon is from https://github.com/mihonapp/mihon/blob/main/.github/assets/logo.png. I think https://github.com/mihonapp/mihon/blob/main/.idea/icon.svg or the icons in the apk are also good alternatives.


MajorTanya:
Does this mean we have to maintain this new changelog file in addition to our current CHANGELOG.md at root?

wh201906:
I'm not sure about that...I added the changelog entry because it's `a What’s New entry for at least one release`, this is not necessary for F-Droid as far as I know.
@linsui Hi can we use the the fastlane changelog entry as a reference to the existing changelog file/website?

linsui:
Yes, but I would rather drop it if the dev doesn't want to maintain it. It's not necessary anyway.

linsui:
Can you also add a full_description.txt for zh-CN? :)

wh201906:
Finished in 01bd2750d2148cc7a83a5d4cdbec8d3d497416d4. I'm not very confident in my translations, though.

linsui:
Thanks!

AntsyLich:
See #3564 (add some translation too)

PR #3499, opened 2026-07-01
Use Shikimori's GraphQL API where possible

MajorTanya:
User list mutation aren't available via GraphQL, so that will still be done with the v2 API.

Shikimori API docs say to prefer the GraphQL API when possible: https://shikimori.io/api/doc

Allows adding some additional data in the search results, namely:
- Authors & Artists
- Description

As a nice bonus, this reduces the number of requests to Shikimori because the findLibManga method no longer needs to do 2 calls to fetch both list and title data.

---

~~Blocked by #3497.~~


MajorTanya:
Wasn't sure what to put in the Changelog (and which section to use), so I pretended this was just adding authors & descriptions to the search results, which is the user-facing change here :)

AntsyLich (approved):

MajorTanya:
Rebased to resolve conflict from #3497

PR #3533, opened 2026-07-06
Show usernames for each tracker in settings

MajorTanya:
> [!NOTE]
> Currently, this requires `saveDisplayName` to be called before `saveCredentials` since `saveCredentials` triggers the `isLoggedIn` flow which in turn is the trigger for the UI update. This could definitely be improved somehow.

To differentiate tracker name from username, I have added a `FontWeight.Medium` to the tracker name at @AntsyLich's request.

Implemented for:
- [x] MyAnimeList
- [x] AniList
- [x] Kitsu
- [x] MangaUpdates
- [x] Shikimori
- [x] Bangumi
- [x] Hikka

**Not** implemented for:
- [ ] Komga
- [ ] Kavita
- [ ] Suwayomi

Because I don't have those readily available for testing out + I don't know how useful it'd be for those.

### Images
| Tracking screen |
| ------- | 
| image |

(yes, I have the same username on all of them, and yes it will actually show the appropriate names if you don't have the same names everywhere)


MajorTanya:
~~I'll try to figure out Hikka as well now that it has been merged.~~

Edit: That was phenomenally easy 🤣

PR #3548, opened 2026-07-08
Fix Hikka search failing due to unclosed response & show toast on remote-removed titles

### LINKED ISSUE #3547: Hikka search sometimes fails due to unclosed response object
### Steps to reproduce

This is _very_ intermittent and I have only seen it a few times, but each time I remember it happening was like this:
1. Open the Hikka search for a title with no Hikka tracking

### Expected behavior

Search results show up or "No results" message

### Actual behavior

Image

This error appears.

### Crash logs

_No response_

### Mihon version

(local build directly based on 41f4579c742059075d695356c5b5484546662719)

### Android version

Android 14

### Device

Xiaomi 11T

### Other details

I have included some log statements in my local in the `HikkaInterceptor.kt` because my suspicion is that one of the requests for token refresh in there isn't properly closed, leading to the exception above.

Regrettably, I didn't encounter this Heisenbug on an emulator with active logcat output, so I don't have any further leads than this.

Pinging @Lorg0n from our discussion earlier in #1386 to move discussion over here to not bother those participants with lots of emails.

### Acknowledgements

- [x] I have updated the app to version **[0.20.0](https://github.com/mihonapp/mihon/releases/latest)**.

MajorTanya:
This closes #3547.

I added an exception to be thrown when the `authTokenInfoResponse` is unsuccessful as otherwise, an exception would be thrown later on when the body of it is read by the `json.decodeFromString()` call below.

I have also included a precautionary refactor of the `getRead` method in HikkaApi.kt, which by the looks of it _could_ leak an unclosed response as well.

The try/catch with `awaitSuccess` pattern has been used in this same way in the Bangumi and MangaBaka implementations as well. The benefit here is that `awaitSuccess()` closes the response on error before throwing an `HttpException`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 ### Added
 - [Hikka](https://hikka.io/) tracker support ([@Lorg0n](https://github.com/Lorg0n)) ([#1386](https://github.com/mihonapp/mihon/pull/1386))
   - Fix Hikka not defaulting to "Plan to Read" for unread titles ([@MajorTanya](https://github.com/MajorTanya)) ([#3534](https://github.com/mihonapp/mihon/pull/3534))
+  - Fix Hikka search throwing error due to unclosed response ([@MajorTanya](https://github.com/MajorTanya)) ([#3548](https://github.com/mihonapp/mihon/pull/3548))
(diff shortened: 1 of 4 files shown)


AntsyLich (approved):

MajorTanya:
I'm actually unsure whether returning null for `getRead` is the right call, at least in `refresh`, since the user has no idea that the title isn't in their collection anymore (e.g. Shikimori will show a toast "Could not find manga").

AntsyLich:
i would say show a toast

MajorTanya:
image

PR #3554, opened 2026-07-08
Rework library search to use an AST-based approach

NGB-Was-Taken:
- Allows for more complex search queries using AND (`&&`), OR (`||`) and NOT (`-`).
    - When no operator is specified, queries can be separated with a whitespace character, comma or any combination of either to imply AND. 
- Search terms can be nested using parentheses `()`
- Added comparison based filtering (`>`, `>=`, `<`, `<=`, `=`) on id, add date, fetch interval, next update date, unread, read and total chapter count (`id`, `added`, `fetchinterval` (or `fi`), `nextupdate` (or `nu`), `unread`, `read`, `total`)
    - Dates use the format of `yyyy-MM-dd`
- Prefixes `title`, `author`, `artist`, `description` (or `desc`), `genre` (or `tag`), `source` (or `src`), `source_id` (or `src_id`/`srcid`/`sourceid`), `notes` (or `note`), `language` (or `lang`),  can be used to search particular fields, or be omitted to search all.
  - Exceptions for general search: `language`, `source_id`


AntsyLich:
i'll give you brownies if you can move this to the db

domain/src/main/java/tachiyomi/domain/library/repository/LibrarySearchRepository.kt
```diff
@@ -0,0 +1,7 @@
+package tachiyomi.domain.library.repository
+
+interface LibrarySearchRepository {
+    suspend fun getFilteredMangaIdsByQuery(queryPart: SqlQueryPart): Set<Long>
```
ghostbear: I would say the LibrarySearchAST is domain layer while SqlQueryPart is data layer. You could have a use case called `LibrarySearchInteractor` which takes in a string for search does the parsing of the sends the AST down to the repository, the repository implementation does it SQL thing returns it to the use case. Would also mean that toSqlQueryPart is a extension function

NGB-Was-Taken: Addressed in the next commit.

ghostbear (approved):

LetianLi:
What's the performance impact of doing SQL calls on every keypress? 

Are we sure we want to migrate away from the already in-memory List\<LibraryItem\>?

LetianLi:
I also see something like `name - arc` getting parsed as negating arc. I expect `name -arc` to negate properly, but not sure if it still should with the whitespace.

NGB-Was-Taken:
> What's the performance impact of doing SQL calls on every keypress?
> 
> Are we sure we want to migrate away from the already in-memory List<LibraryItem>?

I did some rough benchmarks with both SQL code and in-memory filtering on a 1.8k sized library, and in-memory search performs significantly better in terms of execution speed. So, the SQL implementation will be reverted.

> I also see something like `name - arc` getting parsed as negating arc. I expect `name -arc` to negate properly, but not sure if it still should with the whitespace.

This was a slight oversight on my end, will be fixed in an upcoming commit.

NGB-Was-Taken:
@AntsyLich This should be good to merge. It would be nice to get a preview build out to see if end users notice any slowdowns or regressions, and then we can go from there.

domain/src/main/java/tachiyomi/domain/library/model/search/LibrarySearchAst.kt
```diff
@@ -0,0 +1,92 @@
+package tachiyomi.domain.library.model.search
```
AntsyLich:
```suggestion
package mihon.domain.library.model.search
```

app/src/main/java/eu/kanade/presentation/library/QueryNodeExtensions.kt
```diff
@@ -0,0 +1,146 @@
+package eu.kanade.presentation.library
```
AntsyLich:
```suggestion
package mihon.feature.library
```

AntsyLich (approved):

LetianLi:
Should escaping be added or nah? very little benefit honestly

NGB-Was-Taken:
> Should escaping be added or nah? very little benefit honestly

You can already surround your query with either `"` or `'` to treat it as a raw string. You can even wrap `"` around `'` or `'` around `"` to include them in a query.

AntsyLich (approved):

AntsyLich:
resolve conflict thanks

mujina3:
This looks to have broken tapping the source name under a series title, for anyone with more than one source language enabled. On a build from main it filters the library to nothing; before this change it filtered to that source.

The label is `Source.getNameForMangaInfo()`, which appends a language tag when several languages are enabled, so it reads "Foo (JA)", and that whole string is what gets searched for. Previously `LibraryItem.matches` used `getNameForMangaInfo()` as well, so the two agreed. The rework indexes `sourceManager.getOrStub(...).name.lowercase()`, so the decorated query no longer matches.

Passing the undecorated name from `MangaInfoHeader` seems the right way round, since a typed `source:` query should match the plain name rather than requiring the tag.

PR #3561, opened 2026-07-09
Enable parallel sync for Gradle 9.4+

MajorTanya:
Today, my freshly updated Android Studio Quail 2 2026.1.2 RC 2 showed me this:

image

Enabling that seems to be a non-brainer to me, and clicking the button added this line to the `gradle.properties` file.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,3 +7,6 @@ org.gradle.caching=true
 org.gradle.configureondemand=true
 org.gradle.jvmargs=-Xmx4g -Dfile.encoding=UTF-8
 org.gradle.parallel=true
+
+# Enabled parallel sync for Gradle 9.4+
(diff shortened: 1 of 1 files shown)


ghostbear:
According to the docs it should already be enabled as Mihon uses `org.gradle.parallel`

https://docs.gradle.org/current/userguide/build_environment.html

MajorTanya:
> org.gradle.tooling.parallel=(true,false) Since 9.4.0
> 
>    Allows running [Tooling Model Builders](https://docs.gradle.org/current/userguide/service_injection.html#toolingmodelbuilderregistry) in parallel, which can improve IDE Sync times. An explicit value will override org.gradle.parallel in this context.
> 
>    Default is the value of org.gradle.parallel.

Okay true. Weird that it showed me the message then.

PR #3578, opened 2026-07-11
Fix MangaBaka User-Agent using raw StringResource

MajorTanya:
Using the hardcoded app name for the time being until a better solution can be found.

Apparently we've been identifying as

```txt
StringResource(resourceId=2131886239)/v0.20.1-7812 (app.mihon.debug 217b20123) (Android) (https://github.com/mihonapp/mihon)
```


AwkwardPeak7:
How about a mihon flavoured useragent in `AppInfo` object, all trackers can reuse it

MajorTanya:
> How about a mihon flavoured useragent in `AppInfo` object, all trackers can reuse it

That'd mean putting the app name in the Build Config which would have solved this already anyway.

Plus, pretty sure MAL still blocks us based on the `app.mihon` part of the UA anyway.

MajorTanya:
A reusable UA could be considered at another point I guess.

PR #3589, opened 2026-07-13
Add category filters for Updates tab

MajorTanya:
Works just like Library Updates.

Provides a tab in the Updates tab filter dialog to allow for including & excluding of updates belonging to a title from a given category.

Excludes overpower Includes, as with Library Updates. A little helpful text informs users about this.

Demo video:

[Screen_recording_20260713_221930.webm](https://github.com/user-attachments/assets/dfbff8c5-c0d2-4851-81b9-d3be8577d84d)

Filtering happens entirely within the `updatesView` SQL so it should not be too slow.


data/src/main/sqldelight/tachiyomi/view/updatesView.sq
```diff
@@ -32,7 +32,7 @@ WHERE dateUpload > :after
 LIMIT :limit;
 
 getRecentUpdatesWithFilters:
-SELECT *
+SELECT DISTINCT *
```
AntsyLich: why distinct?

data/src/main/sqldelight/tachiyomi/view/updatesView.sq
```diff
@@ -46,6 +46,33 @@ AND (:bookmarked IS NULL OR bookmark = :bookmarked)
 AND (
     (excludedScanlator IS NULL OR :hideExcludedScanlators = 0)
 )
+AND (
+    -- includeEmpty being true expresses "don't care" state and bypasses the membership filter
+    :includeEmpty = 1
```
AntsyLich: is it possible to make `includedCategories` nullable and check if it's not null? same for exclude

MajorTanya: For posterity: No. SQLDelight would see `:includedCategories IS NULL` and then infer the param to be `List<Long?>` as opposed to `List<Long>?` which is what we'd want as a replacement (e.g. via `includedCategories.takeUnless { it.isEmpty() }`.

See: https://github.com/lysine-dev/sqldelight/discussions/5160

But while looking at this I figure out how to make it a proper Boolean param as opposed to the silly Long one (see aedb6cc)

data/src/main/sqldelight/tachiyomi/view/updatesView.sq
MajorTanya: I had issues of duplicate updates being returned which caused a crash with the LazyColumn keys (which are based on the update id). DISTINCT here solved that.

I will go and check if it's still required, though I do believe it was required at the time I opened the PR.

MajorTanya: Okay, apparently whatever caused it to be required has been resolved at some later point. I've removed the DISTINCT clause from the query and it still works properly.

PR #3594, opened 2026-07-14
Migrate to AndroidX ViewModel from Voyager ScreenModel

AntsyLich:


H1ghSyst3m:
Hey @AntsyLich, while testing #3500 against the latest main, the app crashes when opening the More tab. I assume it's because MoreViewModel is private, so AndroidX cannot instantiate it. 

I also got the same error when opening the migration screen with MigrationConfigScreen.ViewModel, so there may be a few more affected ViewModels.

```
java.lang.RuntimeException: Cannot create an instance of class eu.kanade.tachiyomi.ui.more.MoreViewModel
Caused by: java.lang.IllegalAccessException: java.lang.Class<eu.kanade.tachiyomi.ui.more.MoreViewModel> is not accessible from java.lang.Class<androidx.lifecycle.viewmodel.internal.JvmViewModelProviders>
```

Edit: If you need, I can send you the full crash log. I tested it on my phone and also Emulator

AntsyLich:
@H1ghSyst3m thanks for the report. There's a follow up PR I'm working on that should fix this as a byproduct

PR #3607, opened 2026-07-17
Add category filters to Upcoming calendar

### LINKED ISSUE #607: Add Category Filtering for Upcoming Page

Some users would prefer to include / exclude certain categories from the upcoming page. Add filter support.

### Other details

_No response_

### Acknowledgements

MajorTanya:
Closes #607.

What it says on the tin, filtering by in-/exclusion of categories.

Excludes overpower includes, same as in Library Updates. A little text reminds users of this.

Logic is basically the same as #3589 so why not.

### DIFF
--- a/app/src/main/java/mihon/feature/upcoming/UpcomingViewModel.kt
+++ b/app/src/main/java/mihon/feature/upcoming/UpcomingViewModel.kt
@@ -1,10 +1,16 @@
 package mihon.feature.upcoming
 
+import androidx.compose.runtime.Immutable
 import androidx.compose.ui.util.fastMap
 import androidx.compose.ui.util.fastMapIndexedNotNull
 import androidx.lifecycle.viewModelScope
 import eu.kanade.core.util.insertSeparatorsReversed
+import eu.kanade.tachiyomi.util.lang.toLocalDate
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.collectLatest
+import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
 import kotlinx.datetime.LocalDate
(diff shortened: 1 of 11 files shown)


MajorTanya:
Conflicts because I was basing the work on a purposely out-of-date main due to the issue reported over at https://github.com/mihonapp/mihon/pull/3594#issuecomment-4986452136. Will rebase later.

app/src/main/java/mihon/feature/upcoming/UpcomingViewModel.kt
```diff
@@ -1,37 +1,60 @@
...
     init {
-        viewModelScope.launch {
-            getUpcomingManga.subscribe().collectLatest {
-                mutableState.update { state ->
-                    val upcomingItems = it.toUpcomingUIModels()
-                    state.copy(
-                        items = upcomingItems,
-                        events = upcomingItems.toEvents(),
-                        headerIndexes = upcomingItems.getHeaderIndexes(),
+        viewModelScope.launchIO {
```
MajorTanya, line 41: One notable thing I'm unsure about: I changed this from `launch` to `launchIO` because I saw during my work on #3589 that `UpdatesViewModel.kt` uses that one. It makes sense in my head to use IO here since it goes to the DB but I'm entirely open to suggestions/corrections on this.

data/src/main/sqldelight/tachiyomi/data/mangas.sq
```diff
@@ -164,6 +164,33 @@ FROM mangas
 WHERE next_update >= :startOfDay
 AND favorite = 1
 AND status IN :statuses
+AND (
+    -- includedEmpty being true expresses "don't care" state and bypasses the membership filter
+    :includedEmpty = 1
```
AntsyLich: is it possible to make `includedCategories` nullable and check if it's not null? same for exclude

MajorTanya: No, as explained in https://github.com/mihonapp/mihon/pull/3589#discussion_r3660594878

But I made those params true Booleans now at least.

PR #3608, opened 2026-07-17
Migrate to Metro DI

AntsyLich:


app/src/main/java/eu/kanade/tachiyomi/data/download/model/Download.kt
```diff
@@ -71,19 +66,4 @@ data class Download(
...
-            getManga: GetManga = Injekt.get(),
-            sourceManager: SourceManager = Injekt.get(),
-        ): Download? {
-            val chapter = getChapter.await(chapterId) ?: return null
-            val manga = getManga.await(chapter.mangaId) ?: return null
-            val source = sourceManager.get(manga.source) as? HttpSource ?: return null
-
-            return Download(source, manga, chapter)
-        }
-    }
```
MajorTanya, line 88: Is this not needed anymore or why is it removed?

AntsyLich: moved to DownloadManager

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
```diff
@@ -22,7 +24,7 @@ val Manga.readerOrientation: Long
 
 val Manga.downloadedFilter: TriState
     get() {
-        if (Injekt.get<BasePreferences>().downloadedOnly.get()) return TriState.ENABLED_IS
+        if (Injekt.get<Context>().appGraph.basePreferences.downloadedOnly.get()) return TriState.ENABLED_IS
```
MajorTanya, line 27: So we're not getting rid of Injekt with this? (aside from still having to provide Injekt things for extensions)

app/src/main/java/eu/kanade/domain/track/service/DelayedTrackingUpdateJob.kt
```diff
@@ -8,30 +8,36 @@ import androidx.work.ExistingWorkPolicy
...
-import uy.kohesive.injekt.Injekt
-import uy.kohesive.injekt.api.get
 import java.util.concurrent.TimeUnit
 
 class DelayedTrackingUpdateJob(private val context: Context, workerParams: WorkerParameters) :
     CoroutineWorker(context, workerParams) {
 
+    private val graph: AppGraph = context.metroGraph()
+
+    @Inject lateinit var getTracks: GetTracks
```
MajorTanya, line 28: Can't say I'm super happy about all the `lateinit var`s we're creating with this new approach...

core/common/src/main/kotlin/mihon/core/common/FeatureFlags.kt
```diff
@@ -1,12 +1,22 @@
 package mihon.core.common
 
+import kotlin.properties.Delegates
 import kotlin.uuid.ExperimentalUuidApi
 import kotlin.uuid.Uuid
 
 object FeatureFlags {
 
+    var verboseLoggingDefault by Delegates.notNull<Boolean>()
+        private set
```
MajorTanya: Can't say I understand why Verbose Logging is a Feature Flag now?

source-api/src/main/kotlin/eu/kanade/tachiyomi/source/model/Filter.kt
MajorTanya, line 1: Am I reading this right as "we got rid of commonMain entirely"? Why? Is it a necessity for Metro to work or something?

source-api/src/main/AndroidManifest.xml
MajorTanya, line 1: A manifest for the source-api?

app/src/main/java/eu/kanade/domain/manga/model/Manga.kt
AntsyLich: It'll require more rework and the pr is already bloated as is.

PR #3609, opened 2026-07-17
Fix crash caused by inaccessible view model classes during creation

AntsyLich:


MajorTanya:
```diff
--- a/app/src/main/java/mihon/feature/migration/config/MigrationConfigScreen.kt	(revision c913bdf8af82156cf6229b72ddfd31d1fee6c966)
+++ b/app/src/main/java/mihon/feature/migration/config/MigrationConfigScreen.kt	(date 1784326785812)
@@ -306,7 +306,7 @@
         }
     }
 
-    private class ViewModel(
+    class ViewModel(
         val sourcePreferences: SourcePreferences = Injekt.get(),
         private val sourceManager: SourceManager = Injekt.get(),
     ) : StateViewModel<ViewModel.State>(State()) {
```

Missed this one in MigrationConfigScreen (causes a crash when clicking the Migrate button)

Edit: Might also be worth renaming it since it's literally just `ViewModel` and I foresee some issues with that with #3608 or also just discoverability.

PR #3616, opened 2026-07-20
Fix incorrect progress when resuming downloads

xMohnad:
When resuming a partially downloaded file, the progress calculation was incorrect.

This happened because:
- `contentLength()` returns only the remaining bytes for HTTP 206 responses, not the total file size. As a result, `totalBytesRead` (which already includes `existingSize`) was being divided by only the remaining size.
- `existingSize` was always added to the total, even in cases where the server ignored the Range request and returned a full `200 OK` response instead of `206 Partial Content`.


AntsyLich:
changelog. append as a sub entry to the original

Issue #646, opened 2024-04-10
Manga recommendations


Cjbeconfused:
### Describe your suggested feature

Give manga suggestions like the one you are reading just like on the tachiyomisy app

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have updated the app to version **[0.16.5](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I will fill out all of the requested information in this form.

ishanpwork:
We can have a recommendations tab in the app that populates titles using an AI model like gemini that allows for free usage. Over time the database can be improved for better recommendations. I would advise against actively searching for titles that are recommended on websites via this features to avoid hitting rate limits.

Kom1ch1:
I'm preparing a conservative, source-scoped implementation for this request and would like early feedback on the boundaries before asking for full review.

Proposed scope:

- Two optional rows on manga details: **More by this creator** and **Similar manga**.
- Results always belong to the currently open source; navigation never crosses `sourceId`.
- No AI and no reading-history or profile upload.
- Local initialized entries are the zero-network candidate pool.
- Extra source searches are disabled by default per source. If enabled, requests are serial, at least one second apart, and limited to four source requests per page. Only the first page is used; there is no popular/latest padding.
- AniList community recommendations are queried only when the manga already has an AniList track `remoteId`; titles are not sent without that existing link.
- No source/site-specific implementation and no public `Source` API changes.
- HTTP 429 stops the current recommendation round, preserves partial results, honours `Retry-After` or bounded in-memory backoff, and never delays manga details, chapters, browsing, or reading.

The implementation is being based on the current `MangaViewModel` architecture and split into small internal components with tests for source isolation, multilingual metadata normalization, request budgets, cancellation, and phone/tablet UI.

I'll link a Draft PR here once the checks and non-sensitive screenshots are ready. This supersedes my self-closed #3595.

Kom1ch1:
Draft implementation: #3626

The clean branch is now available for architectural and scope feedback. It is based on the current main branch and supersedes #3595. Screenshots will be added before marking the Draft PR ready for review.

PR #3626, opened 2026-07-22
Add source-scoped manga recommendations

Kom1ch1:
## Summary

Related to #646.

This adds two optional, source-scoped rows to manga details:

- **More by this creator**
- **Similar manga**

## Design boundaries

- Recommendations always resolve through the currently open `sourceId`.
- There is no AI model and no reading-history or profile upload.
- Initialized manga from the current source form the zero-network candidate pool.
- Extra source requests are disabled by default and can be enabled per installed online source.
- There is no source/site-specific implementation, public `Source` API change, database migration, popular/latest padding, or second-page request.
- Existing AniList community recommendations are queried only when the manga already has an AniList track `remoteId`; an untracked title is never sent to AniList.

## Recommendation pipeline

1. Normalize creator, tag, title, and URL metadata with Unicode NFKC and locale-independent casing.
2. Generate same-source candidates from initialized local entries.
3. Optionally add one conservative creator route and one similar route, or map at most two AniList recommendation titles.
4. Verify candidate identity and metadata, score target-tag coverage and weighted Jaccard, combine route ranks with RRF, and use MMR plus quality-weighted sampling for diversity.
5. Exclude the current manga, recent recommendation-navigation ancestors, aliases, and duplicates between both rows.

Rows are published progressively. Cards already displayed are not replaced or removed during the same load, and returning from a recommended manga reuses the existing page snapshot until a manual refresh.

## Request and rate-limit behavior

- Network recommendations are opt-in per source.
- At most four source requests per manga page.
- One request at a time, with at least one second between starts.
- At most one creator search, one similar route, and two shared detail verifications.
- HTTP 429 stops the current source's recommendation round without retrying the page.
- `Retry-After` is honored; otherwise an in-memory bounded `15s -> 30s -> 60s -> 120s -> 300s` backoff is used.
- The scheduler wraps only recommendation calls. Manga details, chapters, browsing, covers, and reading are unaffected.

## UI

- Empty rows are omitted entirely.
- Phone and two-panel layouts place the rows immediately before the chapter header.
- Cards use the existing 96 dp cover style, placeholder, favorite badge, and horizontal lazy list behavior.
- Clicking a card validates the current source and persists it through `NetworkToLocalManga` before opening `MangaScreen(localId, true)`.

## Verification

- [x] `git diff --check`
- [x] `spotlessCheck`
- [x] `testDebugUnitTest`
- [x] `verifySqlDelightMigration`
- [x] `:app:lintDebug`
- [x] `:app:assembleRelease -Pinclude-telemetry -Penable-updater`

The root `lintDebug` aggregate is currently blocked on unmodified upstream `presentation-core/src/main/res/values-night/colors.xml`: `splash` has no default declaration in the base values folder. The recommendation branch has no diff in either color resource, and the app-level lint task passes.

Tests cover source-isolated identity, URL query identities, aliases and volumes, multilingual metadata normalization, creator matching, 0/1/4/many tags, coverage/Jaccard/RRF/MMR, deterministic sampling, exposure expiry, network-disabled behavior, source request and detail budgets, progressive append behavior, cancellation, 429 handling, Retry-After parsing, and AniList exact-ID mapping.

### Published fork phone preview

| More by creator | Similar manga |
| --- | --- |
| Phone manga details showing a creator recommendation row | Phone manga details showing a similar manga recommendation row |

> These screenshots are from the published fork build. The Draft PR uses the heading **More by this creator**; the first published screenshot still shows the earlier wording **More by the same creators**.

Still required before marking the Draft PR ready for review:

- English tablet/two-panel layout using non-sensitive manga


AntsyLich:
So how does this work?

Kom1ch1:
> So how does this work?

It uses author and tag metadata to find related manga from the same source. It prefers local data, while optional network recommendations use the source’s existing search API with a strict request limit.

Results are scored, diversified, and de-duplicated. They always open with the current `sourceId`, no reading history is uploaded, and unreliable rows are simply hidden.

PR #3633, opened 2026-07-23
Show clock and battery in the fullscreen reader

eMithun:
## Why

Reading in fullscreen hides the system status bar, so there's currently no way to check the time or battery level without leaving the reader. This adds a lightweight overlay — similar to the notification panel status bar — that shows the time and battery while reading in fullscreen.

## What's included

- Battery indicator changes color with charge level: green (normal), yellow (under 40%), red and blinking (under 10%)
- Shows a charging bolt when plugged in, and a small battery-saver badge when power saving mode is active
- Clock and battery text have a thin (0.5dp) hairline outline in the inverse of the text color, so they stay legible over any page background — white, black, or full-color art
- The overlay hides automatically whenever the real system status bar becomes visible, including the transient peek from swiping down from the top edge, so it never overlaps the notification shade
- "Show clock" and "Show battery" are separate, independently toggleable preferences (default on), available in both Settings > Reader > Display and the in-reader quick settings sheet

## Test plan

- [x] Built and installed a debug build on a physical device
- [x] Verified the overlay appears only when fullscreen and the menu is hidden, and disappears when the menu or real status bar is shown
- [x] Verified the two toggles are independent: turning off "Show battery" leaves only the clock, and vice versa
- [x] Verified outline legibility against a bright, busy manga panel
- [x] Verified the overlay hides during the swipe-down status bar peek and reappears after

-Mithun Hasan

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsReaderScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsReaderScreen.kt
@@ -107,6 +107,12 @@ object SettingsReaderScreen : SearchableSettings {
                     preference = readerPreferences.showPageNumber,
                     title = stringResource(MR.strings.pref_show_page_number),
                 ),
+                Preference.PreferenceItem.SwitchPreference(
+                    preference = readerPreferences.showStatusBar,
+                    title = stringResource(MR.strings.pref_show_status_bar),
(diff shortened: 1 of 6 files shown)


BrutuZ:
> there's currently no way to check the time or battery level without leaving the reader

There is, just tap the center of the screen. Alternatively disable the full-screen mode, all it really does is hide the system drawer and the system navigation buttons or gesture indicator

PR #3651, opened 2026-07-28
Remove workarounds for patched compose bugs

NGB-Was-Taken:
Changes were tested to not break existing behavior


app/src/main/java/eu/kanade/presentation/library/components/LibraryTabs.kt
```diff
@@ -22,29 +18,22 @@ internal fun LibraryTabs(
...
-                    text = {
-                        TabText(
-                            text = category.visualName,
-                            badgeCount = getItemCountForCategory(category),
-                        )
-                    },
-                    unselectedContentColor = MaterialTheme.colorScheme.onSurface,
-                )
-            }
+    PrimaryScrollableTabRow(
```
NGB-Was-Taken, line 21: The `zIndex` modifier was removed but I did not see any changes with or without. I do not know for sure why this was there, or whether removing this breaks anything.

MajorTanya:
Are these fixed in the compose versions we use?

NGB-Was-Taken:
> Are these fixed in the compose versions we use?

Yes. We use alpba BOM `2026.07.00` which translates to `1.12.0-beta02` for most components. 

- `animateItem`: fixed in `1.11.0-beta02` (ref: https://issuetracker.google.com/issues/352584409#comment3)
- `TabRow` divider: marked as fixed around a year ago. I couldn't find an exact version but considering that we are on the alpha BOM, and that it works as expected, I'd say fixed. (ref: https://issuetracker.google.com/issues/261741384)
- Edge-to-edge dialog (the comment removal): fixed around `1.8.0-beta01` (ref: https://issuetracker.google.com/issues/246909281#comment23)
- `focusRequester`: marked as fixed in `1.7` (ref: https://issuetracker.google.com/issues/204502668#comment20)

PR #3645, opened 2026-07-25
Add double-page (side-by-side spread) reading mode

Dreyia:
# Add double-page (side-by-side spread) reading mode

Closes #28 · discussion in #3644

<sub>Note: #28 has its comments locked, so I opened #3644 to flag that I was picking this up.</sub>

## Summary

Adds a **double-page** reading mode to the paged reader: two consecutive pages of the **same chapter** are composited into a single side-by-side spread — like reading a physical book — for wide-screen and landscape reading. It is the conceptual inverse of the existing **Split wide pages** option (which cuts one wide image into two), and the two are mutually exclusive.

**Default is Automatic** — spreads in landscape, single pages in portrait — matching the behaviour of the reference implementations. Users who want the old behaviour unconditionally can set Page layout to Single page.

## Why

This has been the most persistently requested reader feature, going back to Tachiyomi. Besides #28, it's been filed five more times and closed as duplicate each time: #589, #1034, #1355, #2198, #2399. Commenters on #28 consistently describe it as the one thing keeping them on Tachiyomi J2K, where it's still marked experimental.

## What's added (all under Settings → Reading mode → pager)

- **Page layout** — Single page / Double pages / Automatic (double in landscape, single in portrait).
- **Shift one page over** — offsets the pairing by one to correct the cover / odd-page misalignment.
- **Invert double pages** — swaps which page sits left/right (on top of L2R/R2L).
- **Center margin type** — optional gutter between the two pages and/or beside a lone wide page.

All controls live in settings; there are no new reader-bar buttons.

## How it works

- **Pairing** — the pager adapter derives a parallel `joinedItems: List<Pair<ReaderItem, ReaderItem?>>`; each slot holds one page (plus an optional second) or a chapter transition. `setJoinedItems()` segments pages by chapter id, applies the shift, isolates wide/full pages into their own slot, chunks into pairs, reverses for R2L, and restores the reading position after re-layout.
- **Rendering** — `PagerPageHolder` decodes both pages and composites them into one image via `ImageUtil.mergeBitmaps()`, then hands it to the existing zoomable `ReaderPageImageView`, so the spread zooms/pans as a single unit.
- **Wide pages** — detected at decode time and re-laid-out to occupy a slot alone (never squeezed into half a spread).
- **Zoom** — merged spreads fit the screen (they intentionally don't force landscape-zoom, so both pages are visible); single wide pages still honour the Landscape zoom setting.
[...]

PR #3657, opened 2026-07-29
Show two pages as a spread in the paged reader

mujina3:
This lets the paged reader show two pages at once, so facing pages line up as they were published.

A detector works out which page each chapter should start pairing from: whether the first page stands alone, or the first two pages are a pair. It's usually right for manga-like pages, and sometimes helps with others too. When it's uncertain, your default applies. A reader bar button lets you choose manually in each chapter.

### Images

| | |
| --- | --- |
| A facing pair assembled automatically into one spread | A mispaired chapter, and the control that corrects it |
| A facing pair assembled automatically into one spread. | Where it gets one wrong, correcting it takes one press. |

I realise I'm late to this party; the clash with #3645 only became apparent when I was far along in development. If mine is not the Pokémon that gets chosen then I'll be a sad badger, but the detector could sit on top of #3645 instead, and I'd rather it landed there than not at all.

The detector lives in a small library I wrote, `io.github.mujina3:spreadfit` (MIT, 631 lines of Kotlin, no transitive dependencies). Keeping it separate means Mihon doesn't have to carry the tests and tooling, but it's small enough to vendor instead if you'd prefer. Say the word and I will.

Depends on #3656, an otherwise unrelated fix; both touch `PagerViewer`.

This was built with AI assistance. Claude was my co-pilot, but I was flying the plane.

I'm on the Discord if you'd like to chat.

Thanks


mujina3:
## 1. The reader

Commits: `Allow settings checkbox items to be shown disabled` through `Add the spread settings UI and the shift toggle button`.

The first four commits deliberately add no behaviour you can reach from the UI. Pairing and geometry are pure logic with no Android dependencies, so they're unit tested directly, and the fifth commit is what turns the feature on. If you'd rather see the feature work before reading the machinery, start at the last commit and read backwards.

- **Settings and persistence** adds the viewer flags (packed into the existing `Manga.viewerFlags` bitfield) and a `spread_shift` column for remembering a manual correction per chapter. The bitfield packing is the part with a permanent cost, so `ViewerFlagMaskConsistencyTest` pins that the masks stay disjoint; adding a colliding flag later fails there rather than silently corrupting a neighbour.
- **The pairing model** is the core. `SpreadPairing.resolveShift` is the whole precedence rule in one line, and it grows by exactly one input in each of the two later parts, so it's worth reading first.
- **Joint-scale geometry** computes one scale for both halves so a spread renders as a single continuous image rather than two independently fitted pages, and defers the first reveal until that scale is known, so a spread never appears at the wrong size and then jumps.
- **Pager integration** wires those into the adapter and viewer, reusing already-decoded halves when a pairing changes so correcting the alignment doesn't re-decode the page.
- **Settings UI** adds the per-series options, the global defaults, and a button in the reader bar that turns spread mode on and corrects the pairing. The same two actions also have keyboard shortcuts, for the few who read that way.

The one change outside the reader is the first commit, which lets a settings checkbox be shown disabled.

Per-series spread settings

The per-series settings, with the global defaults behind each "Default".

A panel carried across the gutter of a facing pair

A panel carried across the gutter: two page images, placed side by side.

mujina3:
## 2. The detector

Commit: `Auto-detect the spread pairing offset with the spreadfit classifier`.

A chapter has to start pairing from the correct page. If a chapter opens with a cover, pairing from page one splits every facing spread in the rest of it, and the reader has to notice and correct it. This works that out from the pages themselves, by measuring where the margins fall.

`SpreadOffsetSampler` decodes a bounded prefix of the chapter to small grayscale rasters, and `SpreadShiftResolver` runs the classifier off the UI thread under a wall-clock budget, caching the result for the session. Two properties are worth checking:

- **It abstains rather than guessing.** A null verdict means no opinion, and the pairing falls back to whatever the reader chose. It can't force an alignment on a chapter it couldn't read, and any failure while decoding is treated the same as no signal.
- **It decides before displaying.** The pager holds a chapter's first layout until the offset settles, so a chapter is never shown at one pairing and then reflowed under the reader. That's why the hold sits in `setChaptersInternal` rather than adjusting once the pages are up. The budget is set by what a reader would notice, so the hold is over before it registers as a wait.

It only runs for local and downloaded chapters, where every page is readable up front. Streamed sources are left alone, since scanning there would either stall or reflow at an arbitrary moment.

Auto-detect is on by default, but it only acts in spread mode, which is off by default.

mujina3:
## 3. The wide page override

Commit: `Treat wide pages as spreads, anchoring the pairing around them`.

Some series contain wide pages that are themselves facing spreads, already joined into one image. Those should occupy both halves of the screen, and the pages around them should stay aligned to them.

When this is on, the scan also records each chapter's wide pages, and pairing treats a wide page as a full two-column slot that the alignment carries across. The first interior wide page fixes the chapter's alignment, so the pages after it follow from it rather than drifting.

This adds the last input to `SpreadPairing.resolveShift`, which now reads in full:

```kotlin
remembered ?: perMangaForce ?: decisiveWide ?: detected ?: default
```

A wide page outranks the classifier, because a spread that is present in the file is better evidence than a measurement. It still yields to a deliberate choice, so correcting the alignment by hand keeps working on a chapter a wide page governs.

It's off by default and per-series. This suits chapters whose embedded wide pages are themselves facing spreads, but can throw the pairing off by one where that doesn't hold.

A wide page shown across both halves

A wide page shown across both halves, with this option on. Off by default, and a separate chapter from the ones above.

AntsyLich:
Superseded by #3388

PR #3658, opened 2026-07-29
Check for app and extension update on every cold start

AntsyLich:


MajorTanya:
Does this cope well with devices that may be offline or have bad connectivity?

AntsyLich:
Hmm, I can reduce app check to 1 day instead of removing it altogether

AntsyLich:
actually this just run once after every cold boot

Issue #647, opened 2024-04-10
[Question] Why the database does not have unique constraints?


Cologler:
### Steps to reproduce

I have no idea where they come from. It might be because I'm refreshing the database while simultaneously importing data.

### Expected behavior

The database is protected by unique constraints.

### Actual behavior

A lot of duplicated chapters:

![image](https://github.com/mihonapp/mihon/assets/10906962/f61000fd-7a58-41b3-9d07-47e847f4c5bb)

### Mihon version

latest

### Android version

12

### Device

phone

### Other details

Why the database does not have unique constraints like:

- `(manga_id, url)` on `chapters` table
- `(source, url)` on `mangas` table
- `(sort)` on `categories` table
- And more

Some related issues:

- #27 
- #120
- #185

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [X] I have updated the app to version **[0.16.5](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

Cologler:
When two elements have the same key, if we query one from the database, it would be the first one; if we query all from the database and then map by `associateBy()`, it would be the last one. 

This inconsistency leads to numerous issues.

Cologler:
Even though the `chapter_id` of `history` has unique constraints, if the `(manga_id, url)` of `chapter` lacks unique constraints, then this is not unique:

https://github.com/mihonapp/mihon/blob/c94d212ef471012990bfe02ba43044de476ab9cd/app/src/main/java/eu/kanade/tachiyomi/data/backup/restore/restorers/MangaRestorer.kt#L323C13-L323C106

This issue can lead to restoration failure with:

``` txt
// <-- this is in my tachiyomi_restore.txt
ResultSet returned more than 1 row for history.sq:getHistoryByChapterUrl
```

---

*By the way, since this is a relative URL, it might be duplicated across different sources. TachiyomiSY has already tried to [fix](https://github.com/jobobby04/TachiyomiSY/commit/983a80ba42424951e81661c6f311f89eca3cbf5b) this issue by use `mangaId`:*

na-ji:
[...]

Issue #120, opened 2024-01-18
App doesn't restore entries that have errors.


THElegend5:
### Steps to reproduce

1. Restore a backup which contains errors.
2. Wait for it to finish restoring.

### Expected behavior

You get a notification saying there where errors in the restore.

### Actual behavior

You get a notification saying the backup restored successfully.

### Crash logs

```
2024-01-18 01:08:28.309] Why Are You Obsessed With Your Fake Wife? [Official] [Bato.to]: ResultSet returned more than 1 row for chapters.sq:getChapterByUrl
[2024-01-18 01:10:20.189] Diamond in the Rough [Komga]: ResultSet returned more than 1 row for chapters.sq:getChapterByUrl
[2024-01-18 01:43:13.074] Fables [Komga]: ResultSet returned more than 1 row for chapters.sq:getChapterByUrl
[2024-01-18 01:44:04.612] For Each Retweet, Two Straight Girls Who Don't Get Along Will Kiss for One Second - Their First Night [MangaDex]: ResultSet returned more than 1 row for chapters.sq:getChapterByUrl
```
You have to go app data folder to get the failed restore logs

### Mihon version

r6454

### Android version

Android 14

### Device

Samsung S21 FE

### Other details

https://cdn.discordapp.com/attachments/1099009853864812710/1197472596166455296/temp.proto.gz?ex=65bb6442&is=65a8ef42&hm=2207fbd9f20a63f6e6c8b1879dcef13173438953c9e66616a51ee74e91b00430&
Example backup file which contains errors only two entries will fail to restore

Also a thing to not is that before new restore logic these entries restored fine but you would get notification saying there were x errors in restore.
In versions with new logic there is no indication that these entries fail to restore.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [X] I have written a short but informative title.
- [X] If this is an issue with an official extension, I should be opening an issue in the [extensions repository](https://github.com/tachiyomiorg/extensions/issues/new/choose).
- [X] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [X] I have updated the app to version **[0.16.1](https://github.com/mihonapp/mihon/releases/latest)**.
- [X] I have updated all installed extensions.
- [X] I will fill out all of the requested information in this form.

MajorTanya:
[...]

PR #3667, opened 2026-07-30
Fix backup restore dropping library entries

na-ji:
Restoring a backup silently discarded most of the library. 123 entries → 23 rows, 2 in library.

Two causes:

**Unscoped chapter lookup.** `restoreHistory` resolved chapters by URL alone (`chapters.sq:getChapterByUrl`, `history.sq:getHistoryByChapterUrl`). A backup holding the same entry twice under one source matches multiple chapter rows, so `awaitAsOneOrNull()` throws `ResultSet returned more than 1 row`. Same error as #120, which was closed pointing at #647. Now scoped to the entry, with `LIMIT 1` so a duplicate can't crash restore again. This doesn't fix the duplication issue, but prevent the backup restore from crashing. 

**A failed entry rolled back 99 good ones.** Entries restore in chunks of 100 sharing one transaction (#3267). SQLDelight fails the enclosing transaction when a nested one fails, so the per-entry `try`/`catch` never contained the failure and the whole chunk rolled back. Only chunks with zero failures survived. The batched fast path is kept; on failure the chunk is retried entry by entry so only the bad entries are lost.

Verified on device, clean install, full restore of a 123-entry backup:

| | backup | before | after |
| --- | --- | --- | --- |
| entries | 123 | 23 | 123 |
| favorites | 86 | 2 | 86 |
| chapters | 21880 | 2756 | 21880 |
| history | 13773 | 1596 | 13773 |

Refs #647, but doesn't close it. The underlying issue of duplicate entries is not fixed. But since it's already in production, we need to work around it in the backup restore path. 

I can split the two changes if wanted.


AntsyLich:
update changelog

na-ji:
> update changelog

Done

PR #3676, opened 2026-08-02
Ensure R8 keeps methods called externally by Shizuku

NGB-Was-Taken:
This should be the final fix needed for shizuku installer to work in release builds (until Android's internal package installation flow changes in the future).


AntsyLich:
changelog

NGB-Was-Taken:
> changelog

done

PR #3678, opened 2026-08-02
Improve WebView spoofing

AntsyLich:
- Spoof `Sec-CH-UA` client hints to match the user agent
- Fix `X-Requested-With` spoofing leaking to unrelated callers
- Update default user agent to Chrome 149


AwkwardPeak7:
> Raise minimum supported WebView version to 138

There is a OEM issue going around where latest version of webview crashes in opengl, causing mihon to bootloop. known workaround is to downgrade webview, this may cause issues with that.

You could also fix that issue by setting webview to use software rendering on those hardware

AntsyLich:
138 was released more than a year ago. Current latest is 151.

AwkwardPeak7:
yes, but support people are telling people to uninstall webview updates via play store, which usually puts it around 118.

PR #3679, opened 2026-08-02
Disable updater for foss build

linsui:
To comply with F-Droid policy:

> Applications must not download additional executable binary files (e.g. add-ons, auto-updates, etc.) without explicit user consent. Consent means it needs to be opt-in (it must not be harder to decline than to accept or presented in a way users are likely to press accept without reading) and structured in a way that clearly explains to users that they’re choosing to bypass F-Droid’s checks if they activate it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,7 +94,7 @@ jobs:
           cache-disabled: true
 
       - name: Build
-        run: ./gradlew assembleFoss -Penable-updater
+        run: ./gradlew assembleFoss
         env:
           storeFileBase64: ${{ secrets.SIGNING_KEY }}
           storePassword: ${{ secrets.KEY_STORE_PASSWORD }}


AntsyLich:
> and structured in a way that clearly explains to users that they’re choosing to bypass F-Droid’s checks if they activate it.
>
any sample for this?

linsui:
Something like:

"Please note this update is from upstream directly and not checked by F-Droid."

And this should be opt-in so it should be disabled by default. Or add an onboarding screen for it with a note.

AntsyLich:
we don't have any force or auto update. users are prompted that an update is available every cold boot with option to download or ignore

linsui:
It 's enabled by default and can't be disabled. And it pops up on every start which is also annoying. It's best for us the have a flavor with the update check disabled completely. At lease there should be a button to ignore new versions forever.

ManInDark:
Thanks for the progress on this!

I'm still on Tachiyomi, but since it seems like the F-Droid build will soon be a thing, it'll save me the annoyance of downloading the app manually.

PR #3682, opened 2026-08-03
Respect verbose logging preference in crash log dumps

NGB-Was-Taken:

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/util/CrashLogUtil.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/util/CrashLogUtil.kt
@@ -5,6 +5,7 @@ import android.os.Build
 import eu.kanade.domain.base.BasePreferences
 import eu.kanade.tachiyomi.BuildConfig
 import eu.kanade.tachiyomi.extension.ExtensionManager
+import eu.kanade.tachiyomi.network.NetworkPreferences
 import eu.kanade.tachiyomi.util.storage.getUriCompat
 import eu.kanade.tachiyomi.util.system.WebViewUtil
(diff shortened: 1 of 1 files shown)


MajorTanya:
I don't know if going all the way to V is a good idea -- lots of users won't have Nitro to upload the files to the Discord support channel if they get very large.

NGB-Was-Taken:
> I don't know if going all the way to V is a good idea -- lots of users won't have Nitro to upload the files to the Discord support channel if they get very large.

I don't think logs can get that big. The log buffer is finite and is 256KiB for each type of buffer by default (on my device), with maximum allowed to be set (at least on my device via settings) is 8MiB, which is shared across the whole system and a single app is unlikely to use it all.
The file size shouldn't really be a concern.

MajorTanya:
> > I don't know if going all the way to V is a good idea -- lots of users won't have Nitro to upload the files to the Discord support channel if they get very large.
> 
> I don't think logs can get that big. The log buffer is finite and is 256KiB for each type of buffer by default (on my device), with maximum allowed to be set (at least on my device via settings) is 8MiB, which is shared across the whole system and a single app is unlikely to use it all.
> The file size shouldn't really be a concern.
> 
> 

Pretty sure I've personally dumped 2 MiB in logs but we'll see what happens when it does. 

I was mainly thinking that there isn't any useful information in V. D maybe but would also be a lot of useless noise as well. 
Though I guess that'd be the point of Verbose Logging.

NGB-Was-Taken:
> Pretty sure I've personally dumped 2 MiB in logs but we'll see what happens when it does.

You can verify the buffer size on your device by running `adb logcat -g`

AntsyLich:
changelog

PR #3685, opened 2026-08-03
Do not count Android SAF collision as valid pages

TheZ3ro:
A recent GrapheneOS SAF bug "corrupted" part of the content of Mihon's downloaded folder by creating duplicate files.
This PR fixes a bug in how Mihon counts chapter pages when Android SAF create duplicates, preventing similar scenarios in the future.

In particular when Mihon is downloading a chapter's page it renames the file https://github.com/mihonapp/mihon/blob/55be95dd5df7ac985bbc68ea62a5a525611a732f/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt#L494 however if the file already exists SAF creates a duplicate named `$filename (1).$extension` with increasing numbers.

Those duplicated files are counted for determining if a download succeed or not thus always triggering unsuccessful downloads when files already exists.
https://github.com/mihonapp/mihon/blob/55be95dd5df7ac985bbc68ea62a5a525611a732f/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt#L585-L596

Since `isDownloadSuccessful` is never true ^, the downloaded chapter is never promoted to successful so it keeps the `_tmp` suffix, thus re-indexing the downloads from Mihon settings never pick up this chapters.

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/download/Downloader.kt
@@ -590,6 +590,8 @@ class Downloader(
                 fileName.endsWith(".tmp") -> false
                 // Only count the first split page and not the others
                 fileName.contains("__") && !fileName.endsWith("__001.jpg") -> false
+                // Ignore SAF collision copies like "001 (1).jpg"
+                fileName.contains(" (") -> false
                 else -> true
             }
         }


AntsyLich:
it's by design. duplicate downloads should never be considered valid. if you have a alternative solution feel free to open another pr

TheZ3ro:
The thing is that the pages (not the chapters) are duplicated due to a SAF bug, and this corrupts those chapters in an irrecoverable state.

Mihon don't understand that the chapter was indeed downloaded successfully so it is not shown among the downloaded ones, but also cannot re-download them since it will just re-download the files and miscount the pages.

The only solution is to manually delete the offending chapters from a file explorer. This only after you understand what is going on since you get no error whatsoever, you simply cannot download the affected chapters.

Maybe an option to delete the offending chapters would be a better alternative?

PR #3696, opened 2026-08-04
Fix relative date display

MajorTanya:
`DatePeriod.days` provides the number of days in the `DatePeriod` specifically, not the total number. `LocalDate.daysUntil` provides what we were looking for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - `Other` - for technical stuff.
 
 ## [Unreleased]
+### Fixed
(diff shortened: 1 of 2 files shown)


WizaHX:
Can't we have back the actual dates instead this stupid rounding? 
As of now everything rounded 1 month ago gets the same date, and will shift again next day to (today-1 month), same with all the rest of the rounding.

MajorTanya:
> Can't we have back the actual dates instead this stupid rounding? 
> As of now everything rounded 1 month ago gets the same date, and will shift again next day to (today-1 month), same with all the rest of the rounding.

This is an optional setting... Disable relative dates in the settings.

PR #3732, opened 2026-08-08
Add cross-device sync through Google Drive

nkoziel:
Mihon only ships manual and automatic .tachibk backups, so moving a library between devices is a manual export/import. This adds a sync that converges devices on the same state, built on top of the existing backup pipeline rather than beside it.

A sync builds a local snapshot, downloads the remote one from the hidden application data folder of the user's Drive, merges them and uploads the result before applying it locally. The merge settles conflicts per record with the lastModifiedAt timestamps the app already maintains, so a device that has been offline cannot wipe entries added elsewhere, and un-marking a chapter as read is not undone by the other device's stale state. App and source settings are left out on purpose: they describe a device rather than a library.

The OAuth client credentials are entered in the sync settings instead of being baked in at build time, and sign-in uses a loopback redirect. A "desktop" OAuth client is not bound to a package name or a signing certificate, so one client works on every device and every build without a rebuild or a SHA-1 fingerprint.

The application ID becomes app.mihon.sync so this build installs alongside upstream Mihon. Kotlin packages are left untouched to keep rebases cheap.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -30,7 +30,8 @@ android {
     namespace = "eu.kanade.tachiyomi"
 
     defaultConfig {
-        applicationId = "app.mihon"
+        // Distinct from upstream Mihon so both apps can be installed side by side
+        applicationId = "app.mihon.sync"
 
         versionCode = 29
         versionName = "0.20.4"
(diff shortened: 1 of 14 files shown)


nkoziel:
sorry wrong repo

PR #3748, opened 2026-08-13
feat: added ability to open Extension settings within entry

Arkai1:
# Changes: 

Added onExtensionSettingsClicked
parameter to the main MangaScreen
function

Passed this parameter through to both
layout implementations
(MangaScreenSmallmpl and
MangaScreenLargelmpl)

Updated MangaToolbar.kt to include a new "Settings" action in the toolbar menu

Modified the MangaScreen.kt controller to:
Import ConfigurableSource and
SourcePreferencesScreen

Implement the settings click handler to
navigate to SourcePreferencesScreen
when the source is configurable

Added a conditional display (.takelf
(successState.source is
ConfigurableSource))

Here are the attachment 
IMG

https://github.com/user-attachments/assets/2f6b640a-5932-494b-a99d-bcc56466af33

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/manga/MangaScreen.kt
@@ -111,6 +111,7 @@ fun MangaScreen(
     onEditFetchIntervalClicked: (() -> Unit)?,
     onMigrateClicked: (() -> Unit)?,
     onEditNotesClicked: () -> Unit,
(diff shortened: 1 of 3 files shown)


MajorTanya:
What's the point of this?

Arkai1:
> What's the point of this?

Suppose You are reading a Manga and you want to change ext settings so you will have to do it via going to extensions screens then open settings but with this it is more convenient to do it.

No need to navigate away and back Especially useful when you discover mid-reading that you need to enable a feature

Issue #2898, opened 2026-01-25
Manhwa Pages Not Loading Beyond Page 1


Indra9158:
### Steps to reproduce

An issue occurred with the comic image display where the panel or page 2 does not appear as it should. When accessed, only the first page loads normally, while page 2 is blank or fails to display any image. This issue occurs only on comics with long vertical panels such as manhwa, and does not occur on manga with shorter single-panel pages.

### Expected behavior

Under normal conditions, page 2 and subsequent pages should load and display correctly after page 1, just as they do when reading manga.

### Actual behavior

When reading manhwa, pages 2 and onward often fail to load, while this issue does not occur when reading manga. This may be related to the long vertical page format used in manhwa, which differs from the shorter page format of manga.

Please advise on what should be done so that page 2 and subsequent pages can load and display properly, the same way pages appear when reading manga, even when the comic uses long vertical pages such as manhwa.

![Image](https://github.com/user-attachments/assets/1eea4107-6664-4b7c-9134-44019bfb3b74)

![Image](https://github.com/user-attachments/assets/43f99c96-6233-4fb2-aec7-9d0d32a1edc4)

### Mihon version

0.19.3

### Android version

Android 8.1.0

### Device

Oppo A5s

### Other details

I understand that my phone and Android model are outdated, but I would like to know how this issue can be fixed. I initially tested the latest Mihon version 0.19.3, but the same problem occurred. I then rolled back by installing an older version, 0.19.0, and the issue still persists. I hope this problem can be resolved.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [x] I have updated the app to version **[0.19.3](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I have filled out all of the requested information in this form, including specific version numbers.
- [x] I understand that **Mihon does not have or fix any extensions**, and I **will not receive help** for any issues related to sources or extensions.

Smol-Ame:
Searching black/blank pages here & checking the closed requests first would yield the same answer. Also, consider asking basic questions/issues in the Mihon Discord mihon-support channel next time.

>>> ### Devices experiencing blank/black pages:
__Some older devices__ are unable to handle really large/long pages:
[...]

PR #3771, opened 2026-08-15
Use tiled decoding when webtoon images cannot use hardware bitmaps

rodrigdoliveira:
## Summary

- use Mihon's existing tiled decoder for Long strip images that cannot use a hardware bitmap
- keep the current default loader for compatible Long strip images
- preserve the existing "always use legacy decoder" preference

## Problem

The default Long strip path asks Coil to prepare one complete bitmap before passing it to the reader view. A hardware bitmap stores its pixels in Android's graphics memory and is the fast path used when the image and device support it.

When `ImageUtil.canUseHardwareBitmap()` returns false, the current Long strip path still prepares the complete image as a regular bitmap. Exceptionally tall pages can make that allocation expensive or fail on device-specific decoder and memory limits.

Mihon already has a second path that gives the image stream directly to `SubsamplingScaleImageView`. That path reads smaller sections of the image as needed instead of requiring the complete decoded bitmap up front. It is currently used for paged reading and when the user enables the legacy Long strip decoder preference.

## Change

The reader now checks hardware-bitmap eligibility before selecting the Long strip decoder:

- compatible Long strip images keep the current Coil path
- images that cannot use a hardware bitmap use the existing tiled path
- the legacy preference continues to force the tiled path
- non-Long-strip reading is unchanged

## Validation

- `./gradlew spotlessCheck`
- `./gradlew :app:testDebugUnitTest`
- focused tests cover the compatible, incompatible, and legacy-preference cases
- tested on an Android 16 device with three 1800x12000 JPEG pages
- the unchanged build logged three complete-image loader results during the tall-image chapter; the patched build logged none until that chapter was closed and an ordinary 40-page chapter was opened
- 26 paired screenshots had 99.94% median visual similarity, with equivalent page dimensions and scaling
- four repeated device runs completed without blank pages, Retry screens, crashes, or memory errors

The unchanged build also completed the controlled local test because the reported failure is intermittent. The device trace demonstrates that this change selects the intended decoder without changing normal output; it does not claim that every blank Long strip page has the same cause.

Related context: #2898

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/ReaderImageDecoder.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/reader/viewer/ReaderImageDecoder.kt
@@ -1,0 +1,9 @@
+package eu.kanade.tachiyomi.ui.reader.viewer
+
+internal fun shouldUseSubsamplingDecoder(
+    isWebtoon: Boolean,
+    alwaysDecodeLongStripWithSSIV: Boolean,
+    canUseHardwareBitmap: Boolean,
+): Boolean {
+    return !isWebtoon || alwaysDecodeLongStripWithSSIV || !canUseHardwareBitmap
+}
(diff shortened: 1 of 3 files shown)


AntsyLich:
Redundant with #3786

PR #3776, opened 2026-08-15
Add `id:` prefix search for rest of third-party trackers

### LINKED ISSUE #514: Expand Tracker Search

There are times when a tracker I know has a manga, but it doesn't appear in the list of available manga when I search the tracker through mihon. I sometimes find I have to come back to an entry in my library and hope that the next time I search a tracker it pops up so I can select it.

What I'm suggesting is a "load more" feature to allow users to search a bit further. I wouldn't suggest loading up all search results at once (seems bad for both mihon and tracking service), but having the ability to search more results would be helpful.

If this is optional for users, I think that would best to keep them from spamming trackers. If there is a toggle in the settings to expand tracking services searches with a warning regarding spamming the tracker could result in consequences from the tracking sevice (if that's an issue), I believe it'd be best.

### Other details

_No response_

### Acknowledgements

MajorTanya:
This adds the `id:123` style prefix search to all remaining third-party trackers, with some important footnotes.

Trackers who had `id:` support added:
- AniList
- Bangumi
- Kitsu*
- MangaUpdates**
- Shikimori
- Hikka***

\*: While Kitsu displays slugs in the URL, you can get the integer ID from any of the poster URLs.

\*\*: ~~MangaUpdates works hard to hide its `Long` IDs from users (not ints). Neither the ID in the browser URL (e.g. `rmnzb04` taken from [Youjo Senki Restaurant](https://www.mangaupdates.com/series/rmnzb04/youjo-senki-restaurant)'s URL) nor any poster ID (which are image resource IDs) works, only the proper ID that apparently is only returned by the API.~~
\*\*: Edit: @BrutuZ pointed out that the strings (`rmnzb04`) in the series URLs are base36 encoded versions of the Long (`60143657332`) the API expects. I updated the search to accept both the pure `Long` as well as the base36 strings. On decoding error of a string, we just return an empty list.

\*\*\*: Hikka uses slugs instead of an integer-based ID system. I chose to stick with `id:` as the prefix here for consistency and discoverability, especially since most lay users wouldn't know what slug means anyway.

Closes #514 (ID prefix search was the accepted basic solution originally and the further suggestions like additional pages would usually return even less relevant results).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - `Other` - for technical stuff.
 
 ## [Unreleased]
+### Added
+- Add `id:` prefix search to remaining trackers (AniList, Bangumi, Kitsu, MangaUpdates, Shikimori, and Hikka) ([@MajorTanya](https://github.com/MajorTanya)) ([#3776](https://github.com/mihonapp/mihon/pull/3776))
+
 ### Improved
 - Show updates and upcoming filter icon as active for categories ([@Secozzi](https://github.com/Secozzi)) ([#3772](https://github.com/mihonapp/mihon/pull/3772))
 
--- a/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/data/track/anilist/Anilist.kt
@@ -28,6 +28,8 @@ class Anilist(id: Long) : BaseTracker(id, "AniList"), DeletableTracker {
         const val POINT_10_DECIMAL = "POINT_10_DECIMAL"
         const val POINT_5 = "POINT_5"
         const val POINT_3 = "POINT_3"
+
+        private const val SEARCH_ID_PREFIX = "id:"
     }
 
     private val json: Json by injectLazy()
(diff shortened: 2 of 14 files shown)


MajorTanya:
Draft because I have some open questions:

- [ ] Is ID-only fine even though they're not easily accessible everywhere?
- [ ] Can Hikka stay `id:` or should it be `slug:`

BrutuZ:
> **: MangaUpdates works hard to hide its Long IDs from users (not ints). Neither the ID in the browser URL (e.g. rmnzb04 taken from [Youjo Senki Restaurant](https://www.mangaupdates.com/series/rmnzb04/youjo-senki-restaurant)'s URL) nor any poster ID (which are image resource IDs) works, only the proper ID that apparently is only returned by the API.

You can derive the numeric ID from alphanumeric one: `"rmnzb04".toInt(36)`

> * Can Hikka stay `id:` or should it be `slug:`

Unsolicited opinion: I agree with you, keep `id:` for consistency

MajorTanya:
> You can derive the numeric ID from alphanumeric one: "rmnzb04".toInt(36)

Holy shit I didn't think of that. Time to add that in, then MU becomes much less stupid than I thought it was (insert AShawke_smile here)

AntsyLich (approved):

MajorTanya:
> Draft because I have some open questions:
> 
>     * [ ]  Is ID-only fine even though they're not easily accessible everywhere?
> 
>     * [ ]  Can Hikka stay `id:` or should it be `slug:`

After some external discussion, ID-only is acceptable for all trackers. Kitsu will be expanded in a future PR.

Hikka will stay as `id:` to be consistent.

AntsyLich:
> After some external discussion, [...]
>
*internal

PR #3792, opened 2026-08-16
Migrate to Kitsu's GraphQL API

MajorTanya:
While there are still some nasty corners (multiple shapes of error response, unhandled error types on their end, etc.), this is a functional, feature-parity replacement for the JSON:API API we were using before.

According to the GraphQL docs, searching by title still goes via Algolia.
Whether that's true, I don't know.

If you spot any weirdness that's not explained by context or comments, please do ask.


AntsyLich (approved):
if this breaks blame tanya

MajorTanya:
I've asked in the Kitsu Discord whether clients should prefer the JSON:API API or the GraphQL API but it might be a long time before we hear back from them.

Edit:
(all of these are the same person for (Developer))

- Noticed that Kitsu did an announcement in 2021 saying, quote:
  >  - [GraphQL API] We've fleshed this out and made it blazing fast. Devs can now start using it reliably [...]

- April 2024 conversation:
  > (Developer): I recommend building on the GraphQL API

- June 2025 conversation:
  > But obviously [field inclusion and normalisation] sucks ass so I personally recommend using the graphql api
  and later (same conversation, still June 2025)
  > > (User): how "unstable" is the gql api?
  > > (Developer): Not very
  > [...]
  > > (Developer): For reading data, extremely stable and unchanging
  > > (Developer): Mutations are a bit WIP and feeds are missing as we migrate to a new model off getstream

---

Overall looks like they want people to use the GraphQL API. I'm still hesitant since Mihon's impact is pretty hefty and I've heard that GraphQL is more compute-intensive than a RESTful API.

MajorTanya:
Okay so we got confirmation from the devs that they consider the GraphQL API more optimised than the JSON:API one when we asked in the Discord and didn't get any indication that we shouldn't switch, so undrafting it is.

Best of luck to us, them, and also me again. At least GraphQL schemas come with nullability info so _surely_ I did all those correctly right? /s

Oh and with #3827 all this hand-crafted artisanal assortment of dataclasses will disappear again anyway, yay

PR #3827, opened 2026-08-20
Use Apollo GraphQL for Trackers with GraphQL APIs

MajorTanya:
Apollo takes the GraphQL schema of an API through introspection and then generates classes for the types based on the schema.

You define queries and mutations in dedicated `.graphql` files which Apollo then turns into generated classes you use with the GraphQL client (see `AnilistApi.kt`). It's also got ~type safety based on the schema's reported types which is nice.

I chose the package location on a whim so that could always be changed to somewhere else. Especially since I didn't realise originally how awkward the path to the files is (e.g. `src/main/graphql/mihon/graphql/anilist`).

The diff is massive because Apollo pulls the API's schema files and stores them locally. This is a one-time addition.
(these are the `tracker.graphqls` files)
- `anilist.graphqls` is 6,319 lines
- `kitsu.graphqls` is 6,618 lines
- `shikimori.graphqls` is 1,913 lines
- `suwayomi.graphqls` is 4,883 lines
That's just the schema they provide and nothing can be done about their sizes.

Tracker-specific notes:
- Kitsu has some nonstandard responses (e.g. HTTP 500 with `error` not `errors` body for touching library IDs that don't exist) so its handling is a bit strange
- Shikimori has a GraphQL API but it offers no mutations, so list changes (add/update/remove) still go via the normal API
- Suwayomi (potentially? I'm not entirely sure) depends on a filtering fix that was pre-released in October and released in June, which feels appropriate enough to depend on

---
For anyone who finds this while wanting to adapt to schema and/or typing changes:

To download the schemas from the remote service:
(The pattern is always `download[ServiceName]ApolloSchemaFromIntrospection`)
- AniList:
```console
gradle downloadAnilistApolloSchemaFromIntrospection
```
- Kitsu:
```console
gradle downloadKitsuApolloSchemaFromIntrospection
```
- Shikimori:
```console
gradle downloadShikimoriApolloSchemaFromIntrospection
```
- Suwayomi (make sure you have a Suwayomi instance running)
```console
gradle downloadSuwayomiApolloSchemaFromIntrospection
```

To (re-)generate the generated sources for types, queries, mutaitons, etc.:
(this covers all services at once, service-specific tasks are available with `generate[ServiceName]ApolloSources`)
```console
gradle generateApolloSources
```

Additional recommendation:
If you can, install the official [Apollo GraphQL IDE Plugin](https://plugins.jetbrains.com/plugin/20645-apollo-graphql) in your Android Studio, it lets you use Go to Definition and such to get to the query files as well as the generated code, which can be super helpful. Also provides a way to dispatch queries from the IDE (no header customisation) or to open it in Apollo Sandbox (where you can manually customise headers) & it shows which attributes are unused for a given query file (like unused attributes of a class, they get greyed out)


MajorTanya:
Code level migrations are done (I believe) but I will still be testing these out for a bit more so I can be more confident that they work properly. Specifically, for each of AL, Kitsu, Shikimori, and Suwayomi I'll look at:

- [ ] Add fresh track
  - [ ] Public
  - [ ] Private
- [ ] Add remote-existing track
- [ ] Update track
  - [ ] Status
  - [ ] Progress
  - [ ] Score
  - [ ] Start date
  - [ ] Finish date
  - [ ] Private state
- [ ] Search
  - [ ] Search by ID
  - [ ] Search by Slug
- [ ] Remove track
  - [ ] Local only
  - [ ] Local + Remote
- [ ] Handling update of a remote-removed track
- [ ] Handling fetch of a remote-removed track (tracker sheet open)
- [ ] Handling of automatic progress bumps during reading

---

if I missed a specific behaviour that was touched by this (aka only API level interactions), let me know here or on Discord

MajorTanya:
From my testing, this appears to be working properly now (glad I caught d70cc67 in testing...)
I only tested with one title on each tracker, so not a great variety of them (meaning I could have missed some corner cases for strange titles).
Though with the actual schema being used as the basis for the generated code's type safety, this would be more of an upstream contract violation than it would be a typical me issue re: nullability etc. No more follow-up PRs that are just "adding a question mark and elvis here and there"[^1]

[^1]: And now I've said that, it will obviously happen

Please bring up any strange calls I may have made, it may be an actual issue rather than me being weird about writing it.

To reiterate: This diff only _looks_ intimidating. The `*.graphqls` files are created by Apollo based on the schema discovered on the remote API. As per the PR description, these account for 19,733 added lines out of the 21,171 in the diff. Excluding those four files means the actual diff is

> 69 files changed, 1,439 insertions(+), 1,966 deletions(-)

so this is actually a 527-loc negative change :)

---

Edit:
Overview of things I tested per tracker:

<details>

<summary>AniList</summary>

- [x] Add fresh track
  - [x] Public
  - [x] Private
- [x] Add remote-existing track
- [x] Update track
  - [x] Status
  - [x] Progress
  - [x] Score
  - [x] Start date
  - [x] Finish date
  - [x] Private state
- [x] Search
  - [x] Normal (by text/title)
  - [x] Search by ID
- [x] Remove track
  - [x] Local only
  - [x] Local + Remote
- [x] Handling update of a remote-removed track
- [x] Handling fetch of a remote-removed track (tracker sheet open)
- [x] Handling of automatic progress bumps during reading
</details>

<details>

<summary>Kitsu</summary>

- [x] Add fresh track
  - [x] Public
  - [x] Private
- [x] Add remote-existing track
- [x] Update track
  - [x] Status
  - [x] Progress
  - [x] Score
  - [x] Start date
  - [x] Finish date
  - [x] Private state
- [x] Search
  - [x] Normal (by text/title)
  - [x] Search by ID
  - [x] Search by Slug
- [x] Remove track
  - [x] Local only
  - [x] Local + Remote
- [x] Handling update of a remote-removed track
- [x] Handling fetch of a remote-removed track (tracker sheet open)
- [x] Handling of automatic progress bumps during reading
</details>

<details>

<summary>Shikimori</summary>
- [x] Add fresh track
  - [x] Public
- [x] Add remote-existing track
- [x] Update track
  - [x] Status
  - [x] Progress
  - [x] Score
- [x] Search
  - [x] Normal (by text/title)
  - [x] Search by ID
- [x] Remove track
  - [x] Local only
  - [x] Local + Remote
- [x] Handling update of a remote-removed track
- [x] Handling fetch of a remote-removed track (tracker sheet open)
- [x] Handling of automatic progress bumps during reading
</details>

<details>

<summary>Suwayomi</summary>
- [x] Update track
  - [x] Status
  - [x] Progress
- [x] Correct matching on Suwa sheet entry tap
- [x] Handling of automatic progress bumps during reading
- [x] Correct sending of deletion pref on progress update
</details>

app/build.gradle.kts
```diff
@@ -337,6 +341,76 @@ dependencies {
...
+            )
+            endpointUrl.set("https://kitsu.app/api/graphql")
+            schemaFile.set(file("$srcDir/kitsu.graphqls"))
+        }
+
+        // A date, expressed as an ISO8601 string
+        mapScalarToKotlinString("Date")
+        // An ISO 8601-encoded datetime
+        mapScalarToKotlinString("ISO8601DateTime")
+        // A loose key-value map in GraphQL
```
Syer10, line 382: Apollo has KotlinX-DateTime Scalars if you want to swap to that. https://github.com/apollographql/apollo-kotlin-adapters

app/build.gradle.kts
```diff
@@ -337,6 +341,76 @@ dependencies {
...
+        mapScalar(
+            "Map",
+            "kotlin.collections.Map<String, Any?>",
+            "mihon.graphql.kitsu.adapter.KitsuMapAdapter",
+        )
```
Syer10, line 387: This is so dumb, not part of the GQL spec. Really Kitsu...

app/build.gradle.kts
MajorTanya: I saw those but we only show `Date` values as display text (e.g. Search results that show the start date).

Though maybe it'd be worth it for the user's start/finish dates which are those ISO dates. But it's also just an Instant parse/toString each. I'll take a look at ~~whether they have kotlinx.datetime ones and see~~ how it works.

MajorTanya: Have decided to use those adapters, but only for Kitsu's ISO8601DateTime, as all other date/time types are partial or date-only.

Had to opt-in to ExperimentalTime because the library still has the outdated annotation for `kotlin.time.Instant`. Tracked by apollographql/apollo-kotlin-adapters#42.

AntsyLich (approved):
i ain't reading all that if it breaks you can fix it

core/common/src/main/kotlin/eu/kanade/tachiyomi/network/ApolloExtensions.kt
MajorTanya, line 1: @Syer10 Here's the new dataOrElse shape I was thinking of.

Syer10: Looks good

MajorTanya:
One thing I noticed is that `dataOrElse` is hiding connection problems as "normal" errors.

For example, if I DNS block AniList, _currently_ the app shows `Failed to connect to graphql.anilist.co`.
With these changes, it will instead result in the function call invoking the default value of `dataOrElse` and returning that.

On the tracker sheet this means it will show "AniList: Could not find manga", the message that is meant to indicate "Hey User, this manga is not in your list on the servers of AniList" which is not correct.

Since this is an ApolloNetworkException, should `dataOrElse` unconditionally rethrow those? Should it throw the `cause` of it? (which'd be something like `java.net.ConnectException` with the error message being shown on `main` right now (`Failed to connect to ...`)

I'm not sure what to do there.

MajorTanya:
> One thing I noticed is that `dataOrElse` is hiding connection problems as "normal" errors.
> 
> For example, if I DNS block AniList, _currently_ the app shows `Failed to connect to graphql.anilist.co`. With these changes, it will instead result in the function call invoking the default value of `dataOrElse` and returning that.
> 
> On the tracker sheet this means it will show "AniList: Could not find manga", the message that is meant to indicate "Hey User, this manga is not in your list on the servers of AniList" which is not correct.
> 
> Since this is an ApolloNetworkException, should `dataOrElse` unconditionally rethrow those? Should it throw the `cause` of it? (which'd be something like `java.net.ConnectException` with the error message being shown on `main` right now (`Failed to connect to ...`)
> 
> I'm not sure what to do there.

Solved in 390bffa by rethrowing `ApolloNetworkException.cause` (if it exists). Not great, not terrible.

As a concrete reference: In Apollo's code, it assigns [an IOException thrown by OkHttp to the cause of a new ApolloNetworkException](https://github.com/apollographql/apollo-kotlin/blob/68aef25e9388697a73b3a7960efaf684ea4f6d93/libraries/apollo-runtime/src/jvmCommonMain/kotlin/com/apollographql/apollo/network/http/DefaultHttpEngine.jvm.kt#L108-L133). Sounds about as good as it gets to me.

Example screenshot:
(I used a fake domain to provoke a connection failure since my DNS block wasn't working properly for my phone, idk caching probably).

PR #3828, opened 2026-08-21
Add refresh button to tracker for new profile info

MajorTanya:
As-is, allows a user to manually refresh their current user data from the tracker in order to update Mihon's knowledge of remote settings (e.g. nickname changes, AL/MB/Kitsu score types, etc.).

Avoids doing a whole logout-login circus.

Open issues:
- [x] **Crashes the entire app if the fetch fails**
  How to surface errors best? I couldn't figure out how to make the flow I used emit some sort `sealed interface RefreshEvent` since Kotlin was unhappy about it. Also didn't figure out how to then show that error. Also also don't know where the exception catching should actually happen and what to do with the caught exception. Please advise.
  - possible solution in 9bfaca57fea3057f065a3c29b7bb528608ac09ad but I'd like feedback on that approach

- [x] Enhanced trackers get the button but it does nothing
  Should they get it at all (and what should it do)?
  **Solution**: No button for enhanced trackers.

What the buttons look like:
image


MajorTanya:
I think I figured something out for the error surfacing actually, please check the approach in 9bfaca57fea3057f065a3c29b7bb528608ac09ad.

What it looks like:
(this is from a test HttpException I threw in Kitsu's method)

[Screen_recording_20260821_074717.webm](https://github.com/user-attachments/assets/e9481ccf-06a7-497a-9f42-dd2e1e92a801)

MajorTanya:
- centred the tracker name when the user is not logged in (not strictly in scope but it's one line so I thought whatever) (296f7e9).
- Enhanced trackers no longer show the refresh button since it has no purpose on them (443e555). This resolves the open question re: button for them.

Below is a before/after comparison of what the menu looks like:
(note that the icons look slightly different due to our migration to Valkyrie for icon generation)
(I also logged out of MAL to show what a normal tracker looks like when logged out in the list, you can see what Suwayomi looks like in the before image)
(I have also replaced the original PR body image with the "After" to reflect the actual changes)

| Before | After |
|-|-|
| image | image |

---

I can revert 296f7e9 if you want me to PR that separately or something.

MajorTanya:
@AntsyLich undrafted for your reviewing pleasure

AntsyLich (approved):

PR #3842, opened 2026-08-22
CloudflareInterceptor: Exit early in WebView when challenge is interactive

tigeryu8900:
Listens for the `interactiveBegin` message event in the WebView, which is only sent when the challenge is interactive.

Check the `cf-mitigated` header to detect a Cloudflare challenge [as mentioned in Cloudflare's documentation](https://developers.cloudflare.com/cloudflare-challenges/challenge-types/challenge-pages/detect-response/). I've verified that geo-blocked responses don't have the `cf-mitigated` header.

### DIFF
--- a/core/common/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
+++ b/core/common/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
@@ -86,7 +86,7 @@ class CloudflareInterceptor(
                     // The challenge cannot be solved non-interactively, abort.
                     latch.countDown()
                 }
-            }, "jsInterface")
+            }, "mihon")
 
             webview.webViewClient = object : WebViewClient() {
                 override fun onPageFinished(view: WebView, url: String) {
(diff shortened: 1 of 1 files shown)


core/common/src/main/kotlin/eu/kanade/tachiyomi/network/interceptor/CloudflareInterceptor.kt
```diff
@@ -87,6 +79,15 @@ class CloudflareInterceptor(
...
             webview = createWebView(originalRequest)
 
+            webview.addJavascriptInterface(object {
+                @Suppress("unused")
+                @JavascriptInterface
+                fun interactiveDetected() {
+                    // The challenge cannot be solved non-interactively, abort.
+                    latch.countDown()
+                }
+            }, "jsInterface")
```
AntsyLich:
```suggestion
            }, "mihon")
```

AntsyLich (changes requested):

AntsyLich (approved):

tigeryu8900:
Forgot to run spotless

PR #3858, opened 2026-08-27
CloudflareInterceptor: Attempt to solve challenge when interactive

tigeryu8900:
When an interactive challenge is detected, attempt to solve the challenge instead of aborting by attaching the WebView to the current view and sending the Tab and Space keys.

I also added a locking mechanism to make sure that only one request per host is solving the challenge at a time.

I took inspiration from an unrelated project.

This is the overall logic of the solver:
1. Obtain a read lock and try `chain.proceed(request)`. If the response is successful, release the read lock and return the response. Otherwise, get the `cf_clearance` cookie before releasing the read lock.
2. Obtain a write lock and check if Cloudflare is already bypassed in another call. If it is, release the write lock and return `chain.proceed(request)`.
3. Check if the view group of the current activity is available. If it isn't, inject the fallback JavaScript solver.
4. Load the page in a WebView.
5. On an `interactiveBegin` event, if the fallback solver isn't injected, attach the WebView to the view group of the current activity and simulate Tab and Space keys. If the view group is unavailable or keys fail to send, inject the fallback solver and reload the page. Abort if the `complete` event doesn't happen within five seconds.
6. After the solver completes, release the write lock before returning `chain.proceed(request)`.


AntsyLich:
i'm not sure about the auto solver. it might get us blocked permanently by cloudflare. the other changes seems to be useful.

tigeryu8900:
I see. Just to clarify, by the auto solver you mean the simulated key events and the injected JavaScript solver, or just the JavaScript solver?

AntsyLich:
both

tigeryu8900:
Got it. I removed the solver.

tigeryu8900:
I discussed the permanent block issue on an unrelated project, and I don't think Cloudflare has a way of detecting a Mihon user if we don't inject JavaScript in the main world.

I'm thinking we can add back the key events based solver but make it off by default in the settings.

If you think it's a too big of a risk, I'll leave it out of the PR.

Smol-Ame:
Don't put random projects unaffiliated with Mihon without approval, thanks.

If whatever discussion is that important & it can't be easily explained in a message or two, come to the Mihon Discord and check with Antsy. Otherwise keep it between your PR and how it affects Mihon.

tigeryu8900:
Got it. I'll move to Discord if this becomes a long discussion.

PR #3876, opened 2026-08-29
Clean up some missed unused imports

MajorTanya:
We should consider explicitly enabling the ktlint rule for unused imports, which used to be enabled by default until v1.7.0 (which we've been using since last July in d805f0cd with v1.7.1).

ktlint issue about disabling it:
https://github.com/ktlint/ktlint/issues/3038

docs about the rule:
https://ktlint.github.io/ktlint/latest/rules/standard/#no-unused-imports

The required change which I could add if wanted (can confirm that enabling this would have made Spotless complain on a `spotlessCheck` run on the project):
(Note that ktlint themselves believe this rule has shortcomings which are described in the GH issue)
```patch
Index: .editorconfig
IDEA additional info:
<+>UTF-8
diff --git a/.editorconfig b/.editorconfig
--- a/.editorconfig	(revision 110058815a8a944475dfe5c3f78354edd0ed4a69)
+++ b/.editorconfig	(date 1788047333771)
@@ -30,3 +30,4 @@
 ktlint_standard_type-argument-comment = disabled
 ktlint_standard_type-parameter-comment = disabled
 ktlint_standard_blank-line-between-when-conditions = disabled
+ktlint_standard_no-unused-imports = enabled

```

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/components/TabbedDialog.kt
+++ b/app/src/main/java/eu/kanade/presentation/components/TabbedDialog.kt
@@ -9,7 +9,6 @@ import androidx.compose.foundation.layout.wrapContentSize
 import androidx.compose.foundation.pager.HorizontalPager
 import androidx.compose.foundation.pager.PagerState
 import androidx.compose.foundation.pager.rememberPagerState
-import androidx.compose.material3.HorizontalDivider
 import androidx.compose.material3.Icon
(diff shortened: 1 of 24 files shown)


AntsyLich (approved):

AntsyLich:
we can probably use detekt for this, only enabling it for import related issue.

PR #3877, opened 2026-08-30
Migrate composeGrid to new group ID

cheonjaeung:
Hi, I'm the maintainer of this library.
After version 2.0.0, the group ID and package name were changed to `com.cheonjaeung.compose.grid`.
This can be drop-in replacement because this project only uses `VerticalGrid`. Additionally, you'll get several bug fixes and performance improvements.


AntsyLich:
Thanks for the PR <3

PR #3887, opened 2026-09-03
Fixed crash caused by incorrect handling of kitsu score type from login

Secozzi:
Migration is there so users won't have to relogin


MajorTanya:
oh yeah that sure was an oversight huh

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Fixed app and extension update check running again on configuration change ([@AntsyLich](https://github.com/AntsyLich)) ([#3708](https://github.com/mihonapp/mihon/pull/3708))
 - Fixed MangaBaka user start/finish dates drifting in negative offset timezones ([@MajorTanya](https://github.com/MajorTanya)) ([#3711](https://github.com/mihonapp/mihon/pull/3711))
 - Fixed MangaBaka scores being wrong when score step size was set to > 1 ([@MajorTanya](https://github.com/MajorTanya)) ([#3740](https://github.com/mihonapp/mihon/pull/3740))
+- Fixed crash caused by incorrect handling of kitsu score type from login ([@Secozzi](https://github.com/Secozzi)) ([#3887](https://github.com/mihonapp/mihon/pull/3887))
```
AntsyLich: Don’t think the issue was part of any stable release

Secozzi: I never checked but based on the commit date ig it wasn't

MajorTanya:
actually, couldn't this be solved without migration if we lowercase it on the calling end?

Secozzi:
> actually, couldn't this be solved without migration if we lowercase it on the calling end?

Good point

PR #3902, opened 2026-09-05
Add paged reader option to disable swipe

Ben-Kerman:
This improves usability on e-ink devices as it prevents accidental large-scale animations triggered by swiping.

Also helps with system-level swipe gestures, e.g. I have my e-reader tablet configured to do a full redraw when swiping from the left, which doesn't really work if it also shifts the image slightly (and thus can cause new ghosting).

---

I don't have a good overview of the codebase and very little experience with Android development, so I really hope that I did everything right and didn't accidentally break something else.

I did confirm that double-tap zoom, pinch zoom, and moving by swiping when zoomed in still work with the option enabled, and that the reader is still very much usable. I didn't find anything that was broken by this, but I also haven't used the app that much, so maybe there's something I didn't check or didn't realize was broken.

---

The new WebGPU renderer already allows having the same effect by disabling page transition animations, with the added benefit that you can still use swipes to switch pages even without an animation. I tried looking into if that would also be possible to implement for the old paged reader, but I'd guess you'd have to change the DirectionalViewPager directly for that and that's way beyond my abilities, so I instead opted for disabling swiping completely.

### DIFF
--- a/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsReaderScreen.kt
+++ b/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsReaderScreen.kt
@@ -277,6 +277,10 @@ object SettingsReaderScreen : SearchableSettings {
                     title = stringResource(MR.strings.pref_page_rotate_invert),
                     enabled = rotateToFit,
                 ),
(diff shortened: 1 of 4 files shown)


AntsyLich:
Hard blocking touch interceptions will probably do more bad than good

PR #3903, opened 2026-09-05
Add filter dialog in Migrate Mangas Screen

frankete333:
Add a Filter action at the top right of the Migrate Manga Screen (After selecting a source)
That action opens a pretty similar dialog to the one in `Updates`, to filter by:
```
- Downloaded
- Unread
- Started
- Bookmarked
- By Category
```

(The warning in the `Categories` section is the same one as in `Updates`, since it uses the same logic to filter out Items)

<table>
  <tr>
    <td><b>Filter</b></td>
    <td><b>Category</b></td>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

### DIFF
--- a/app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/manga/MigrateMangaFilterDialog.kt
+++ b/app/src/main/java/eu/kanade/tachiyomi/ui/browse/migration/manga/MigrateMangaFilterDialog.kt
@@ -1,0 +1,125 @@
+package eu.kanade.tachiyomi.ui.browse.migration.manga
+
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.ColumnScope
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.rememberScrollState
+import androidx.compose.foundation.verticalScroll
+import androidx.compose.material3.HorizontalDivider
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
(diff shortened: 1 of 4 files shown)


AntsyLich:
What’s the point?

frankete333:
> What’s the point?

To make it easier to migrate big libraries when sources die
Last time it happened (Comick), it would have been great to be able to filter by my `actively reading` collection, to focus on migrating those first, so that I can get the updates on new chapters, since I wasn't in a rush to migrate the 100+ in the plan to read, or to hide the ones I still have downloaded chapters for, since I don't need to migrate those until I finish with the downoladed chapters, etc.

Smol-Ame:
You can already do that in the Library page though, category(s) & filtering, long-press to select the ones you want, then hit Migrate & continue the migration process

frankete333:
Not really, because there you can't really filter by the dead source (Or at least I haven't seen a filter by source on the library page), and you would be seeing all the titles, not only the ones from the source you want to migrate out of.
If your solution to this would be having that filter by source on the library page, I can change the PR to that, but it does feel like we would be adding extra things to the library page to support better migrations instead of adding them to the actual migration page

Smol-Ame:
I mean, the Migrate page is for you to migrate by source, but you wanted specifics so the Library page is the specifics.

You can semi-search by source if you search the source name or with `src:` if you have the extension's ID (in your library).

MajorTanya:
> I mean, the Migrate page is for you to migrate by source, but you wanted specifics so the Library page is the specifics.
> 
> You can semi-search by source if you search the source name or with `src:` if you have the extension's ID (in your library).

It's `src_id:`.

PR #3941, opened 2026-09-12
Don't try to create slider state when slider can't or won't be rendered

AntsyLich:


app/src/main/java/eu/kanade/presentation/reader/components/ChapterNavigator.kt
```diff
@@ -77,14 +77,20 @@ fun ChapterNavigator(
...
+    val state = if (totalPages > 1) {
+        key(totalPages) {
+            rememberSliderState(
+                value = currentPage.toFloat(),
+                steps = totalPages - 2,
+                trackRange = 1f..totalPages.toFloat(),
+            )
+        }
+            .also {
+                it.value = currentPage.toFloat()
```
MajorTanya, line 89: Isn't this redundant with line 84?

AntsyLich: if you mean like 83 then no

MajorTanya: Oh yeah 83 obviously. But how it assigning the same float here not redundant with the float passed in L83?

AntsyLich: because the state is shared across ui changes as such its value can be different than currentPage

PR #3952, opened 2026-09-14
Allow user to load installed extensions regardless of content warning

### LINKED ISSUE #1673: Installed NSFW Extensions Should Remain Accessable & Receive Updates Even if NSFW Visibility is Off

AntsyLich:
Closes #1673


AwkwardPeak7:
should this be changed to allow library entries from those sources to continue to work?

AntsyLich:
I still need to load the extension in that case

PR #3948, opened 2026-09-13
Use modern OpenDocument for backup file selection

MajorTanya:
This might also fix the underlying problem of #3947 by including `takePersistableUriPermission`.

I don't have any nonstandard devices available, but this works fine in an A17 Pixel emulator & on my A14 Xiaomi device. Whether it actually fixes the issue experienced by the user in the aforementioned PR, I cannot say. Maybe they can try this out as a patch and report back if it works.

The reason why I think this _might_ work is that restoration happens in a Work unit which comes ~later, so any grants received by GetContent may have expired already.

PR #3947, opened 2026-09-13
Fix restoring backups picked from ExternalStorage volume root

nazzeDe:
## Summary

Restore backup from the system picker can start on the ExternalStorage volume root (on Waydroid this is **WayDroid x86_64 Device**). Picking a `.tachibk` there returns a raw document URI:

```
content://com.android.externalstorage.documents/document/primary:Documents/.../autobackup/*.tachibk
```

`contentResolver.openInputStream(uri)` returns null for that URI, so `BackupDecoder.decode` crashed with NPE and the UI showed **Invalid backup file**.

The same file picked from Recent / Documents (MediaStore) already worked.

This keeps `GetContent` as-is and opens the file through the app storage location instead:

1. original picker URI (MediaStore / granted documents)
2. rewrite to a tree document URI using the storage directory and persistable grants
3. automatic backup directory (`StorageManager` / UniFile `autobackup/<filename>`)
4. filesystem path as a last resort

Files outside the storage tree still fail with `IOException` (not NPE).

## Test plan

- [x] JVM: `BackupDecoderTest` (plain protobuf, gzip, JSON rejection)
- [x] JVM: `BackupLocationTest` (HITL volume-root URI relative path / filename / filesystem path; tree document suffix; other storage tree; Download; MediaStore)
- [x] HITL (Waydroid): Restore from **WayDroid x86_64 Device** → `Documents/Mihon_test/autobackup/*.tachibk` succeeds after setting storage location
- [x] Recent / Documents picks still open

PR #3960, opened 2026-09-14
Fix backup restore NPE for ExternalStorage document URIs

nazzeDe:
Restore from a DocumentsUI volume-root pick still NPEs after #3948.                             
                                                                                                   
   `takePersistableUriPermission` succeeds, but `BackupDecoder` does `openInputStream(uri)!!` on   
 an ExternalStorage document URI. On Waydroid that throws NPE (null fd), on the confirmation       
 screen, before restore starts.                                                                    
                                                                                                   
   The same file opens as a MediaStore URI (what Recent already returns):                          
                                                                                                   
   `MediaStore.getMediaUri(uri)` → `content://media/external_primary/file/…`                       
                                                                                                   
   This retries that when the picker URI cannot be opened, and drops `!!`.                         
                                                                                                   
   Related: #3947, #3948


MajorTanya:
This is sounding more and more like a Waydroid-side issue to me. Mihon is interacting with the file system the way it's supposed to as per the SAF as far as I can tell.

nazzeDe:
I agree you judge and thank you for review

PR #3966, opened 2026-09-15
Add an option to disable the MD5 suffix in chapter names

### LINKED ISSUE #3880: Add a settings option to disable chapter hashes

Recently a bunch of issues were created about chapter hashes, which makes it impossible to rename files to accommodate to source changes. I suggest to add a simple switch in settings to turn hashes on and off.

### Other details

Chapter hashes were supposed to solve a problem with several chapters resolving to the same download name.
But they created another, much more common issue. When something is changed in chapter naming by a source, all downloaded chapters become invalid, and cannot be simply fixed by mass renaming. From my experience, I have these problems almost every week. Without hashes, it take just a minute or two to fix it. Now I need to also remove hashes from all downloaded files, which is much harder. I have to mess with regular expressions, and that's always scary, because with one error I can totally mess up all file names. But the only alternative is to re-download all chapters. For a title with 1000 chapters taking 10 GB this can take quite long.
Sometimes it's even worse than just one title. If a translator group is renamed, all its titles become broken.

In ideal world, these hashes could only be added to chapters that _really_ have a duplicate one. But if that's not possible, a simple switch to opt-out from this would be welcome. That should be very simple change, because chapters without hashes are still work normally. So only a condition in one place is needed.

P.S.: I believe it's not a duplicate of former tickets. They were about removing hashes completely, I suggest to get people a choice.

### Acknowledgements

- [x] I have updated the app to version **[0.20.4](https://github.com/mihonapp/mihon/releases/latest)**.

choppeh:
Closes #3880

Adds an advanced setting to toggle the 6-character MD5 hash suffix appended to downloaded chapter directory names and `.cbz` archives:
- **Enabled by default** to preserve the app's current behavior.
- Legacy directory lookup ensures previously downloaded chapters (both with and without the hash suffix) continue to be recognized correctly regardless of the setting state.


AntsyLich:
Make it a setting to enable and add a migration so for old user it gets automatically enabled so the behavior stays consistent for them.

AntsyLich:
Also if possible look into adding support to load the hashed folders even when the settings is turned off

domain/src/main/java/tachiyomi/domain/library/service/LibraryPreferences.kt
```diff
@@ -219,6 +219,10 @@ class LibraryPreferences(
         "disallow_non_ascii_filenames",
         false,
     )
+    val enableChapterNameHash: Preference<Boolean> = preferenceStore.getBoolean(
+        "pref_enable_chapter_name_hash",
+        true,
```
AntsyLich: Disable by default. Add a migration to enable it for old users.

AntsyLich:
Run `./gradlew spotlessApply`

Issue #2030, opened 2025-04-21
Make download cache persistent and indexing efficient


AntsyLich:
### Describe your suggested feature

Downloads should be reindexed every time app comes to foreground instead of every hour. Further more we should utilize last modified and other approaches to indexing efficient

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have updated the app to version **[0.18.0](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I will fill out all of the requested information in this form.

PR #3968, opened 2026-09-16
Fix download index cache persistence and startup race condition

choppeh:
Previously, calling invalidateCache() deleted diskCacheFile immediately, leaving the cache cleared if renewal was cancelled or delayed, which caused downloaded chapters to incorrectly appear missing in the UI. Additionally, an uncoordinated startup sequence allowed UI queries to trigger full filesystem scans before the disk cache finished reading.

To resolve this, diskCacheFile now uses AtomicFile to prevent file corruption upon abrupt process termination, while initial disk loading is synchronized via initJob to ensure UI queries do not trigger premature re-indexing. Invalidation now marks RootDirectory as expired through its embedded creation timestamp and TTL rather than deleting the file upfront. Furthermore, disk writes are deferred until indexing successfully completes, and ProtoBuf serialization is made thread-safe

Additionally, a new preference has been added under Advanced Settings allowing users to customize the download cache TTL (expiration time), defaulting to 1 hour with a validated minimum threshold of 30 minutes (1800 seconds).


AntsyLich:
Run `./gradlew spotlessApply`

AntsyLich:
Imo the PR doesn’t solve #2030. Can you remove that from changelog and PR body

choppeh:
> Imo the PR doesn’t solve #2030. Can you remove that from changelog and PR body

Updated the changelog with the startup race condition fix

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is a modified version of [Keep a Changelog](https://keepachangelog.co
 - Add support for using the user's chosen rating system for Kitsu ([@MajorTanya](https://github.com/MajorTanya)) ([#3818](https://github.com/mihonapp/mihon/pull/3818))
 - Add support for the Year, Month, and Day fields in ComicInfo.xml files for chapter dating ([@MajorTanya](https://github.com/MajorTanya)) ([#3967](https://github.com/mihonapp/mihon/pull/3967))
 - Add refresh buttons to trackers in Settings to update displayed usernames (all trackers) & rating systems (where supported) ([@MajorTanya](https://github.com/MajorTanya)) ([#3828](https://github.com/mihonapp/mihon/pull/3828))
+- Add setting to configure download cache reindex interval ([@choppeh](https://github.com/choppeh)) ([#3968](https://github.com/mihonapp/mihon/pull/3968))
```
```suggestion
- Add setting to configure the download cache reindex interval ([@choppeh](https://github.com/choppeh)) ([#3968](https://github.com/mihonapp/mihon/pull/3968))
```

PR #3984, opened 2026-09-19
Replace voyager with compose navigation 3

Secozzi:
Why? Voyager doesn't seem to be that well maintained and navigation 3 is the official library for navigation.

This also fixes an issue where the permission checkboxes during onboarding aren't checked if using a theme other than the default.


gradle/libs.versions.toml
```diff
@@ -57,6 +57,7 @@ materialKolor = "5.0.1"
 metro = "1.4.4"
 mockk = "1.14.11"
 moko-resources = "0.26.4"
+nav3-core = "1.2.0-rc01"
```
AntsyLich:
```suggestion
androidx-navigation3 = "1.2.0-rc01"
```

app/src/main/java/eu/kanade/tachiyomi/extension/util/ExtensionLoader.kt
```diff
@@ -8,6 +8,7 @@ import android.os.Build
 import androidx.core.content.pm.PackageInfoCompat
 import eu.kanade.domain.extension.interactor.TrustExtension
 import eu.kanade.tachiyomi.extension.model.Extension
+import eu.kanade.tachiyomi.extension.util.ExtensionLoader.loadExtension
```
AntsyLich: ?

Issue #3512, opened 2026-07-03
Unable to download chapters on r7775


Pecs1:
### Steps to reproduce

1. open any manga
2. click to download new chapters

### Expected behavior

chapters should be downloaded

### Actual behavior

in **More > Download Queue** it gets stuck for e.g. at 25/25
in the manga view there is the red error icon

### Mihon version

Beta r7775

### Android version

Android 15

### Device

Samsung a53

### Other details

Happened to me on multiple sources + i was told on discord to create an issue request

<img width="1080" height="2272" alt="Image" src="https://github.com/user-attachments/assets/fefeb468-0dd9-4475-af76-095c6257f6df" />

<img width="1080" height="2272" alt="Image" src="https://github.com/user-attachments/assets/390ad24e-c5aa-4f51-8e24-ec81dac51ec7" />

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open or closed issue.
- [x] I have written a short but informative title.
- [x] I have gone through the [FAQ](https://mihon.app/docs/faq/general) and [troubleshooting guide](https://mihon.app/docs/guides/troubleshooting/).
- [x] I have updated the app to version **[0.20.0](https://github.com/mihonapp/mihon/releases/latest)**.
- [x] I have filled out all of the requested information in this form, including specific version numbers.
- [x] I understand that **Mihon does not have or fix any extensions**, and I **will not receive help** for any issues related to sources or extensions.

shaffinX:
yes same issue with me i downloaded chapters but it stuck on 10/10 and the same exclaimation mark, i usually download the test build after few days. My device is Galaxy S21 5G.

MajorTanya:
Does Mihon have any notifications regarding the error? Alternatively, could someone enable verbose logging in the advanced settings, restart the app, clear the download queue, and try to download one (1) chapter, dump the logs, and attach them here?

Pecs1:
Nope no notification regarding that

and here is the log as requested
[mihon_crash_logs.txt](https://github.com/user-attachments/files/29634562/mihon_crash_logs.txt)

MajorTanya:
```log
2026-07-03 13:40:11.424 +0200 17240 17331 E Downloader: Failed to split downloaded image
2026-07-03 13:40:11.424 +0200 17240 17331 E Downloader: java.lang.IllegalStateException: Page 6 not found while splitting
[...]

PR #3990, opened 2026-09-20
Fix chapters stuck in error after an interrupted tall-image split

mehdiacho:
Killing the app while a chapter with "Split tall images" enabled is downloading leaves that chapter permanently in error. Every retry fails the same way, with all of its pages already on disk.

`ImageUtil.splitTallImage` writes each part and deletes the original page **last**. If the process dies part way through, the unsplit original is still sitting in the temp directory next to however many parts got written:

```
009.webp        1016808
009__001.jpg          0
```

On the next attempt, `splitTallImageIfNeeded` picks any file with the page's prefix via `firstOrNull`. If that happens to be a part, the `startsWith("${filenamePrefix}__")` check concludes the page is already split and returns, leaving `009.webp` where it is.

`isDownloadSuccessful` then counts files, and its "count a split page once" rule only skips parts after the first:

```kotlin
fileName.contains("__") && !fileName.endsWith("__001.jpg") -> false
```

So `009.webp` and `009__001.jpg` are both counted. The file count comes out higher than the page count, the chapter goes to `ERROR`, and the next retry reaches exactly the same state.

The fix picks the original deliberately rather than taking whatever `firstOrNull` returns, and treats the page as already split only when no original remains. `splitTallImage` deletes pre-existing parts before writing each one, so re-splitting is safe and a directory left in this state repairs itself on the next attempt.

That branch in `downloadChapter` also set `ERROR` without logging or notifying anything, which is the only failure path in that function that reports nothing at all. It's also where a download resumed after the process was killed lands, so it now logs what didn't add up.

A second commit fixes a different way the same function picked the wrong file, which I noticed while looking at this one. `getOrDownloadImage` sizes the page number to the chapter's page count, so in a chapter of a thousand pages page 1 is written as `0001.webp`, while `splitTallImageIfNeeded` formatted its own prefix with a hard-coded `%03d` and matched on it bare. It looked for `001`, missed its own file, and matched `0010.webp` instead, which is page 10. It takes the prefix the file was named with now, and anchors the match on what follows the number, the way `isDownloadedPageImage` already does. `.tmp` files are excluded there for the same reason.

**Related:** #3512 was the same visible symptom (stuck at n/n with the error icon) from a different cause in the same code path.

Not a visual change, so no screenshots.

**Tested:** on a 22101316UCP running Android 14, on a build from `main`. Force-stopped the app mid-download of a long-strip chapter, confirmed the leftover original and the zero-byte part on disk over adb, and confirmed that a chapter which had been stuck for over half an hour completed on the first resume after the fix. `spotlessCheck`, `testDebugUnitTest` and `verifySqlDelightMigration` all pass locally.


MajorTanya:
Can you get your Copilot out of here please

mehdiacho:
Sorry about that, it was on by default on my account. Turned off now.

AntsyLich:
make this less ai slop

mehdiacho:
Rewrote the description and cut the comments down to what isn't obvious from the code. Same two commits otherwise.

Issue #3958, opened 2026-09-14
Preference.PreferenceItem.SliderPreference+BasicListPreference has performance issues


LagradOst:
I have deeply inspected and tested the Preference system in mihon because I implemented a similar system in my app. And while the system is great overall, I have noticed some minor oversights. One of these is that `SliderPreference` does not use `PreferenceData<Int>`. This causes the entire screen to be recomposed after 1 edit. Check in the Layout Inspector or adding a single `println("getPreferences got called again!")` and you will see that SliderPreference (and BasicListPreference) causes everything to recompose. 

While this does not seam like a great issue, it can cause lag spikes when working with larger uis or when editing it quickly. However this is somewhat masked by the static nature of settings. collectAsState should never be used in a getPreferences block, but instead inside the widget so only the widget gets updated. 

https://github.com/mihonapp/mihon/blob/cbafe9bf3f274b5655415b539ba0e45f31ec2015/app/src/main/java/eu/kanade/presentation/more/settings/screen/SettingsDownloadScreen.kt#L37

nullNitesh:
Thanks for pointing this out! I've implemented a fix for this issue.
- https://github.com/mihonapp/mihon/pull/3992

PR #3992, opened 2026-09-20
refactor: update SliderPreference to use PreferenceData and collect state internally

nullNitesh:
The root cause of the performance issue was that `SliderPreference` accepted an `Int` value directly instead of a `PreferenceData<Int>`. This forced the parent composables (like `SettingsDownloadScreen` and `SettingsReaderScreen`) to `collectAsState()` on the preferences, which caused the entire screen to recompose whenever a slider was dragged.

To resolve this, I made the following changes:

1. **Updated `SliderPreference` definition:**
   Modified `SliderPreference` to accept `val preference: PreferenceData<Int>` instead of a raw `Int`. I also replaced `valueString: String?` with a composable lambda `val valueText: @Composable (Int) -> String` so the text can be computed dynamically without requiring the parent to recompose.

2. **Scoped `collectAsState()` to the widget:**
   In `PreferenceItem.kt`, the `SliderPreference` matcher now calls `item.preference.collectAsState()` directly inside the `SliderPreference` item component. Now, dragging the slider only recomposes the slider widget itself.

3. **Migrated usages & cleaned up scaling logic:**
   Updated all `SliderPreference` usages across the app. Notably, for `flashMillisPref` which is an unscaled property that operates in intervals, I was able to remove the manual scaling entirely by configuring `valueRange = 300..4500 step 300` and relying on native Kotlin progressions to handle the steps cleanly.

This completely resolves the layout stutter and aligns the preferences with how they should be localized within widgets.


MajorTanya:
Didn't we have an issue about this already? Why not link it?

Edit:
Read about [autoclosing keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) please.

MajorTanya:
Autoclosing keywords have to be in the [PR body not the comments](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue).

You can edit the PR body and add "Closes #3958" (without quotes obviously).

AntsyLich:
tell ai to remove the redundant variables too

